PR #27, opened 2020-05-29
Adding maxPatchPublishDate as config settings input

rane-rajasi:


src/core/src/core_logic/ExecutionConfig.py
```diff
@@ -51,6 +51,7 @@ def __init__(self, env_layer, composite_logger, execution_parameters):
         self.included_classifications_list = self.__get_execution_configuration_value_safely(self.config_settings, Constants.ConfigSettings.CLASSIFICATIONS_TO_INCLUDE, [])
         self.included_package_name_mask_list = self.__get_execution_configuration_value_safely(self.config_settings, Constants.ConfigSettings.PATCHES_TO_INCLUDE, [])
         self.excluded_package_name_mask_list = self.__get_execution_configuration_value_safely(self.config_settings, Constants.ConfigSettings.PATCHES_TO_EXCLUDE, [])
+        self.max_patch_publish_date = self.__get_execution_configuration_value_safely(self.config_settings, Constants.ConfigSettings.MAX_PATCH_PUBLISH_DATE, [])
```
kjohn-msft, line 54: The last parameter is the default value. Max patch publish date is not an array like the others -- ensure the code works correctly when this value is not specified (this will be a valid case when imperative goes out). Ensure the intended behavior in that case is also in the design doc (currently missing).

src/core/tests/library/ArgumentComposer.py
```diff
@@ -68,7 +69,8 @@ def get_composed_arguments(self):
             "rebootSetting": self.reboot_setting,
             "classificationsToInclude": self.classifications_to_include,
             "patchesToInclude": self.patches_to_include,
-            "patchesToExclude": self.patches_to_exclude
+            "patchesToExclude": self.patches_to_exclude,
+            "maxPatchPublishDate": self.max_patch_publish_date
```
kjohn-msft, line 73: Since this is an optional parameter - tests need to account for it being present and not. To ensure both these high-level cases are considered, the value should not be present by default, and should be dynamically added to the object in the tests that will rely on it.

src/extension/tests/TestExtConfigSettingsHandler.py
```diff
@@ -170,17 +170,18 @@ def test_are_config_settings_valid(self):
                         self.config_public_settings_fields.include_classifications: ["Critical","Security"],
                         self.config_public_settings_fields.include_patches: ["*", "test*", "*ern*=1.2*", "kern*=1.23.45"],
                         self.config_public_settings_fields.exclude_patches: ["*", "test", "*test"],
-                        self.config_public_settings_fields.internal_settings: "<serialized-json>"
+                        self.config_public_settings_fields.internal_settings: "<serialized-json>",
+                        self.config_public_settings_fields.max_patch_publish_date: "2019-07-20T12:12:14Z"
```
kjohn-msft, line 174: Same note about two test paths being required -- this is an optional field that's not guaranteed to be there. The behavior will vary significantly in code that's not written yet so please ensure tests account for that and existing tests that don't have it included remain with it not included. (New tests for the feature - both positive and negative cases - need to have it considered).

src/extension/tests/helpers/1234.settings
```diff
@@ -11,7 +11,8 @@
                 "rebootSetting": "IfRequired",
                 "classificationsToInclude":["Critical","Security"],
                 "patchesToInclude": ["*", "test*","*ern*=1.2*", "kern*=1.23.45"],
-                "patchesToExclude": ["*", "test", "*test"]
+                "patchesToExclude": ["*", "test", "*test"],
+                "maxPatchPublishDate": "2019-07-20T12:12:14Z"
```
kjohn-msft, line 15: Same comment about bifurcation of tests -- for all new tests for the feature, use a different settings file.

src/core/src/core_logic/ExecutionConfig.py
rane-rajasi: Changed the default value to empty string. 
For Linux, max patch publish date is only used in status reporting, and that part will all scenarios are mentioned [here](https://microsoft.sharepoint.com/:w:/r/teams/GuestPatching/_layouts/15/Doc.aspx?sourcedoc=%7B3BB6D095-C8A3-4C5B-AC68-68F69FE3A0BC%7D&file=Patch%20Management%20V2%20-%20Extension%20Spec%20(OS-Agnostic).docx&action=default&mobileredirect=true), which is also linked in the design doc.

src/core/tests/library/ArgumentComposer.py
rane-rajasi: Changed. The test with a value present will be added later, with the status changes

src/extension/tests/TestExtConfigSettingsHandler.py
rane-rajasi: The config settings on line 182 test for the value not specified

src/extension/tests/helpers/1234.settings
rane-rajasi: Same answer as above :)

PR #28, opened 2020-08-03
Bug Fix: Using python version on machine while invoking core

rane-rajasi:


src/extension/src/ProcessHandler.py
```diff
@@ -55,13 +57,21 @@ def get_env_settings(self, ext_env_handler):
...
 
         args = " -sequenceNumber {0} -environmentSettings \'{1}\' -configSettings \'{2}\'".format(str(seq_no), env_settings, public_config_settings)
-        command = ["python " + exec_path + " " + args]
+
+        # Verify the python version available on the machine to use
+        self.logger.log("Python version: " + " ".join(sys.version.splitlines()))
+        if sys.version_info.major == 3:
+            python_cmd = "python3"
+        else:
+            python_cmd = "python"
```
kjohn-msft, line 72: As discussed, please mirror the logic in shim (probe-based discovery). Real world behavior is not as clear cut as this code as it depends on what came with the distribution and/or what was installed after.

Also add additional suitable logging for the behavior seen where core does not seem to run. A good way to do that maybe to capture data from stdout and stderr for some brief time and log it into the handler log if any data is seen there.

src/extension/src/Constants.py
```diff
@@ -84,6 +84,7 @@ class PatchOperationErrorCodes(EnumBackport):
         DEFAULT_ERROR = "ERROR"  # default error code
 
     ERROR_ADDED_TO_STATUS = "Error_added_to_status"
+    PYTHON_NOT_FOUND = "No python version found on the machine"
```
kjohn-msft, line 87: After some thought about this -- "Python version could not be discovered for core invocation."  If we use the message as is, it appears as though the blame is being shifted to the machine environment configuration while that's not true since the code throwing the error is in Python itself.

src/extension/src/ProcessHandler.py
```diff
@@ -55,20 +58,124 @@ def get_env_settings(self, ext_env_handler):
...
         env_settings = base64.b64encode(json.dumps(self.get_env_settings(ext_env_handler)).encode("utf-8")).decode("utf-8")
 
         args = " -sequenceNumber {0} -environmentSettings \'{1}\' -configSettings \'{2}\'".format(str(seq_no), env_settings, public_config_settings)
-        command = ["python " + exec_path + " " + args]
+
+        # Verify the python version available on the machine to use
+        self.logger.log("Python version: " + " ".join(sys.version.splitlines()))
+        python_cmd = self.get_python_cmd()
+        if python_cmd == Constants.PYTHON_NOT_FOUND:
+            self.logger.log("Cannot execute patch operation due to [Error={0}]".format(Constants.PYTHON_NOT_FOUND))
```
kjohn-msft, line 72: Cannot execute patch operation due to **error.** [Error={0}]

src/extension/src/ProcessHandler.py
```diff
@@ -55,20 +58,124 @@ def get_env_settings(self, ext_env_handler):
...
         process = subprocess.Popen(command, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         if process.pid is not None:
             self.logger.log("New shell process launched successfully. [Process ID (PID)={0}]".format(str(process.pid)))
+            # Wait for 5 seconds
+            time.sleep(5)
+            # if process is not running, log stdout and stderr
+            if process.poll() is not None:
+                self.logger.log("Process not running for [sequence={0}]".format(seq_no))
+                self.logger.log("Standard output for the inactive process: [Output={0}]".format(str(process.stdout)))
+                self.logger.log("Error logs for the inactive process: [Error={0}]".format(str(process.stderr)))
```
kjohn-msft, line 86: Minor:
Stderr for the inactive process since it's an industry standard term. For consistency, Stdout for line 85.
If more friendly terms are preferred, 'Output' instead of 'Standard output' in 85 and 'Errors reported for_[/by]_.. '.
Whichever you prefer but it's preferrable to stay consistent.

PR #30, opened 2020-08-07
Auto patching: Renaming maxpatchpublishdate to patchrolloutId and wri…

rane-rajasi:
…ting patch metadata summary to health store

### DIFF
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -63,13 +63,13 @@ class Constants(object):
     INSTALLATION = "Installation"
     PATCH_ASSESSMENT_SUMMARY = "PatchAssessmentSummary"
     PATCH_INSTALLATION_SUMMARY = "PatchInstallationSummary"
-    PATCH_METADATA_FOR_HEALTH_STORE = "PatchMetadataForHealthStore"
+    PATCH_METADATA_FOR_HEALTHSTORE = "PatchMetadataForHealthStore"
 
     # patch versions for health store when there is no patch rollout id
     PATCH_VERSION_UNKNOWN = "UNKNOWN"
 
     # wait time after status updates
-    WAIT_FOR_AFTER_STATUS_UPDATE_IN_SECS = 20
+    WAIT_TIME_AFTER_HEALTHSTORE_STATUS_UPDATE_IN_SECS = 20
 
     # Status file states
     STATUS_TRANSITIONING = "Transitioning"
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -65,7 +65,7 @@ class Constants(object):
     PATCH_INSTALLATION_SUMMARY = "PatchInstallationSummary"
     PATCH_METADATA_FOR_HEALTHSTORE = "PatchMetadataForHealthStore"
 
-    # patch versions for health store when there is no patch rollout id
+    # patch versions for healthstore when there is no patch rollout id
     PATCH_VERSION_UNKNOWN = "UNKNOWN"
 
     # wait time after status updates
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -56,13 +56,20 @@ class Constants(object):
         CLASSIFICATIONS_TO_INCLUDE = 'classificationsToInclude'
         PATCHES_TO_INCLUDE = 'patchesToInclude'
         PATCHES_TO_EXCLUDE = 'patchesToExclude'
-        MAX_PATCH_PUBLISH_DATE = 'maxPatchPublishDate'
+        PATCH_ROLLOUT_ID = 'patchRolloutId'
 
     # Operations
     ASSESSMENT = "Assessment"
     INSTALLATION = "Installation"
     PATCH_ASSESSMENT_SUMMARY = "PatchAssessmentSummary"
     PATCH_INSTALLATION_SUMMARY = "PatchInstallationSummary"
+    PATCH_METADATA_FOR_HEALTH_STORE = "PatchMetadataForHealthStore"
+
+    # patch versions for health store when there is no patch rollout id
+    PATCH_VERSION_UNKNOWN = "UNKNOWN"
+
+    # wait time after status updates
+    WAIT_FOR_AFTER_STATUS_UPDATE_IN_SECS = 20
 
     # Status file states
     STATUS_TRANSITIONING = "Transitioning"
--- a/src/core/src/core_logic/PatchInstaller.py
+++ b/src/core/src/core_logic/PatchInstaller.py
@@ -96,15 +96,15 @@ class PatchInstaller(object):
             self.status_handler.set_installation_substatus_json(status=Constants.STATUS_SUCCESS)
             # update patch metadata in status for auto patching request, to be reported to health store
             if self.execution_config.patch_rollout_id is not None:
-                patch_version = None
                 try:
                     patch_version = self.env_layer.datetime.utc_to_standard_datetime(self.execution_config.patch_rollout_id).date()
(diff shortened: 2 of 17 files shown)


src/core/tests/TestCoreMain.py
```diff
@@ -24,36 +24,122 @@
...
+            substatus_file_data = json.load(file_handle)[0]["status"]["substatus"]
+        self.assertEquals(len(substatus_file_data), 3)
+        self.assertTrue(substatus_file_data[0]["name"] == Constants.PATCH_ASSESSMENT_SUMMARY)
+        self.assertTrue(substatus_file_data[0]["status"] == Constants.STATUS_SUCCESS.lower())
+        self.assertTrue(substatus_file_data[1]["name"] == Constants.PATCH_INSTALLATION_SUMMARY)
+        self.assertTrue(substatus_file_data[1]["status"] == Constants.STATUS_SUCCESS.lower())
+        self.assertTrue(substatus_file_data[2]["name"] == Constants.PATCH_METADATA_FOR_HEALTH_STORE)
+        self.assertTrue(substatus_file_data[2]["status"] == Constants.STATUS_SUCCESS.lower())
+        self.assertEqual(json.loads(substatus_file_data[2]["formattedMessage"]["message"])["patchVersion"], Constants.PATCH_VERSION_UNKNOWN)
+        runtime.stop()
```
rane-rajasi, line 122: Had to move runtime init and stop to individual test functions, since every test uses a different patch_rollout_id which has to be set before runtime init

src/core/src/core_logic/PatchInstaller.py
```diff
@@ -94,8 +94,20 @@ def start_installation(self, simulate=False):
         overall_patch_installation_successful = bool(update_run_successful and not maintenance_window_exceeded)
         if overall_patch_installation_successful:
             self.status_handler.set_installation_substatus_json(status=Constants.STATUS_SUCCESS)
+            # update patch metadata in status for auto patching request, to be reported to health store
+            if self.execution_config.patch_rollout_id is not None:
+                patch_version = None
+                try:
+                    patch_version = self.env_layer.datetime.utc_to_standard_datetime(self.execution_config.patch_rollout_id).date()
+                except ValueError as e:
+                    patch_version = Constants.PATCH_VERSION_UNKNOWN
```
kjohn-msft, line 103: The execution cannot continue with this assumption. This will just result in perpetual retries until the problem goes away. If this happens, we need to log, stop and fail (until such time that rollout id changes to more than just a datetime).

In the future, this will be a pass through but there's no legitimate reason this should happen, and no reason to continue from here.

src/core/src/bootstrap/Constants.py
```diff
@@ -56,13 +56,20 @@ class ConfigSettings(EnumBackport):
...
     INSTALLATION = "Installation"
     PATCH_ASSESSMENT_SUMMARY = "PatchAssessmentSummary"
     PATCH_INSTALLATION_SUMMARY = "PatchInstallationSummary"
+    PATCH_METADATA_FOR_HEALTH_STORE = "PatchMetadataForHealthStore"
+
+    # patch versions for health store when there is no patch rollout id
+    PATCH_VERSION_UNKNOWN = "UNKNOWN"
+
+    # wait time after status updates
+    WAIT_FOR_AFTER_STATUS_UPDATE_IN_SECS = 20
```
kjohn-msft, line 72: minor edit for clarity: WAIT_TIME_AFTER_HEALTHSTORE_STATUS_UPDATE_IN_SECS

kjohn-msft: Also reminder: validate wait time with Jemal.

src/core/src/core_logic/PatchInstaller.py
kjohn-msft: Bascially, log error and throw to stop execution.

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -65,6 +70,13 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
 
         self.__current_operation = None
 
+        # Update patch metadata summary in status for auto patching installation requests, to be reported to health store
+        if execution_config.patch_rollout_id is not None and execution_config.operation.lower() == Constants.INSTALLATION.lower():
+            if self.__installation_reboot_status != Constants.RebootStatus.STARTED:
+                self.set_patch_metadata_for_health_store_substatus_json(report_to_health_store=True, wait_after_update=True)
+            # updating metadata summary again with reporting to health store turned off
+            self.set_patch_metadata_for_health_store_substatus_json(report_to_health_store=False, wait_after_update=False)
```
kjohn-msft, line 78: Why isn't this indented like 76?

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -65,6 +70,13 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
 
         self.__current_operation = None
 
+        # Update patch metadata summary in status for auto patching installation requests, to be reported to health store
+        if execution_config.patch_rollout_id is not None and execution_config.operation.lower() == Constants.INSTALLATION.lower():
+            if self.__installation_reboot_status != Constants.RebootStatus.STARTED:
+                self.set_patch_metadata_for_health_store_substatus_json(report_to_health_store=True, wait_after_update=True)
```
kjohn-msft, line 76: Please include an else condition and log why nothing is being written. This is to decipher the flow of the code looking only at logs. Applies in general everywhere.

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -280,9 +292,37 @@ def __new_installation_summary_json(self, installation_packages_json):
...
+        self.__metadata_for_health_store_substatus_json = self.__new_substatus_json_for_operation(Constants.PATCH_METADATA_FOR_HEALTH_STORE, status, code, json.dumps(self.__metadata_for_health_store_summary_json))
+
+        # Update status on disk
+        self.__write_status_file()
+
+        # wait period required in cases where we need to ensure HealthStore reads the status from GA
+        if wait_after_update:
+            time.sleep(Constants.WAIT_FOR_AFTER_STATUS_UPDATE_IN_SECS)
+
+    def __new_patch_metadata_for_health_store_json(self, patch_version=Constants.PATCH_VERSION_UNKNOWN, report_to_health_store=False):
```
kjohn-msft, line 312: healthstore is a single word here and elsewhere. 'health store' to 'healthstore', 'health_store' to 'healtstore' etc.

src/core/tests/TestCoreMain.py
kjohn-msft: Can take this forward to a future review as it is not breaking. Can you please leave the original setup and teardown code commented in place, with the text in your PR comment in this review (as a comment in the code instead at setup())?

src/extension/tests/TestExtConfigSettingsHandler.py
```diff
@@ -240,6 +240,41 @@ def test_read_file_failures(self):
         self.assertRaises(Exception, ext_config_settings_handler.read_file, seq_no)
         shutil.rmtree(test_dir)
 
+    def test_read_all_config_settings_from_file(self):
+        ext_config_settings_handler = ExtConfigSettingsHandler(self.logger, self.json_file_handler, os.path.join(os.path.pardir, "tests", "helpers"))
+        seq_no = ext_config_settings_handler.get_seq_no()
+        config_settings = ext_config_settings_handler.read_file(seq_no)
+
+        # verify operation is read successfully
```
kjohn-msft, line 248: To be pedantic, a successful read validates the value against the expected value. Not none only validates that there is some value present.

src/core/src/bootstrap/Constants.py
rane-rajasi: VM agent sends status blob every 15 seconds. Adding +5 secs to it to make it greater

PR #31, opened 2020-08-11
Auto Patching: first draft of disabling auto OS updates for Ubuntu

rane-rajasi:
ToDo: Adding UTs


src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -359,6 +365,27 @@ def get_package_size(self, output):
...
 
+    # region auto OS updates
+    def get_current_auto_os_update_settings(self):
+        """Get current auto OS updates behaviour set on the machine using the command input"""
+        self.composite_logger.log_debug("Fetching current system settings for auto OS updates")
+        try:
+            return self.invoke_package_manager(self.find_current_auto_os_updates_settings_cmd)
+        except Exception as error:
+            self.composite_logger.log_error("Error fetching current system settings for auto OS updates. [Error={0}]".format(str(error)))
+            return None
```
rane-rajasi, line 376: Do not assume this on behalf of the system. We should fail the operation in this case

src/core/src/bootstrap/ConfigurationFactory.py
```diff
@@ -163,7 +163,7 @@ def new_prod_configuration(self, package_manager_name, package_manager_component
             },
             'package_manager': {
                 'component': package_manager_component,
-                'component_args': ['env_layer', 'composite_logger', 'telemetry_writer', 'status_handler'],
+                'component_args': ['env_layer', 'composite_logger', 'telemetry_writer', 'status_handler', 'execution_config'],
```
kjohn-msft, line 166: Introduce 'execution_config' as the second parameter. The rest of ConfigurationFactory was once refactored to have the most commonly used components always appear at the beginning and in the same order (the only exception was if a certain component wasn't needed). Let's not drift from that for readability.

src/core/src/bootstrap/Constants.py
```diff
@@ -56,7 +56,10 @@ class ConfigSettings(EnumBackport):
         CLASSIFICATIONS_TO_INCLUDE = 'classificationsToInclude'
         PATCHES_TO_INCLUDE = 'patchesToInclude'
         PATCHES_TO_EXCLUDE = 'patchesToExclude'
-        MAX_PATCH_PUBLISH_DATE = 'maxPatchPublishDate'
+        PATCH_ROLLOUT_ID = 'patchRolloutId'
+
+    # File to save default settings for auto OS updates
+    DEFAULT_AUTO_OS_UPDATE_LOG = "DefaultAutoUpdates.log"
```
kjohn-msft, line 62: ImageDefaultPatchMode.bak

kjohn-msft: And IMAGE_DEFAULT_PATCH_MODE_BACKUP

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -25,8 +27,8 @@ class AptitudePackageManager(PackageManager):
     """Implementation of Debian/Ubuntu based package management operations"""
 
     # For more details, try `man apt-get` on any Debian/Ubuntu based box.
-    def __init__(self, env_layer, composite_logger, telemetry_writer, status_handler):
-        super(AptitudePackageManager, self).__init__(env_layer, composite_logger, telemetry_writer, status_handler)
+    def __init__(self, env_layer, composite_logger, telemetry_writer, status_handler, execution_config):
+        super(AptitudePackageManager, self).__init__(env_layer, composite_logger, telemetry_writer, status_handler, execution_config)
```
kjohn-msft, line 31: Reminder to mirror configurationfactory argument ordering after comment there.

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -47,6 +49,10 @@ def __init__(self, env_layer, composite_logger, telemetry_writer, status_handler
         # Package manager exit code(s)
         self.apt_exitcode_ok = 0
 
+        # auto OS updates
+        self.find_current_auto_os_updates_settings_cmd = 'sudo cat /etc/apt/apt.conf.d/20auto-upgrades'
```
kjohn-msft, line 53: Ensure true sample output (taken from a machine; but scrubbed of machine/user identity) of these commands exist as comments inside the test code.

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -359,6 +365,27 @@ def get_package_size(self, output):
...
+        self.composite_logger.log_debug("Fetching current system settings for auto OS updates")
+        try:
+            return self.invoke_package_manager(self.find_current_auto_os_updates_settings_cmd)
+        except Exception as error:
+            self.composite_logger.log_error("Error fetching current system settings for auto OS updates. [Error={0}]".format(str(error)))
+            return None
+
+    def disable_auto_os_update(self):
+        """ Disables auto OS updates on the machine only if they are enabled and logs the default settings the machine comes with """
+        self.log_default_auto_os_updates()
```
kjohn-msft, line 380: backup_image_default_patch_mode - remind me to discuss in the next call.

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -359,6 +365,27 @@ def get_package_size(self, output):
...
+        self.composite_logger.log_debug("Disabling auto OS updates if they are enabled")
+        try:
+            self.invoke_package_manager(self.disable_auto_os_update_command)
+            self.composite_logger.log("Successfully disabled auto OS updates")
+        except Exception as error:
+            self.composite_logger.log_error("Could not disable auto OS updates.")
```
kjohn-msft, line 386: Reminder to bring this discussion up in scrum. Fail fully or not - deliberate choice.

src/core/src/package_managers/PackageManager.py
```diff
@@ -316,6 +320,42 @@ def set_package_manager_setting(self, setting_key, setting_value=""):
         self.package_manager_settings[setting_key] = setting_value
     # endregion
 
+    # region auto OS updates
+    def disable_auto_os_update(self):
```
kjohn-msft, line 324: @abstractmethod

src/core/src/package_managers/PackageManager.py
```diff
@@ -316,6 +320,42 @@ def set_package_manager_setting(self, setting_key, setting_value=""):
...
+                self.composite_logger.log_debug("Extension already has a record of the default system settings for auto OS updates. No need to log the current settings again"
+                                                "[Default Auto OS update settings={0}] [File path={1}]"
+                                                .format(str(default_auto_os_update), self.default_auto_os_update_log_path))
+                return True
+            return False
+        except Exception as error:
+            self.composite_logger.log_error("Unable to read default auto OS update log. [Exception={0}]".format(repr(error)))
+            return False
+
+    def get_current_auto_os_update_settings(self):
```
kjohn-msft, line 355: @abstractmethod

src/core/src/package_managers/PackageManager.py
```diff
@@ -316,6 +320,42 @@ def set_package_manager_setting(self, setting_key, setting_value=""):
...
+        """ Records the default system settings for auto OS updates within patch extension artifacts for future reference.
+        We only log the default setting a VM comes with, any subsequent updates will not be recorded"""
+        if not self.default_auto_os_update_log_exists():
+            current_auto_os_update_settings = self.get_current_auto_os_update_settings()
+            if current_auto_os_update_settings is not None:
+                self.composite_logger.log_debug("Logging default system settings for auto OS updates. [Settings={0}] [Log file path= {1}]"
+                                                .format(str(current_auto_os_update_settings), self.default_auto_os_update_log_path))
+                self.env_layer.file_system.write_with_retry(self.default_auto_os_update_log_path, '[{0}]'.format(current_auto_os_update_settings), mode='w+')
+
+    def default_auto_os_update_log_exists(self):
```
kjohn-msft, line 337: Scan through the code and comments for standard terminology discussed elsewhere in the PR. Can discuss in call.

src/core/tests/TestAptitudePackageManager.py
```diff
@@ -24,10 +24,18 @@ class TestAptitudePackageManager(unittest.TestCase):
     def setUp(self):
         self.runtime = RuntimeCompositor(ArgumentComposer().get_composed_arguments(), True, Constants.APT)
         self.container = self.runtime.container
+        self.valid_auto_os_update_settings = 'APT::Periodic::Update-Package-Lists "1";' + \
```
kjohn-msft, line 27: Updating package lists shouldn't install any packages. It may be beneficial to leave it alone so that reporting within the OS outside of cloud orchestration remains accurate (like listing number of packages available for update when you login).

src/core/tests/TestAptitudePackageManager.py
```diff
@@ -111,5 +119,121 @@ def test_install_package_only_upgrades(self):
...
+        package_manager.disable_auto_os_update()
+        self.assertTrue(package_manager.default_auto_os_update_log_exists())
+        default_auto_os_update = self.runtime.env_layer.file_system.read_with_retry(package_manager.default_auto_os_update_log_path)
+        self.assertTrue(default_auto_os_update is not None)
+        self.assertTrue('APT::Periodic::Update-Package-Lists "1"' in default_auto_os_update)
+        self.assertTrue('APT::Periodic::Unattended-Upgrade "1"' in default_auto_os_update)
+
+    def test_disable_auto_os_update_failure(self):
+        # disable with non existing log file
+        self.runtime.set_legacy_test_type('SadPath')
```
kjohn-msft, line 151: I don't see the specified path used inside legacyenvlayerextensions... reminder to discuss. (Etc. in other tests).

src/core/src/package_managers/PackageManager.py
```diff
@@ -321,37 +321,43 @@ def set_package_manager_setting(self, setting_key, setting_value=""):
     # endregion
 
     # region auto OS updates
+    @abstractmethod
     def disable_auto_os_update(self):
         pass
 
-    def log_default_auto_os_updates(self):
+    def add_image_default_patch_mode_backup(self):
```
kjohn-msft, line 328: backup_image_default_patch_mode

src/core/src/package_managers/PackageManager.py
```diff
@@ -321,37 +321,43 @@ def set_package_manager_setting(self, setting_key, setting_value=""):
...
+                current_auto_os_update_settings = self.get_current_auto_os_update_settings()
                 self.composite_logger.log_debug("Logging default system settings for auto OS updates. [Settings={0}] [Log file path= {1}]"
-                                                .format(str(current_auto_os_update_settings), self.default_auto_os_update_log_path))
-                self.env_layer.file_system.write_with_retry(self.default_auto_os_update_log_path, '[{0}]'.format(current_auto_os_update_settings), mode='w+')
+                                                .format(str(current_auto_os_update_settings), self.image_default_patch_mode_backup_path))
+                self.env_layer.file_system.write_with_retry(self.image_default_patch_mode_backup_path, '[{0}]'.format(current_auto_os_update_settings), mode='w+')
+        except Exception as error:
+            error_message = "Exception during fetching and logging default auto update settings on the machine. [Exception={0}]".format(repr(error))
+            self.composite_logger.log_error(error_message)
+            raise Exception
```
kjohn-msft, line 340: raise or raise specific message

src/core/src/package_managers/PackageManager.py
```diff
@@ -321,37 +321,43 @@ def set_package_manager_setting(self, setting_key, setting_value=""):
...
 
-    def default_auto_os_update_log_exists(self):
+    def image_default_patch_mode_backup_exists(self):
         """ Checks whether default auto OS updates have been recorded earlier within patch extension artifacts """
         self.composite_logger.log_debug("Reading auto OS updates log...")
-        if not os.path.exists(self.default_auto_os_update_log_path) or not os.path.isfile(self.default_auto_os_update_log_path):
+        if not os.path.exists(self.image_default_patch_mode_backup_path) or not os.path.isfile(self.image_default_patch_mode_backup_path):
             self.composite_logger.log_debug("Default system settings for auto OS updates aren't recorded in the extension")
```
kjohn-msft, line 346: Check if anything needs to be done with the config folder in an extension upgrade scenario.

src/core/tests/TestAptitudePackageManager.py
kjohn-msft: TODO: Review again based on disucssion.

src/core/src/package_managers/PackageManager.py
```diff
@@ -316,6 +320,42 @@ def set_package_manager_setting(self, setting_key, setting_value=""):
...
+
+    def default_auto_os_update_log_exists(self):
+        """ Checks whether default auto OS updates have been recorded earlier within patch extension artifacts """
+        self.composite_logger.log_debug("Reading auto OS updates log...")
+        if not os.path.exists(self.default_auto_os_update_log_path) or not os.path.isfile(self.default_auto_os_update_log_path):
+            self.composite_logger.log_debug("Default system settings for auto OS updates aren't recorded in the extension")
+            return False
+        try:
+            default_auto_os_update = self.env_layer.file_system.read_with_retry(self.default_auto_os_update_log_path)
+            if 'Update-Package-Lists' in str(default_auto_os_update) or 'Unattended-Upgrade' in str(default_auto_os_update):
```
rane-rajasi, line 350: This should be in Apt Package Manager

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -359,6 +365,28 @@ def get_package_size(self, output):
...
+            error_msg = "Error fetching current system settings for auto OS updates. [Error={0}]".format(str(error))
+            self.composite_logger.log_error(error_msg)
+            raise Exception(error_msg)
+
+    def disable_auto_os_update(self):
+        """ Disables auto OS updates on the machine only if they are enabled and logs the default settings the machine comes with """
+        try:
+            self.add_image_default_patch_mode_backup()
+            self.composite_logger.log_debug("Disabling auto OS updates if they are enabled")
+            self.invoke_package_manager(self.disable_auto_os_update_command)
```
kjohn-msft, line 384: Manually validate that this setting has taken effect (additional code) + leave a comment why you can't validate from return code or output.

src/core/tests/TestAptitudePackageManager.py
rane-rajasi: As discussed, disabling updating packages as well

src/core/src/core_logic/ExecutionConfig.py
```diff
@@ -51,7 +51,7 @@ def __init__(self, env_layer, composite_logger, execution_parameters):
         self.included_classifications_list = self.__get_execution_configuration_value_safely(self.config_settings, Constants.ConfigSettings.CLASSIFICATIONS_TO_INCLUDE, [])
         self.included_package_name_mask_list = self.__get_execution_configuration_value_safely(self.config_settings, Constants.ConfigSettings.PATCHES_TO_INCLUDE, [])
         self.excluded_package_name_mask_list = self.__get_execution_configuration_value_safely(self.config_settings, Constants.ConfigSettings.PATCHES_TO_EXCLUDE, [])
-        self.patch_rollout_id = self.__get_execution_configuration_value_safely(self.config_settings, Constants.ConfigSettings.PATCH_ROLLOUT_ID, "")
+        self.patch_rollout_id = self.__get_execution_configuration_value_safely(self.config_settings, Constants.ConfigSettings.PATCH_ROLLOUT_ID)
```
kjohn-msft: maintenance_run_id

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -395,18 +396,44 @@ def disable_auto_os_update(self):
...
+                    if self.update_package_list in str(setting):
+                        self.update_package_list_value = re.search(self.update_package_list + ' *"(.*?)".', str(setting)).group(1)
+                    if self.unattended_upgrade in str(setting):
+                        self.unattended_upgrade_value = re.search(self.unattended_upgrade + ' *"(.*?)".', str(setting)).group(1)
+
+                backup_image_default_patch_mode_json = {
+                    self.update_package_list: self.update_package_list_value,
+                    self.unattended_upgrade: self.unattended_upgrade_value
+                }
+                self.composite_logger.log_debug("Logging default system settings for auto OS updates. [Settings={0}] [Log file path= {1}]"
```
rane-rajasi, line 430: Log file path= <- extra space

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -359,6 +367,73 @@ def get_package_size(self, output):
...
+            self.update_image_default_patch_mode(self.unattended_upgrade, "0")
+            self.composite_logger.log("Successfully disabled auto OS updates")
+        except Exception as error:
+            self.composite_logger.log_error("Could not disable auto OS updates. [Error={0}]".format(repr(error)))
+
+    def backup_image_default_patch_mode(self):
+        """ Records the default system settings for auto OS updates within patch extension artifacts for future reference.
+        We only log the default patch_mode a VM comes with, any subsequent updates will not be recorded"""
+        try:
+            if not self.image_default_patch_mode_backup_exists():
```
kjohn-msft, line 386: Include an else block and log that a backup already exists.

src/core/src/package_managers/AptitudePackageManager.py
kjohn-msft: Issue still exists..

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -359,6 +367,73 @@ def get_package_size(self, output):
...
+                image_default_patch_mode = self.env_layer.file_system.read_with_retry(self.image_default_patch_mode_file_path)
+                settings = image_default_patch_mode.strip().split('\n')
+                for patch_mode in settings:
+                    if self.update_package_list in str(patch_mode):
+                        self.update_package_list_value = re.search(self.update_package_list + ' *"(.*?)".', str(patch_mode)).group(1)
+                    if self.unattended_upgrade in str(patch_mode):
+                        self.unattended_upgrade_value = re.search(self.unattended_upgrade + ' *"(.*?)".', str(patch_mode)).group(1)
```
kjohn-msft, line 393: Needs more logging. For any significant line that can fail, logging must exist prior to its execution. The outcome must be recorded after completion. For any code path that will execute only in an if condition, the opposite condition must be logged in an else block to explain execution flow only looking at logs.

This comment generally applies to all code - review for applicability everywhere.

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -359,6 +367,73 @@ def get_package_size(self, output):
...
+                self.env_layer.file_system.write_with_retry(self.image_default_patch_mode_backup_path, '{0}'.format(json.dumps(backup_image_default_patch_mode_json)), mode='w+')
+        except Exception as error:
+            error_message = "Exception during fetching and logging default auto update settings on the machine. [Exception={0}]".format(repr(error))
+            self.composite_logger.log_error(error_message)
+            raise
+
+    def is_image_default_patch_mode_backup_valid(self, image_default_patch_mode_backup):
+        if self.update_package_list in image_default_patch_mode_backup or self.unattended_upgrade in image_default_patch_mode_backup:
+            return True
+        return False
```
kjohn-msft, line 410: Given the way you backup, wouldn't the condition check be 'and' not 'or'? 

Also this is a good example for where logging should be better. Will it be clear why the function evaluated the way it did looking only at logs?

Side note (doesn't apply here once logging comes into the picture): An expression that evaluates directly to boolean can be returned directly:
return <expression>
vs.
if <expression evaluates to True>:
     return True;
return False

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -47,6 +48,13 @@ def __init__(self, env_layer, composite_logger, telemetry_writer, status_handler
         # Package manager exit code(s)
         self.apt_exitcode_ok = 0
 
+        # auto OS updates
+        self.update_package_list = 'APT::Periodic::Update-Package-Lists'
+        self.unattended_upgrade = 'APT::Periodic::Unattended-Upgrade'
+        self.image_default_patch_mode_file_path = '/etc/apt/apt.conf.d/20auto-upgrades'
```
kjohn-msft, line 54: The term image_default is overloaded here and causes confusion when reading code and function names further below.

image_default = configuration the image started off with. The file you're looking at is just os_patch_mode_settings_file_path (emphasis on os and settings for clarity). That file simply happen to have what we consider the image_default settings in the beginning.

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -359,6 +367,73 @@ def get_package_size(self, output):
...
+            error_message = "Exception during fetching and logging default auto update settings on the machine. [Exception={0}]".format(repr(error))
+            self.composite_logger.log_error(error_message)
+            raise
+
+    def is_image_default_patch_mode_backup_valid(self, image_default_patch_mode_backup):
+        if self.update_package_list in image_default_patch_mode_backup or self.unattended_upgrade in image_default_patch_mode_backup:
+            return True
+        return False
+
+    def update_image_default_patch_mode(self, patch_mode, value="0"):
```
kjohn-msft, line 412: update_os_patch_mode_settings

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -359,6 +367,73 @@ def get_package_size(self, output):
...
+        """ Updates (or adds if it doesn't exist) the given patch_mode with the given value in image_default_patch_mode_file """
+        try:
+            # todo: adding space between the patch_mode and value since, we will have to do that if we have to add a patch_mode that did not exist before
+            self.composite_logger.log("Updating default system settings for auto OS updates. [Patch Mode={0}. Value={1}]".format(str(patch_mode), value))
+            image_default_patch_mode = self.env_layer.file_system.read_with_retry(self.image_default_patch_mode_file_path)
```
kjohn-msft, line 417: os_patch_mode_settings instead of image_default_patch_mode

src/core/src/package_managers/AptitudePackageManager.py
kjohn-msft: The confusion with the name as it it now is it made me wonder if you were overwriting the backed up image_default setting. image_default = original settings; not current settings

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -359,6 +367,73 @@ def get_package_size(self, output):
...
+    def is_image_default_patch_mode_backup_valid(self, image_default_patch_mode_backup):
+        if self.update_package_list in image_default_patch_mode_backup or self.unattended_upgrade in image_default_patch_mode_backup:
+            return True
+        return False
+
+    def update_image_default_patch_mode(self, patch_mode, value="0"):
+        """ Updates (or adds if it doesn't exist) the given patch_mode with the given value in image_default_patch_mode_file """
+        try:
+            # todo: adding space between the patch_mode and value since, we will have to do that if we have to add a patch_mode that did not exist before
+            self.composite_logger.log("Updating default system settings for auto OS updates. [Patch Mode={0}. Value={1}]".format(str(patch_mode), value))
```
kjohn-msft, line 416: [Patch Mode={0}] [Value={1}]

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -359,6 +367,73 @@ def get_package_size(self, output):
...
+            self.composite_logger.log_error(error_message)
+            raise
+
+    def is_image_default_patch_mode_backup_valid(self, image_default_patch_mode_backup):
+        if self.update_package_list in image_default_patch_mode_backup or self.unattended_upgrade in image_default_patch_mode_backup:
+            return True
+        return False
+
+    def update_image_default_patch_mode(self, patch_mode, value="0"):
+        """ Updates (or adds if it doesn't exist) the given patch_mode with the given value in image_default_patch_mode_file """
```
kjohn-msft, line 413: Update comment to reference any renamed variables, etc.

src/core/src/package_managers/AptitudePackageManager.py
kjohn-msft: Also the first parameter is not really patch_mode. Patch_mode is only if it is restricted to one of the settings we have as patch_mode across all our components. Please check all variable and function names for clarity - in this case, I'd use something like os_patch_configuration_sub_setting (or something similar).

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -359,6 +367,73 @@ def get_package_size(self, output):
...
+        try:
+            # todo: adding space between the patch_mode and value since, we will have to do that if we have to add a patch_mode that did not exist before
+            self.composite_logger.log("Updating default system settings for auto OS updates. [Patch Mode={0}. Value={1}]".format(str(patch_mode), value))
+            image_default_patch_mode = self.env_layer.file_system.read_with_retry(self.image_default_patch_mode_file_path)
+            patch_mode_to_update = patch_mode + ' "' + value + '";'
+            patch_mode_found_in_file = False
+            updated_patch_mode = ""
+            settings = image_default_patch_mode.strip().split('\n')
+            for i in range(len(settings)):
+                # ToDo: do we need a check for value as well, to check if the value is not the same as the one sent in to replace it, as that will be an unnecessary write
```
kjohn-msft, line 423: Since it's a simple to do, please act on the to do prior to final iteration or just remove it. I'm okay with an unnecessary write given the trivial file size and complexity, and that there should be any increase in probability of consequential failure in such a write.

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -359,6 +367,73 @@ def get_package_size(self, output):
...
+                if patch_mode in settings[i]:
+                    settings[i] = patch_mode_to_update
+                    patch_mode_found_in_file = True
+                updated_patch_mode += settings[i] + "\n"
+            if not patch_mode_found_in_file:
+                updated_patch_mode += patch_mode_to_update + "\n"
+            #ToDo: This adds some whitespace at the beginning of the first line in the settings file which is auto adjusted in the file later, so shouldn't have any issues right now. strip()/lstrip() on the string, does not work, will have to test accross versions and identify teh impact
+            self.env_layer.file_system.write_with_retry(self.image_default_patch_mode_file_path, '{0}'.format(updated_patch_mode.lstrip()), mode='w+')
+        except Exception as error:
+            self.composite_logger.log_error("Error occurred while updating default system settings for auto OS updates. [Patch Mode={0}] [Error={1}]".format(str(patch_mode), repr(error)))
```
kjohn-msft, line 433: "Error occurred while updating default system settings for auto OS updates"
- what does 'default' signify here? It doesn't seem to qualify system settings any further than the only interpretation possible. As I wrote this, it dawned on me that the code is only set up to do that initial disable. Since we know configuration has to come pretty much soon after this, look further ahead if possible now.

src/core/src/package_managers/PackageManager.py
kjohn-msft: Please ensure that the extension upgrade story is covered to retain the image_default configuration.

src/core/tests/library/LegacyEnvLayerExtensions.py
```diff
@@ -466,6 +466,13 @@ def run_command_output(self, cmd, no_output=False, chk_err=True):
                         output = "      bash | 4.3-14ubuntu1.3 | http://us.archive.ubuntu.com/ubuntu xenial-updates/main amd64 Packages\n" + \
                                  "      bash | 4.3-14ubuntu1.2 | http://security.ubuntu.com/ubuntu xenial-security/main amd64 Packages\n" + \
                                  "      bash | 4.3-14ubuntu1 | http://us.archive.ubuntu.com/ubuntu xenial/main amd64 Packages"
+                    elif cmd.find('cat /etc/apt/apt.conf.d/20auto-upgrades') > -1:
+                        code = 0
+                        output = 'APT::Periodic::Update-Package-Lists "1";' + \
+                                 'APT::Periodic::Unattended-Upgrade "1";'
+                    elif cmd.find('''sed -i 's/APT::Periodic::Update-Package-Lists "1"/APT::Periodic::Update-Package-Lists "0"/;s/APT::Periodic::Unattended-Upgrade "1"/APT::Periodic::Unattended-Upgrade "0"/' '''):
+                        code = 0
+                        output = ""
```
kjohn-msft, line 475: This is dead code right?

src/core/tests/library/RuntimeCompositor.py
```diff
@@ -98,3 +98,8 @@ def start_reboot(self, message="Test initiated reboot mock"):
     def mock_sleep(self, seconds):
         pass
 
+    @staticmethod
+    def write_image_default_patch_mode_file(path, data):
+        with open(path, "w+") as file_handle:
+            file_handle.write(data)
```
kjohn-msft, line 104: This is a generic function that writes some data to a file. While you may only be using it a specific way, is this the appropriate name for the function given how it's fully generic? Given what it does, does the function need to exist?

src/extension/src/file_handlers/ExtConfigSettingsHandler.py
```diff
@@ -79,13 +79,16 @@ def read_file(self, seq_no):
                 include_patches = self.get_ext_config_value_safely(config_settings_json, self.public_settings_all_keys.include_patches, raise_if_not_found=False)
                 exclude_patches = self.get_ext_config_value_safely(config_settings_json, self.public_settings_all_keys.exclude_patches, raise_if_not_found=False)
                 internal_settings = self.get_ext_config_value_safely(config_settings_json, self.public_settings_all_keys.internal_settings, raise_if_not_found=False)
-                patch_rollout_id = self.get_ext_config_value_safely(config_settings_json, self.public_settings_all_keys.patch_rollout_id, raise_if_not_found=False)
-
+                maintenance_run_id = self.get_ext_config_value_safely(config_settings_json, self.public_settings_all_keys.maintenance_run_id, raise_if_not_found=False)
+                if maintenance_run_id is None:
+                    maintenance_run_id = self.get_ext_config_value_safely(config_settings_json, self.public_settings_all_keys.patch_rollout_id, raise_if_not_found=False)
+                # read maintenenacerunId first, if that doesn't exist,read patch rollout id and pass it through as maintenance run id
```
kjohn-msft, line 85: Leave the comments currently on 85,86  above line 84 with the same indentation.

src/extension/tests/helpers/test.settings
```diff
@@ -0,0 +1,20 @@
...
+                "operation": "Installation",
+                "activityId": "12345-2312-1234-23245-32112",
+                "startTime": "2019-07-20T12:12:14Z",
+                "maximumDuration": "PT2H",
+                "rebootSetting": "IfRequired",
+                "classificationsToInclude":["Critical","Security"],
+                "patchesToInclude": ["*ern*=1.2*", "kern*=1.23.45"],
+                "patchesToExclude": ["test", "*test"],
+                "internalSettings": "test",
+                "patchRolloutId": "2019-07-22T12:12:14Z"
```
kjohn-msft, line 16: maintenanceRunId

kjohn-msft: This appears to be deliberate -- if so name files with more specificity as to their reason for existence and intended use. back_compat_test.settings?

src/extension/src/ProcessHandler.py
```diff
@@ -83,8 +83,8 @@ def start_daemon(self, seq_no, config_settings, ext_env_handler):
             # if process is not running, log stdout and stderr
             if process.poll() is not None:
                 self.logger.log("Process not running for [sequence={0}]".format(seq_no))
-                self.logger.log("Stdout for the inactive process: [Output={0}]".format(str(process.stdout)))
-                self.logger.log("Stderr for the inactive process: [Error={0}]".format(str(process.stderr)))
+                self.logger.log("Stdout for the inactive process: [Output={0}]".format(str(process.stdout.read())))
+                self.logger.log("Stderr for the inactive process: [Error={0}]".format(str(process.stderr.read())))
```
kjohn-msft, line 87: I’m pretty much convinced the broken pipe error happens because the extension handler captures the stdout and stderr and later goes away while Core continues to write to it.

See here for someone complaining about this:
https://stackoverflow.com/questions/4106565/newbie-python-subprocess-write-error-broken-pipe

This was never a problem in V1 where what is now Core had the exact same behavior (relating to output/errors) but had a long running equivalent process capture that never went away and constantly kept reading from the IO buffers.

Basically this issue needs to be addressed first in the extension handler and to a lesser degree in Core.

We shouldn’t get a sigpipe if we redirect stdout and stderr to devnull when the Core process is created:

with open(os.devnull, 'r+b', 0) as DEVNULL:
    process = Popen(command, stdin=DEVNULL, stdout=DEVNULL, stderr=STDOUT, close_fds=True)

What this does mean for code in lines 86 and 87 is that IF the process doesn’t start correctly, we can run it again in the mode we run it right now and capture what the failure is.

For Core itself, this means that we can doubly guard against this by comment out line 37 in StdOutFileMirror or try to put exception handling around it (line 37) and ignore any exceptions from writing to the “terminal” which is captured by the dead extension handler process.

kjohn-msft (commented):
Comments inline.

src/core/src/package_managers/PackageManager.py
rane-rajasi: Copying image_default backup, CoreState.json and ExtState.json in update handler action. Do we also need to copy the config settings file?

kjohn-msft: The agent is responsible for maintaining the settings file. Is there a reason you think we should copy it?

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: Logging that within image_default_patch_mode_backup_exists(), line 345 in PackageManager.py

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: The word 'default' is misleading here. The function is meant to update any system setting with a given value, which was with the future changes in mind. So the function only writes one setting in the file for every invocation and since the value is passed as an arg, it can be enabled or disabled. Let me know if this answers your concern here, or if there is something I'm missing

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: One question, if the file the settings are being read from is referred to as os_patch_mode_settings_file, then should the individual settings within it be, os_patch_mode_setting or os_patch_mode_sub_setting? This function updates only one setting every invocation in the os_patch_mode_settings_file, so having the function name as update_os_patch_mode_settings and the param name as os_patch_configuration_sub_setting seems like a mismatch.

rane-rajasi: Using os_patch_mode_settings and os_patch_mode_sub_setting for now, let me know if that makes sense

src/core/src/package_managers/PackageManager.py
rane-rajasi: The only one I can think of is, if we are debugging any previous extension runs and want to find out what config settings were sent in for that

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: removed the comment, keeping the extra write

src/core/tests/library/RuntimeCompositor.py
rane-rajasi: This is used in a lot of tests in TestAptitudePackagerManager.py and would be used in the other package manager tests once disable auto OS updates is added for those, which was the reason to add this function here. Updated the name to a more generic one

src/extension/src/ActionHandler.py
```diff
@@ -63,9 +66,41 @@ def update(self):
...
-        # todo: in the test run verify if CoreState.json, ExtState.json and the .status files are deleted, if yes, move them to a separate location
-        self.logger.log("Extension updated")
-        return Constants.ExitCode.Okay
+
+        # config folder path is usually something like: /var/lib/waagent/Microsoft.CPlat.Core.LinuxPatchExtension-<version>/config
+        try:
+            self.logger.log("Extension is being updated to the latest version. Copying the required extension artifacts from previous version to the current one")
+            new_version_config_folder = self.ext_env_handler.config_folder
+            extension_pardir = os.path.abspath(os.path.join(new_version_config_folder, os.path.pardir, os.path.pardir))
+            paths_to_all_versions = self.get_all_versions(extension_pardir)
```
kjohn-msft, line 75: Log what was calculated to be the parent dir and the list of all versions found on the system.

src/extension/src/ProcessHandler.py
rane-rajasi: Will make the changes in the other PR on broken pipe bugfix

src/extension/src/ActionHandler.py
```diff
@@ -63,9 +66,41 @@ def update(self):
...
+            if len(paths_to_all_versions) <= 1:
+                self.logger.log_error("No earlier versions found for the extension")
+                return Constants.ExitCode.HandlerFailed
+
+            self.logger.log("Sorting paths to all version specific extension artifacts on the machine in descending order on version and fetching the immediate previous version path...")
+            paths_to_all_versions.sort(reverse=True)
```
kjohn-msft, line 81: Ensure there are good test for these (since I'm looking at the code in order, please resolve if at least the following work):
- 1.0.0 to 2.0.00
- 1.4.897 to 1.5.23
- 1.5.2 to 1.6.1
- 1.6.99 to 1.6.100
etc.

kjohn-msft: Check with multiple entries in the file system too. This is a critical section of code.

src/extension/src/ActionHandler.py
```diff
@@ -63,9 +66,41 @@ def update(self):
...
+            previous_version_path = paths_to_all_versions[1]
+            if previous_version_path is None or previous_version_path == "" or not os.path.exists(previous_version_path):
```
kjohn-msft, line 83: When any one of a set of conditions trigger an outcome, it must be clear in the logs, which one (or combination) qualified the outcome to occur. Read all code as though you only have the logs to look at to explain an outcome and log as required to help with that.

src/extension/src/ActionHandler.py
```diff
@@ -63,9 +66,41 @@ def update(self):
...
+                self.logger.log_error("Could not find path where previous extension version artifacts are stored. Cannot copy the required artifacts to the latest version")
+                return Constants.ExitCode.HandlerFailed
+
+            self.logger.log("Previous version path. [Path={0}]".format(str(previous_version_path)))
+            for root, dirs, files in os.walk(previous_version_path):
+                for file_name in files:
+                    #ToDo: do we copy .settings file also?
```
kjohn-msft, line 90: As discussed, we don't need to do this.

src/extension/src/ActionHandler.py
```diff
@@ -63,9 +66,41 @@ def update(self):
...
+                    if Constants.EXT_STATE_FILE in file_name or Constants.CORE_STATE_FILE in file_name or ".bak" in file_name:
+                        file_path = os.path.join(root, file_name)
+                        self.logger.log("Copying file. [Source={0}] [Destination={1}]".format(str(file_path), str(new_version_config_folder)))
+                        shutil.copy(file_path, new_version_config_folder)
```
kjohn-msft, line 94: You need exception handling and possibly retries on this. And a plan on what to do if it fails. All of the code in this file lie on a critical path that is much less tolerant to failure than Core.

src/extension/src/ActionHandler.py
```diff
@@ -63,9 +66,41 @@ def update(self):
...
+            previous_version_path = paths_to_all_versions[1]
+            if previous_version_path is None or previous_version_path == "" or not os.path.exists(previous_version_path):
+                self.logger.log_error("Could not find path where previous extension version artifacts are stored. Cannot copy the required artifacts to the latest version")
+                return Constants.ExitCode.HandlerFailed
+
+            self.logger.log("Previous version path. [Path={0}]".format(str(previous_version_path)))
+            for root, dirs, files in os.walk(previous_version_path):
+                for file_name in files:
+                    #ToDo: do we copy .settings file also?
+                    if Constants.EXT_STATE_FILE in file_name or Constants.CORE_STATE_FILE in file_name or ".bak" in file_name:
```
kjohn-msft, line 91: Please log what files we chose not to copy over too, in case a question arises as to why some file that was expected to get copied was missed.

src/extension/src/ActionHandler.py
```diff
@@ -63,9 +66,41 @@ def update(self):
...
+        except Exception as error:
+            self.logger.log_error("Error occurred during extension update. [Error={0}]".format(repr(error)))
+            return Constants.ExitCode.HandlerFailed
+
+    @staticmethod
+    def get_all_versions(extension_pardir):
+        return glob.glob(extension_pardir + '/*LinuxPatchExtension*')
 
     def uninstall(self):
         # ToDo: verify if the agent deletes config files. And find out from the extension/agent team if we need to delete older logs
```
kjohn-msft, line 106: One of two things need to happen in this function:
- Either we don't allow the uninstall to happen (because we know it will never get called for our extension).
- or, we restore the image default backup data into the os settings.

src/core/tests/TestAptitudePackageManager.py
rane-rajasi: Not using this path in per the current implementation

src/extension/src/ActionHandler.py
rane-rajasi: There is an exception handling for the overall update function, where we fail the request if an exception is raised.

src/extension/src/ActionHandler.py
rane-rajasi: uninstall is also called in an extension update scenario.  """ as per the extension user guide, upon update request, Azure agent calls
         1. disable on the prev version
         2. update on the new version
         3. uninstall on the prev version
         4. install (if updateMode is UpdateWithInstall)
         5. enable on the new version
         on uninstall the agent deletes removes configuration files"""
So we we can't disable it and also not restore the os settings

src/core/src/package_managers/PackageManager.py
```diff
@@ -316,6 +321,47 @@ def set_package_manager_setting(self, setting_key, setting_value=""):
...
+        We only log the default system settings a VM comes with, any subsequent updates will not be recorded"""
+        pass
+
+    def image_default_patch_mode_backup_exists(self):
+        """ Checks whether default auto OS update settings have been recorded earlier within patch extension artifacts """
+        self.composite_logger.log_debug("Checking if extension contains a backup for default auto OS updates...")
+        if not os.path.exists(self.image_default_patch_mode_backup_path) or not os.path.isfile(self.image_default_patch_mode_backup_path):
+            self.composite_logger.log_debug("Default system settings for auto OS updates aren't recorded in the extension")
+            return False
+        try:
```
kjohn-msft, line 342: Please ensure readability with appropriate new lines. Relates to an old analogy I shared of the readability of book without paragraphs and instead having one long block of text. Apply the comment everything as needed, not just this particular instance.

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -326,6 +326,23 @@ def get_package_size(self, output):
...
+        """ Disables auto OS updates on the machine only if they are enabled and logs the default settings the machine comes with """
+        pass
+
+    def is_image_default_patch_mode_backup_valid(self, image_default_patch_mode_backup):
+        return True
+
+    def backup_image_default_patch_mode(self):
+        """ Records the default system settings for auto OS updates within patch extension artifacts for future reference.
+        We only log the default system settings a VM comes with, any subsequent updates will not be recorded"""
+        pass
```
kjohn-msft, line 340: Minor comment to think about: Look at this outside of the context you have as the author and me as the reviewer, if someone else reads this function name, would they have reason to believe it will silently not do what it says it will in some cases (it'd be different if it threw an exception)? The improvement possible here would be for the caller to make the determination as to whether to call it in the first place and not encapsulate that in the function itself. Alternatively the function could be named something like backup_image_default_patch_mode_if_not_exists. At your discretion, and food for thought.

src/extension/src/ActionHandler.py
```diff
@@ -63,9 +68,46 @@ def update(self):
...
-        return Constants.ExitCode.Okay
+
+        # config folder path is usually something like: /var/lib/waagent/Microsoft.CPlat.Core.LinuxPatchExtension-<version>/config
+        try:
+            self.logger.log("Extension is being updated to the latest version. Copying the required extension artifacts from previous version to the current one")
+            new_version_config_folder = self.ext_env_handler.config_folder
+            extension_pardir = os.path.abspath(os.path.join(new_version_config_folder, os.path.pardir, os.path.pardir))
+            self.logger.log("Parent directory for all extension version artifacts [Directory={0}]".format(str(extension_pardir)))
+            paths_to_all_versions = self.get_all_versions(extension_pardir)
+            self.logger.log("List of all extension versions found on the machine. [All Versions={0}]".format(paths_to_all_versions))
```
kjohn-msft, line 79: Have to call out -- extra new lines for readability.

kjohn-msft: (Not exactly where my comment landed but in the region as appropriate).

src/extension/src/ActionHandler.py
kjohn-msft: The Todos in this file are >6 months old. Since you're touching this logic and have context (and recent conversations with Laveesh), can you resolve all the todos here? (if there are pieces of information persistently useful, comment them in like you have done in the update function)

kjohn-msft (commented):
Comments inline

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: Using patch_configuration instead of patch_mode for everything on the VM or extension, like discussed. patch_mode to be used for any setting that comes into the extension

src/extension/src/ActionHandler.py
rane-rajasi: Added previous_version_path to the log to help with debugging

src/extension/src/Utility.py
```diff
@@ -39,7 +39,7 @@ def delete_file(self, dir_path, file, raise_if_not_found=True):
                     return True
                 except Exception as e:
                     error_msg = "Trial {0}: Could not delete file. [File={1}] [Exception={2}]".format(retry+1, file, repr(e))
-                    self.logger.log_error(error_msg)
```
kjohn-msft, line 42: Why are we constructing an error message and not logging it? Why is the log statement removed here?

src/extension/src/file_handlers/JsonFileHandler.py
```diff
@@ -74,7 +72,7 @@ def write_to_json_file(self, dir_path, file, content):
                         return
                 except Exception as error:
                     error_message = "Trial {0}: Could not write to file. [File={1}] [Location={2}] [Exception={3}]".format(retry+1, file, str(file_path), error)
-                    self.logger.log_error(error_message)
+
```
kjohn-msft, line 75: Same thing here.

kjohn-msft (changes requested):
Just 2 minor comments

src/extension/src/Utility.py
rane-rajasi: It was adding an error log for each trial attempt, making it look like an actual error in the logs, similar to this:
2020-09-11 19:29:42> "ERROR: Trial 1: Could not read file. [File=0.status] [Location=/var/lib/waagent/Microsoft.CPlat.Core.LinuxPatchExtension-1.6.7/status/0.status] [Exception=IOError(2, 'No such file or directory')]"
2020-09-11 19:29:43> "ERROR: Trial 2: Could not read file. [File=0.status] [Location=/var/lib/waagent/Microsoft.CPlat.Core.LinuxPatchExtension-1.6.7/status/0.status] [Exception=IOError(2, 'No such file or directory')]"

So removed the individual trial logs and keeping a single error log at the end of all attempts. the error_msg set above is used in the final error log below

src/extension/src/file_handlers/JsonFileHandler.py
rane-rajasi: Same reason as mentioned in the other comment

src/extension/src/Utility.py
rane-rajasi: Added the statements back in as warning.

kjohn-msft (approved):

PR #35, opened 2020-08-31
Fix in core for broken pipe error

rane-rajasi:


src/core/src/local_loggers/CompositeLogger.py
```diff
@@ -34,9 +38,17 @@ def __init__(self, file_logger=None, current_env=None):
...
-            print(line)
+        try:
+            message = CompositeLogger.__remove_substring_from_message(message, Constants.ERROR_ADDED_TO_STATUS)
+            for line in message.splitlines():  # allows the extended file logger to strip unnecessary white space
+                sys.stdout.write(line)
+                sys.stdout.flush()
+        except IOError as e:
+            #ToDo: Error Handling??
+            if e.errno == errno.EPIPE:
+                return
```
kjohn-msft, line 49: Since we cannot continue execution without log output working (to avoid blind execution that cannot be inspected later), a viable solution must be implemented to correct the issue. If the attempt to correct it fails (with retries, if retries are predicted to help in any way), execution must still stop.

src/core/src/local_loggers/CompositeLogger.py
```diff
@@ -34,9 +38,17 @@ def __init__(self, file_logger=None, current_env=None):
...
     def log(message):
         """log output"""
-        message = CompositeLogger.__remove_substring_from_message(message, Constants.ERROR_ADDED_TO_STATUS)
-        for line in message.splitlines():  # allows the extended file logger to strip unnecessary white space
-            print(line)
+        try:
+            message = CompositeLogger.__remove_substring_from_message(message, Constants.ERROR_ADDED_TO_STATUS)
+            for line in message.splitlines():  # allows the extended file logger to strip unnecessary white space
+                sys.stdout.write(line)
+                sys.stdout.flush()
```
kjohn-msft, line 45: Please note that print implicitly includes a new line feed at the end of the line, which sys.stdout.write() doesn't (so you'd have to add one). Also is this a confirmed or a hypothesis? I don't expect this to work because:
 - sys.stdout is not really sys.stdout here. It's StdOutFileMirror, which deliberately does nothing in a flush scenario.
 - If sys.stdout was really sys.stdout with a functional flush(), it would only expedite the failure. As noted below, we cannot ignore the error if we don't have a path to recovery.

The exception handler here would not catch anything except a failure to write to the terminal (it's good to understand if this is the actual cause -- simply comment out line 37 in StdOutFileMirror for this and can be the entirety of the fix on a trial basis). 

All errors occurring outside of that line including broken pipe error will be caught in two levels of exception handling in FileLogger and CompositeLogger. And failures will be written to a file with the same name as the log file but with an additional '.failure' extension.

kjohn-msft: **Copying from the other PR since it pertains to this actually:**
I’m pretty much convinced the broken pipe error happens because the extension handler captures the stdout and stderr and later goes away while Core continues to write to it.

See here for someone complaining about this:
https://stackoverflow.com/questions/4106565/newbie-python-subprocess-write-error-broken-pipe

This was never a problem in V1 where what is now Core had the exact same behavior (relating to output/errors) but had a long running equivalent process capture that never went away and constantly kept reading from the IO buffers.

Basically this issue needs to be addressed first in the extension handler and to a lesser degree in Core.

We shouldn’t get a sigpipe if we redirect stdout and stderr to devnull when the Core process is created:

with open(os.devnull, 'r+b', 0) as DEVNULL:
    process = Popen(command, stdin=DEVNULL, stdout=DEVNULL, stderr=STDOUT, close_fds=True)

What this does mean for code in lines 86 and 87 is that IF the process doesn’t start correctly, we can run it again in the mode we run it right now and capture what the failure is.

For Core itself, this means that we can doubly guard against this by comment out line 37 in StdOutFileMirror or try to put exception handling around it (line 37) and ignore any exceptions from writing to the “terminal” which is captured by the dead extension handler process.

rane-rajasi: modified

src/core/src/local_loggers/CompositeLogger.py
```diff
@@ -36,7 +38,7 @@ def log(message):
         """log output"""
         message = CompositeLogger.__remove_substring_from_message(message, Constants.ERROR_ADDED_TO_STATUS)
         for line in message.splitlines():  # allows the extended file logger to strip unnecessary white space
-            print(line)
+            sys.stdout.write(line)
```
kjohn-msft, line 41: As noted in my last comment on this, print automatically inserts a new line character at the end of the line. It also ensures that whatever is passed in is converted to string if required. Sys.stdout.write doesn't. Please ensure the change is equivalent or optionally revert to print.

src/core/src/local_loggers/StdOutFileMirror.py
```diff
@@ -34,7 +34,11 @@ def __init__(self, env_layer, file_logger):
             sys.stdout.write("WARNING: StdOutFileMirror - Skipping as FileLogger is not initialized")
 
     def write(self, message):
-        self.terminal.write(message)  # enable standard job output
+        try:
+            self.terminal.write(message)  # enable standard job output
+        except Exception as error:
+            # Ignore any exceptions from writing to the terminal which is mostly observed to be captured by the dead extension handler process.
```
kjohn-msft, line 40: Use a flag within the class to log to file the first time an error is seen (repeated errors can be ignored as such).

This allows us to be sure this has actually caught the issue, and also help us keep tabs on the health of this overall.

src/extension/src/ProcessHandler.py
```diff
@@ -75,19 +75,26 @@ def start_daemon(self, seq_no, config_settings, ext_env_handler):
...
+                self.logger.log("Stdout for the inactive process: [Output={0}]".format(str(process_reattempt.stdout.read())))
+                self.logger.log("Stderr for the inactive process: [Error={0}]".format(str(process_reattempt.stderr.read())))
+                self.logger.log_error("Error launching process for given sequence. [sequence={0}]".format(seq_no))
                 return
-            return process
-        self.logger.log_error("Error launching process for given sequence. [sequence={0}]".format(seq_no))
+            self.logger.log("New shell process launched successfully in second attempt. [Process ID (PID)={0}]".format(str(process_reattempt.pid)))
+            return process_reattempt
+        self.logger.log("New shell process launched successfully in first attempt. [Process ID (PID)={0}]".format(str(process.pid)))
+        return process
```
kjohn-msft, line 97: Please include new lines between logical blocks of code for readability.

src/core/src/local_loggers/StdOutFileMirror.py
kjohn-msft: Please simulate this failure (after addressing the first comment about additional logging) and validate -- insert an long-running loop outputting a lot of text for a long duration of time early in Core's execution. Invoke everything in test as you normally would and let it run in the background. Please review the logs for the expected outcome and to see the mitigation work, and confirm here please. 

It's possible to be know with little overhead if this addresses the full extent of the issue before deploying, and is cheaper overall.

src/tools/Package-Core.py
```diff
@@ -187,10 +222,15 @@ def main(argv):
...
+
     except Exception as error:
-        print('Exception during merge python modules: ' + repr(error))
+        print('Exception during packaging all python modules in core: ' + repr(error))
         raise
 
 
 if __name__ == "__main__":
-    main(sys.argv)
+    main(sys.argv)
```
kjohn-msft, line 236: What changed here? Ensure file ends with a new line. The IDE should have complained about this.

kjohn-msft (commented):
Comments inline...

kjohn-msft (commented):
Comments inline.

src/core/src/local_loggers/StdOutFileMirror.py
rane-rajasi: Tested this locally with a long running loop at the beginning of PatchInstaller. 

- The core was triggered by the first method (where errors are sent to devnull) and handler goes away. 

- Core stops when it's trying to print a long text and it doesn't go into the exception handling part.

- Also, tried everything with using subprocess.PIPE for error logging while launching core the first time. The process after the exeception is not consistent. It intermittently goes to the exception handling part and completes the remaining operations but there are instances when it either stops even before exception handler (i.e. does not write the error in core log file) or it writes the error to the log file and then stops, without completing the remaining steps.

So in both the cases (logging errors in devnull or subprocess.PIPE), core usually stops executing after the error. I'm thinking of limiting the # of characters being written in one print statement. If the text exceeds this limit, we can split the message into multiple lines. Let me know what you think

kjohn-msft: Comment out self.terminal.write and keep the test running again - let me know if the forced repro's behavior changes in any way. The outcome of this can inform what needs to happen next.

kjohn-msft (commented):
Please close this PR as the other one superseded it.

rane-rajasi:
Resolved issue in another PR

PR #39, opened 2020-09-24
Adding a check to allow only certain operations and block all others

rane-rajasi:


src/extension/src/EnableCommandHandler.py
```diff
@@ -40,12 +40,19 @@ def execute_handler_action(self):
         """ Responsible for taking appropriate action for enable command as per the request sent in Handler Configuration file by user """
         try:
             config_settings = self.ext_config_settings_handler.read_file(self.seq_no)
+            operation = config_settings.__getattribute__(self.config_public_settings.operation)
+
+            # Allow only certain operations
+            if operation not in [Constants.NOOPERATION, Constants.ASSESSMENT, Constants.INSTALLATION]:
+                self.logger.log_error("Requested operation is not supported by the extension")
+                exit(Constants.ExitCode.InvalidConfigSettingPropertyValue)
```
kjohn-msft, line 48: If it's a configurePatching call, just persist the sequence number as something you've processed and return 0. Don't let the agent think there is a problem and cause it to retry a configurePatching request that may come in.

The actual configurePatching code can come later.

kjohn-msft (commented):
Minor comment about configurepatching. Just additional code; everything you have now is fine.

src/extension/src/EnableCommandHandler.py
rane-rajasi: Logging the sequence number for configure patching requests

kjohn-msft (approved):

PR #43, opened 2020-10-01
Modifying how seq no is read for different handler commands

rane-rajasi:
Includes:

- Individual log files for every handler action/command
- Increased the log retention limit to accommodate the separate log files 

For Enable command:

- Reading seq no first from environment variable and validating if the corresponding .settings file exists. Failing if file not found
- If seq no is not set in environment variable, try to fetch it from the config settings files

For Non-Enable command:

- Fetch seq no from environment variable and validate if the corresponding .settings file exists, fail if not found.

- If not found, return None. It is upto individual commands to decide what to do here

Pending:

- Writing to status file for non-enable commands

### DIFF
--- a/src/extension/src/ActionHandler.py
+++ b/src/extension/src/ActionHandler.py
@@ -24,11 +24,12 @@ from distutils.version import LooseVersion
 from extension.src.Constants import Constants
 from extension.src.EnableCommandHandler import EnableCommandHandler
 from extension.src.InstallCommandHandler import InstallCommandHandler
+from extension.src.local_loggers.StdOutFileMirror import StdOutFileMirror
 
 
 class ActionHandler(object):
     """Responsible for identifying the action to perform based on the user input"""
-    def __init__(self, logger, utility, runtime_context_handler, json_file_handler, ext_env_handler, ext_config_settings_handler, core_state_handler, ext_state_handler, ext_output_status_handler, process_handler, cmd_exec_start_time, seq_no):
+    def __init__(self, logger, utility, runtime_context_handler, json_file_handler, ext_env_handler, ext_config_settings_handler, core_state_handler, ext_state_handler, ext_output_status_handler, process_handler, cmd_exec_start_time):
         self.logger = logger
         self.utility = utility
         self.runtime_context_handler = runtime_context_handler
@@ -40,7 +41,6 @@ class ActionHandler(object):
         self.ext_output_status_handler = ext_output_status_handler
         self.process_handler = process_handler
         self.cmd_exec_start_time = cmd_exec_start_time
-        self.seq_no = seq_no
 
     def determine_operation(self, command):
         switcher = {
@@ -57,9 +57,28 @@ class ActionHandler(object):
             raise e
 
     def install(self):
-        self.logger.log("Extension installation started")
-        install_command_handler = InstallCommandHandler(self.logger, self.ext_env_handler)
-        return install_command_handler.execute_handler_action()
+        stdout_file_mirror = None
+        file_logger = None
+
+        try:
+            log_file_name = datetime.datetime.utcnow().strftime("%Y-%m-%d_%H-%M-%S") + "_Install"
+            file_logger = self.utility.create_log_file(self.ext_env_handler.log_folder, log_file_name)
+            if file_logger is not None:
+                self.logger.file_logger = file_logger
+                stdout_file_mirror = StdOutFileMirror(file_logger)
+            self.logger.log("Extension installation started")
+            install_command_handler = InstallCommandHandler(self.logger, self.ext_env_handler)
+            return install_command_handler.execute_handler_action()
+
+        except Exception as error:
+            self.logger.log_error("Error occurred during extension install. [Error={0}]".format(repr(error)))
+            return Constants.ExitCode.HandlerFailed
+
+        finally:
+            if stdout_file_mirror is not None:
+                stdout_file_mirror.stop()
+            if file_logger is not None:
+                file_logger.close()
 
     def update(self):
         """ as per the extension user guide, upon update request, Azure agent calls
@@ -71,8 +90,19 @@ class ActionHandler(object):
          on uninstall the agent deletes removes configuration files"""
 
         # config folder path is usually something like: /var/lib/waagent/Microsoft.CPlat.Core.LinuxPatchExtension-<version>/config
+        stdout_file_mirror = None
+        file_logger = None
         try:
+            log_file_name = datetime.datetime.utcnow().strftime("%Y-%m-%d_%H-%M-%S") + "_Update"
+            file_logger = self.utility.create_log_file(self.ext_env_handler.log_folder, log_file_name)
+            if file_logger is not None:
+                self.logger.file_logger = file_logger
+                stdout_file_mirror = StdOutFileMirror(file_logger)
+
             self.logger.log("Extension is being updated to the latest version. Copying the required extension artifacts from preceding version to the current one")
+
+            seq_no = self.ext_config_settings_handler.get_seq_no(is_enable_request=False)
+            #toDo: use seq_no to create status file to log errors
 
             # fetch all earlier extension versions available on the machine
             new_version_config_folder = self.ext_env_handler.config_folder
@@ -102,9 +132,16 @@ class ActionHandler(object):
 
             self.logger.log("All update actions from extension handler completed.")
             return Constants.ExitCode.Okay
+
         except Exception as error:
             self.logger.log_error("Error occurred during extension update. [Error={0}]".format(repr(error)))
             return Constants.ExitCode.HandlerFailed
+
+        finally:
+            if stdout_file_mirror is not None:
+                stdout_file_mirror.stop()
+            if file_logger is not None:
+                file_logger.close()
 
     @staticmethod
     def get_all_versions(extension_pardir):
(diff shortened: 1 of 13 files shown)


src/extension/src/ActionHandler.py
```diff
@@ -57,9 +57,28 @@ def determine_operation(self, command):
...
-        return install_command_handler.execute_handler_action()
+        stdout_file_mirror = None
+        file_logger = None
+
+        try:
+            log_file_name = datetime.datetime.utcnow().strftime("%Y-%m-%d_%H-%M-%S") + "_Install"
+            file_logger = self.utility.create_log_file(self.ext_env_handler.log_folder, log_file_name)
+            if file_logger is not None:
+                self.logger.file_logger = file_logger
+                stdout_file_mirror = StdOutFileMirror(file_logger)
```
kjohn-msft, line 68: This chunk of code has been duplicated several times here. Abstract into a function  and reference from all sections that use the exact same pattern.

src/extension/src/ActionHandler.py
```diff
@@ -71,9 +90,20 @@ def update(self):
...
+
             self.logger.log("Extension is being updated to the latest version. Copying the required extension artifacts from preceding version to the current one")
 
+            seq_no = self.ext_config_settings_handler.get_seq_no(is_enable_request=False)
+            #toDo: use seq_no to create status file to log errors
+
```
kjohn-msft, line 106: Break this out into a second review (branch from this branch if required for work in progress). Makes it easier to catch if there are any oversights by keeping each review (+ iterations) simple.

src/extension/src/ActionHandler.py
```diff
@@ -143,23 +180,97 @@ def copy_config_files(self, src, dst, raise_if_not_copied=False):
...
 
     def enable(self):
-        self.logger.log("Enable triggered on extension")
-        enable_command_handler = EnableCommandHandler(self.logger, self.utility, self.runtime_context_handler, self.ext_env_handler, self.ext_config_settings_handler, self.core_state_handler, self.ext_state_handler, self.ext_output_status_handler, self.process_handler, self.cmd_exec_start_time, self.seq_no)
-        return enable_command_handler.execute_handler_action()
+        stdout_file_mirror = None
+        file_logger = None
+        try:
+            log_file_name = datetime.datetime.utcnow().strftime("%Y-%m-%d_%H-%M-%S") + "_Enable"
+            file_logger = self.utility.create_log_file(self.ext_env_handler.log_folder, log_file_name)
```
kjohn-msft, line 209: Considering that enables can happen more frequently even without any action from CRP, consider if just for enable, using this approach alone will result in other logs getting groomed out when the count increases. As-is is okay for this review, but let's discuss next week if this can be improved on. Develop your thoughts on this and plan on taking a direction here in a separate PR next week.

kjohn-msft: Consider grooming Core logs separately from ext logs. Overflowing ext logs should not groom out core logs.

src/extension/src/file_handlers/ExtConfigSettingsHandler.py
```diff
@@ -33,37 +33,70 @@ def __init__(self, logger, json_file_handler, config_folder):
...
+                            freshest_time = os.path.getmtime(os.path.join(self.config_folder, file))
+                            seq_no = cur_seq_no
+                        else:
+                            current_file_m_time = os.path.getmtime(os.path.join(self.config_folder, file))
+                            if current_file_m_time > freshest_time:
+                                freshest_time = current_file_m_time
+                                seq_no = cur_seq_no
+                except ValueError:
+                    continue
+        return seq_no
```
kjohn-msft, line 98: Please add logging - the goal always is to explain what the function did (and experienced in the environment) looking exclusively at logs.

src/extension/src/file_handlers/ExtConfigSettingsHandler.py
```diff
@@ -33,37 +33,70 @@ def __init__(self, logger, json_file_handler, config_folder):
...
-                                if current_file_m_time > freshest_time:
-                                    freshest_time = current_file_m_time
-                                    seq_no = cur_seq_no
-                    except ValueError:
-                        continue
-            return seq_no
+            # get seq no from settings file
+            return self.__get_seq_no_from_config_settings()
+
         except Exception as error:
```
kjohn-msft, line 56: Why is the actual exception being swallowed?

src/extension/src/file_handlers/ExtConfigSettingsHandler.py
```diff
@@ -33,37 +33,70 @@ def __init__(self, logger, json_file_handler, config_folder):
...
-                                current_file_m_time = os.path.getmtime(os.path.join(self.config_folder, file))
-                                if current_file_m_time > freshest_time:
-                                    freshest_time = current_file_m_time
-                                    seq_no = cur_seq_no
-                    except ValueError:
-                        continue
-            return seq_no
+            # get seq no from settings file
+            return self.__get_seq_no_from_config_settings()
+
```
kjohn-msft, line 55: This line should be under line 46 for readability.

kjohn-msft (commented):
Comments inline

src/extension/src/file_handlers/ExtConfigSettingsHandler.py
```diff
@@ -33,37 +33,70 @@ def __init__(self, logger, json_file_handler, config_folder):
...
+    def get_seq_no(self, is_enable_request=False):
         """ Fetches sequence number, initially from the env variable. If nothing is set in env variable then fetches from config folder based on timestamp, since GA updates the settings file before calling a command """
         try:
-            seq_no = os.getenv(Constants.SEQ_NO_ENVIRONMENT_VAR)
-            if seq_no is not None:
+            # get seq no from env var
+            seq_no = self.__get_seq_no_from_env_var()
+            if seq_no == -1:
+                self.logger.log_error("Error occurred while verifying sequence number between environment variable and configuration settings")
+                return None
```
kjohn-msft, line 43: __get_seq_no_from_env_var is doing more than its name suggests (the name is good, but breaks the single responsibility principle). The check to see if the seq no. exists on the file system after being discovered there should happen outside that function. get_seq_no should take on additional responsibility appropriately.

src/extension/src/file_handlers/ExtOutputStatusHandler.py
```diff
@@ -62,17 +59,18 @@ def __init__(self, logger, utility, json_file_handler, log_file_path, seq_no, di
         self.__current_operation = None
 
         # Load the currently persisted status file into memory
-        self.read_file()
-
-    def write_status_file(self, operation, status=Constants.Status.Transitioning.lower()):
-        self.logger.log("Writing status file to provide patch management data for [Sequence={0}]".format(str(self.__seq_no)))
+        #ToDo: move it to some other location, since seq no is not available at load
+        # self.read_file()
```
kjohn-msft, line 63: This is a bit worrisome. Please document the design for verification including all expected state transitions and the expected behavior of the extension in all those transitions (including benign re-enables, legitmate reboots, etc.). The code can then be double-checked against that.

kjohn-msft: To speed up each review, please keep individual review complexity low - writing to the status for non-enable operations is not integral to this review. (As in the additional ask to split out log grooming between ext. and core, to avoid loss of important logs when benign 'enables' overflow)

src/extension/src/ActionHandler.py
```diff
@@ -196,21 +192,14 @@ def uninstall(self):
...
         file_logger = None
-        try:
-            log_file_name = datetime.datetime.utcnow().strftime("%Y-%m-%d_%H-%M-%S") + "_Enable"
-            file_logger = self.utility.create_log_file(self.ext_env_handler.log_folder, log_file_name)
-            if file_logger is not None:
-                self.logger.file_logger = file_logger
-                stdout_file_mirror = StdOutFileMirror(file_logger)
 
+        try:
+            file_logger, stdout_file_mirror = self.setup_file_logger(action="Enable")
```
kjohn-msft, line 202: To be discussed this week as in the outdated comment. Please set up half an hour to discuss with your proposal on how to mitigate log overflows from repeated re-enables (on the same sequence number).

src/extension/src/ActionHandler.py
```diff
@@ -219,22 +208,14 @@ def enable(self):
...
         file_logger = None
 
         try:
-            log_file_name = datetime.datetime.utcnow().strftime("%Y-%m-%d_%H-%M-%S") + "_Disable"
-            file_logger = self.utility.create_log_file(self.ext_env_handler.log_folder, log_file_name)
-            if file_logger is not None:
-                self.logger.file_logger = file_logger
-                stdout_file_mirror = StdOutFileMirror(file_logger)
-
+            file_logger, stdout_file_mirror = self.setup_file_logger(action="Disable")
```
kjohn-msft, line 218: Strings serving functional purposes in code should ideally be mapped to constants.

kjohn-msft: Applies to every action - not leaving comments for each.

kjohn-msft (commented):
Comments inline. Log overflow is the only concern on re-enable. It can be put in the immediate next review but we should not deploy without that as we risk losing logs that may be critical to explaining outcomes in the current absence of telemetry.

src/extension/src/file_handlers/ExtOutputStatusHandler.py
rane-rajasi: Will include this with the next PR on writing to status file

kjohn-msft (approved):

PR #45, opened 2020-11-13
Bug fix for status handler writes

kjohn-msft:


src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -116,6 +116,15 @@ def set_package_assessment_status(self, package_names, package_versions, classif
     def set_package_install_status(self, package_names, package_versions, status="Pending", classification=None):
         """ Externally available method to set installation status for one or more packages of the **SAME classification and status** """
         self.composite_logger.log_debug("Setting package installation status in bulk. [Count={0}]".format(str(len(package_names))))
+
+        # Data normalization and corruption guards - if these exceptions hit, a bug has been introduced elsewhere
+        if isinstance(package_names, str) != isinstance(package_versions, str):
+            raise Exception("Internal error: Package name and version data corruption detected.")
+        if isinstance(package_names, str):
+            package_names, package_versions = [package_names], [package_versions]
```
mirichmo, line 124: This doesn't seem quite right to me. If both of them are not strings, isn't that a data corruption scenario too?

I interpret line 121 as, "both package_names and package_versions must be either strings or not strings."

mirichmo (changes requested):

src/core/src/service_interfaces/StatusHandler.py
kjohn-msft: Yes both must either be strings or lists. One cannot be a string while the other is a list.

mirichmo (approved):

PR #47, opened 2021-01-08
Adding telemetry to extension handler and core

rane-rajasi:
Added telemetry only in extension handler to review the approach. Once the approach is decided upon, will add telemetry to extension core.

2. Event is a json structure

ToDo:
- [x] Have to add more unit test for writing multiple events in one file 
- [X] Deleting older event files in case directory limit is reached
- [x] Unit test for writing to new event file incase event file size limit is reached
- [x] Changes to ensure we use the telemetry preview version provided by Guest Agent

Will be added in a different PR: 
Make sure all the events are added to < 300 json files, as agent only reads a max of 300 event jsons at a time and ignores and deletes the rest

### DIFF
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -183,7 +183,7 @@ class Constants(object):
         Informational = "Informational"
         LogAlways = "LogAlways"
 
-    DATE_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
+    UTC_DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
 
     # EnvLayer Constants
     class EnvLayer(EnumBackport):
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -171,6 +171,8 @@ class Constants(object):
     TELEMETRY_INFO = "Info"
     TELEMETRY_DEBUG = "Debug"
 
+    DATE_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
+
     # EnvLayer Constants
     class EnvLayer(EnumBackport):
         PRIVILEGED_OP_MARKER = "Privileged_Op_e6df678d-d09b-436a-a08a-65f2f70a6798"
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -29,6 +29,9 @@ class Constants(object):
     GLOBAL_EXCLUSION_LIST = ""   # if a package needs to be blocked across all of Azure
     UNKNOWN = "Unknown"
 
+    # Extension version (todo: move to a different file)
+    EXT_VERSION = "1.6.12"
+
     # Runtime environments
     TEST = 'Test'
     DEV = 'Dev'
@@ -47,6 +50,7 @@ class Constants(object):
         LOG_FOLDER = "logFolder"
         CONFIG_FOLDER = "configFolder"
         STATUS_FOLDER = "statusFolder"
+        EVENTS_FOLDER = "eventsFolder"
 
     class ConfigSettings(EnumBackport):
         OPERATION = 'operation'
@@ -163,13 +167,21 @@ class Constants(object):
 
     ERROR_ADDED_TO_STATUS = "Error_added_to_status"
 
-    # Telemetry Categories
-    TELEMETRY_OPERATION_STATE = "State"
-    TELEMETRY_CONFIG = "Config"
-    TELEMETRY_PACKAGE = "PackageInfo"
-    TELEMETRY_ERROR = "Error"
-    TELEMETRY_INFO = "Info"
-    TELEMETRY_DEBUG = "Debug"
+    # Telemetry Settings
+    TELEMETRY_MSG_SIZE_LIMIT_IN_BYTES = 3072  # 3KB
+    TELEMETRY_EVENT_SIZE_LIMIT_IN_BYTES = 6144  # 6KB
+    TELEMETRY_EVENT_FILE_SIZE_LIMIT_IN_BYTES = 4194304  # 4MB
+    TELEMETRY_DIR_SIZE_LIMIT_IN_BYTES = 41943040  # 40MB
+    TELEMETRY_BUFFER_FOR_DROPPED_COUNT_MSG_IN_BYTES = 80  # buffer for the bytes dropped text added at the end of the truncated telemetry message
+
+    # Telemetry Event Level
+    class TelemetryEventLevel(EnumBackport):
+        Critical = "Critical"
+        Error = "Error"
+        Warning = "Warning"
+        Verbose = "Verbose"
+        Informational = "Informational"
+        LogAlways = "LogAlways"
 
     DATE_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
 
--- a/src/core/src/bootstrap/EnvLayer.py
+++ b/src/core/src/bootstrap/EnvLayer.py
@@ -419,7 +419,7 @@ class EnvLayer(object):
 
         try:
             record = {
-                "timestamp": str(timestamp) if timestamp is not None else datetime.datetime.strptime(str(datetime.datetime.utcnow()).split(".")[0], "%Y-%m-%dT%H:%M:%SZ"), #WRONG
+                "timestamp": str(timestamp) if timestamp is not None else datetime.datetime.strptime(str(datetime.datetime.utcnow()).split(".")[0], Constants.DATE_FORMAT), #WRONG
                 "operation": str(operation),
                 "code": int(code),
                 "output": base64.b64encode(str(output)),
--- a/src/core/src/bootstrap/EnvLayer.py
+++ b/src/core/src/bootstrap/EnvLayer.py
@@ -419,7 +419,7 @@ class EnvLayer(object):
 
         try:
             record = {
-                "timestamp": str(timestamp) if timestamp is not None else datetime.datetime.strptime(str(datetime.datetime.utcnow()).split(".")[0], Constants.DATE_FORMAT), #WRONG
+                "timestamp": str(timestamp) if timestamp is not None else datetime.datetime.strptime(str(datetime.datetime.utcnow()).split(".")[0], Constants.UTC_DATETIME_FORMAT), #WRONG
                 "operation": str(operation),
                 "code": int(code),
                 "output": base64.b64encode(str(output)),
--- a/src/core/src/core_logic/ExecutionConfig.py
+++ b/src/core/src/core_logic/ExecutionConfig.py
@@ -41,6 +41,7 @@ class ExecutionConfig(object):
         self.log_folder = self.environment_settings[Constants.EnvSettings.LOG_FOLDER]
         self.config_folder = self.environment_settings[Constants.EnvSettings.CONFIG_FOLDER]
         self.status_folder = self.environment_settings[Constants.EnvSettings.STATUS_FOLDER]
+        self.events_folder = self.environment_settings[Constants.EnvSettings.EVENTS_FOLDER]
 
         # Config Settings
         self.composite_logger.log_debug(" - Parsing configuration settings... [ConfigSettings={0}]".format(str(self.config_settings)))
@@ -58,6 +59,8 @@ class ExecutionConfig(object):
             self.reboot_setting = self.__get_execution_configuration_value_safely(self.config_settings, Constants.ConfigSettings.REBOOT_SETTING, Constants.REBOOT_NEVER)     # safe extension-level default
 
         # Derived Settings
+        self.composite_logger.telemetry_writer.events_folder_path = self.events_folder
+        self.composite_logger.telemetry_writer.set_operation_id(self.activity_id)
         self.composite_logger.log_debug(" - Establishing data publishing paths...")
         self.log_file_path = os.path.join(self.log_folder, str(self.sequence_number) + ".core.log")
         self.composite_logger.log_debug("  -- Core log: " + str(self.log_file_path))
--- a/src/core/src/local_loggers/CompositeLogger.py
+++ b/src/core/src/local_loggers/CompositeLogger.py
@@ -22,20 +22,26 @@ from core.src.bootstrap.Constants import Constants
 class CompositeLogger(object):
     """ Manages diverting different kinds of output to the right sinks for them. """
 
-    def __init__(self, env_layer=None, file_logger=None, current_env=None):
+    def __init__(self, env_layer=None, file_logger=None, telemetry_writer=None, current_env=None):
         self.env_layer = env_layer
         self.file_logger = file_logger
+        self.telemetry_writer = telemetry_writer
         self.ERROR = "ERROR:"
         self.WARNING = "WARNING:"
         self.DEBUG = "DEBUG:"
         self.VERBOSE = "VERBOSE:"
         self.current_env = current_env
         self.NEWLINE_REPLACE_CHAR = " "
+        self.telemetry_task = "ExtensionCoreLog"
+        self.telemetry_msg_truncated_note = " This message will be truncated in telemetry as it exceeds the size limit."
 
-    def log(self, message):
+    def log(self, message, message_type=Constants.TelemetryEventLevel.Informational):
         """log output"""
         message = self.__remove_substring_from_message(message, Constants.ERROR_ADDED_TO_STATUS)
         message = message.strip()
+        if self.telemetry_writer is not None and self.telemetry_writer.events_folder_path is not None:
+            self.telemetry_writer.write_event(self.telemetry_task, message, message_type)
+            message = self.__add_telemetry_error_note(message)
         if self.current_env in (Constants.DEV, Constants.TEST):
             for line in message.splitlines():  # allows the extended file logger to strip unnecessary white space
                 print(line)
@@ -47,26 +53,34 @@ class CompositeLogger(object):
         """log errors"""
         message = self.__remove_substring_from_message(message, Constants.ERROR_ADDED_TO_STATUS)
         message = self.ERROR + (self.NEWLINE_REPLACE_CHAR.join(message.split(os.linesep))).strip()
-        self.log(message)
+        self.log(message, message_type=Constants.TelemetryEventLevel.Error)
 
     def log_warning(self, message):
         """log warning"""
         message = self.__remove_substring_from_message(message, Constants.ERROR_ADDED_TO_STATUS)
         message = self.WARNING + (self.NEWLINE_REPLACE_CHAR.join(message.split(os.linesep))).strip()
-        self.log(message)
+        self.log(message, message_type=Constants.TelemetryEventLevel.Warning)
 
     def log_debug(self, message):
         """log debug"""
         message = self.__remove_substring_from_message(message, Constants.ERROR_ADDED_TO_STATUS)
         message = message.strip()
+        if self.telemetry_writer is not None and self.telemetry_writer.events_folder_path is not None and self.current_env not in (Constants.DEV, Constants.TEST):
+            self.telemetry_writer.write_event(self.telemetry_task, message, Constants.TelemetryEventLevel.Verbose)
+            message = self.__add_telemetry_error_note(message)
+
         if self.current_env in (Constants.DEV, Constants.TEST):
-            self.log(self.current_env + ": " + message)  # send to standard output if dev or test env
+            self.log(self.current_env + ": " + message, Constants.TelemetryEventLevel.Verbose)  # send to standard output if dev or test env
         elif self.file_logger is not None:
             self.file_logger.write("\n\t" + self.DEBUG + " " + "\n\t".join(message.splitlines()).strip())
 
     def log_verbose(self, message):
         """log verbose"""
         message = self.__remove_substring_from_message(message, Constants.ERROR_ADDED_TO_STATUS)
+        if self.telemetry_writer is not None and self.telemetry_writer.events_folder_path is not None:
+            self.telemetry_writer.write_event(self.telemetry_task, message, Constants.TelemetryEventLevel.Verbose)
+            message = self.__add_telemetry_error_note(message)
+
         if self.file_logger is not None:
             self.file_logger.write("\n\t" + self.VERBOSE + " " + "\n\t".join(message.strip().splitlines()).strip())
 
@@ -75,5 +89,11 @@ class CompositeLogger(object):
         """Remove substring from a string"""
         if substring in message:
             message = message.replace("[{0}]".format(Constants.ERROR_ADDED_TO_STATUS), "")
+        return message
+
+    def __add_telemetry_error_note(self, message):
+        """ Adding telemetry error messages to the original message, so as to log telemetry errors in other logs """
+        if len(message.encode('utf-8')) > Constants.TELEMETRY_MSG_SIZE_LIMIT_IN_BYTES:
+            message = message + self.telemetry_msg_truncated_note
         return message
 
--- a/src/core/src/package_managers/AptitudePackageManager.py
+++ b/src/core/src/package_managers/AptitudePackageManager.py
@@ -26,8 +26,8 @@ class AptitudePackageManager(PackageManager):
     """Implementation of Debian/Ubuntu based package management operations"""
 
     # For more details, try `man apt-get` on any Debian/Ubuntu based box.
-    def __init__(self, env_layer, execution_config, composite_logger, telemetry_writer, status_handler):
-        super(AptitudePackageManager, self).__init__(env_layer, execution_config, composite_logger, telemetry_writer, status_handler)
+    def __init__(self, env_layer, execution_config, composite_logger, status_handler):
+        super(AptitudePackageManager, self).__init__(env_layer, execution_config, composite_logger, status_handler)
         # Repo refresh
         self.repo_refresh = 'sudo apt-get -q update'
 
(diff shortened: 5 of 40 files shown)


src/extension/src/Constants.py
```diff
@@ -55,14 +58,20 @@ def __iter__(self):
...
-    TelemetryConfig = "Config"
-    TelemetryError = "Error"
-    TelemetryWarning = "Warning"
-    TelemetryInfo = "Info"
-    TelemetryDebug = "Debug"
+    # Telemetry Settings
+    TELEMETRY_MSG_SIZE_LIMIT_IN_CHARACTERS = 3072  # 3KB
+    TELEMETRY_EVENT_SIZE_LIMIT_IN_CHARACTERS = 6144  # 6KB
+    TELEMETRY_EVENT_FILE_SIZE_LIMIT_IN_CHARACTERS = 4194304  # 4MB
+    TELEMETRY_DIR_SIZE_LIMIT_IN_CHARACTERS = 41943040  # 40MB
```
kjohn-msft, line 65: IN_BYTES -- all lines 62 to 65.  Bytes per character vary based on encoding used from 1 to 4 bytes. So characters would not be the correct measurement here.

src/extension/src/TelemetryWriter.py
```diff
@@ -13,75 +13,107 @@
...
+        """ Removes line breaks, tabs and restricts message to a character limit """
+        message_size_limit = Constants.TELEMETRY_MSG_SIZE_LIMIT_IN_CHARACTERS
+        formatted_message = re.sub(r"\s+", " ", str(full_message))
+        return formatted_message[:message_size_limit - 3] + '...' if len(formatted_message) > message_size_limit else formatted_message
+
+    def write_event(self, task_name, message, event_level=Constants.TelemetryEventLevel.Informational):
+        # create event json
+        # check if event size complies with restriction i.e. 6k
+        # write event to events file
+        if os.path.getsize(self.events_folder_path) >= Constants.TELEMETRY_DIR_SIZE_LIMIT_IN_CHARACTERS:
```
kjohn-msft, line 57: Please sort out the characters vs. bytes issue, in actual outcomes, then in implementation and backwards to the variable names used.

src/extension/src/TelemetryWriter.py
```diff
@@ -13,75 +13,107 @@
...
 
-    # region Transport layer
-    def try_send_message(self, message, category=Constants.TelemetryInfo):
-        raise NotImplementedError
+    @staticmethod
+    def __ensure_message_restriction_compliance(full_message):
+        """ Removes line breaks, tabs and restricts message to a character limit """
+        message_size_limit = Constants.TELEMETRY_MSG_SIZE_LIMIT_IN_CHARACTERS
+        formatted_message = re.sub(r"\s+", " ", str(full_message))
+        return formatted_message[:message_size_limit - 3] + '...' if len(formatted_message) > message_size_limit else formatted_message
```
kjohn-msft, line 51: Silently dropping characters leaves room for confusion. Leave a count of how many characters were dropped in this process at the end of the message (the message will need to be further shortened to accommodate a brief chunk like "[2345 chars dropped]"). Also ensure the full message is logged in the log file with a note saying it was truncated in telemetry.

src/extension/src/TelemetryWriter.py
```diff
@@ -13,75 +13,107 @@
...
+        event = self.__new_event_json(task_name, event_level, message)
+        if len(json.dumps(event)) > Constants.TELEMETRY_EVENT_SIZE_LIMIT_IN_CHARACTERS:
+            print("Cannot send data to telemetry as it exceeded the acceptable data size. [Data not sent={0}]".format(json.dumps(message)))
+        else:
+            self.write_using_temp_file(self.events_folder_path, event)
+
+    def __delete_older_events(self):
+        """ Delete older events until the atleast one new event file can be added as per the size restrictions """
+        try:
+            if os.path.getsize(self.events_folder_path) < Constants.TELEMETRY_DIR_SIZE_LIMIT_IN_CHARACTERS - Constants.TELEMETRY_EVENT_FILE_SIZE_LIMIT_IN_CHARACTERS:
```
kjohn-msft, line 69: Same bytes != characters comment. Needs to be rationalized everywhere.

src/extension/tests/helpers/RuntimeComposer.py
```diff
@@ -1,13 +1,16 @@
...
 from extension.src.Utility import Utility
 from extension.src.file_handlers.JsonFileHandler import JsonFileHandler
 from extension.src.local_loggers.Logger import Logger
 
 
 class RuntimeComposer(object):
     def __init__(self):
-        self.logger = Logger()
+        telemetry_writer = TelemetryWriter()
+        self.logger = Logger(telemetry_writer=telemetry_writer)
```
kjohn-msft, line 13: Note while reviewing; may remove as review completes -- ensure logger can be initialized without telemetry_writer. 
Regardless, telemetry_writer being on/off should be a configuration setting inside Constants.

src/extension/tests/TestTelemetryWriter.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def test_write_event_msg_size_limit(self):
+        message = "a"*3074
+        self.telemetry_writer.write_event("Test Task", message, Constants.TelemetryEventLevel.Error)
+        with open(os.path.join(self.telemetry_writer.events_folder_path, os.listdir(self.telemetry_writer.events_folder_path)[0]), 'r+') as f:
+            events = json.load(f)
+            self.assertTrue(events is not None)
+            self.assertEquals(events[0]["TaskName"], "Test Task")
+            self.assertTrue(len(events[0]["Message"]) == 3072)
+            self.assertEquals(events[0]["Message"], "a"*3069 + "...")
```
kjohn-msft, line 48: Change this (with the product code) to cover how many chars were truncated.

src/extension/tests/helpers/RuntimeComposer.py
kjohn-msft: Confirmed: logger can be initialized without t_w

src/extension/src/TelemetryWriter.py
```diff
@@ -13,75 +13,107 @@
...
-            'cmd': str(cmd),
-            'code': str(code),
-            'output': str(output)[0:3072]
+        self.events_folder_path = None
+        self.operation_id = ""
+
+    def __new_event_json(self, task_name, event_level, message):
+        return {
+            "Version": Constants.EXT_VERSION,
+            "Timestamp": str((datetime.datetime.utcnow()).strftime("%Y-%m-%dT%H:%M:%SZ")),
```
kjohn-msft, line 37: "%Y-%m-%dT%H:%M:%SZ" is a good candidate to move to constants. Please replicate in core in some subsequent review.

src/extension/src/TelemetryWriter.py
```diff
@@ -13,75 +13,107 @@
...
+            "TaskName": task_name,
+            "EventLevel": event_level,
+            "Message": self.__ensure_message_restriction_compliance(message),
+            "EventPid": "",
+            "EventTid": "",
+            "OperationId": self.operation_id  # we can provide activity id from config settings here, but currently we only read settings file for enable command
```
kjohn-msft, line 43: It's good to have comments like this. Helps with long term maintainability.

So long as we know what non-enable command is running and the time actions related to it occurred, that's a good start.

src/extension/src/TelemetryWriter.py
```diff
@@ -13,75 +13,107 @@
...
+                        os.remove(event_file)
+                        print("Deleted [File={0}]".format(repr(event_file)))
+                except Exception as e:
+                    print("Error deleting event file. [File={0}] [Exception={1}]".format(repr(event_file), repr(e)))
+
+        except Exception as err:
+            print("Error deleting older event files. [EventsFolderPath={0}] [Exception={1}]".format(repr(self.events_folder_path), repr(err)))
+
+    @staticmethod
+    def write_using_temp_file(folder_path, data, mode='w'):
```
kjohn-msft, line 93: Review a better method name to reflect what it does -- this is too generic relative to the specific behavior the code has.

src/extension/src/TelemetryWriter.py
```diff
@@ -13,75 +13,107 @@
...
+                try:
+                    if os.path.getsize(self.events_folder_path) < Constants.TELEMETRY_DIR_SIZE_LIMIT_IN_CHARACTERS - Constants.TELEMETRY_EVENT_FILE_SIZE_LIMIT_IN_CHARACTERS:
+                        print("Not deleting any more event files as the event directory has sufficient space to add at least one new event file")
+                        break
+
+                    if os.path.exists(event_file):
+                        os.remove(event_file)
+                        print("Deleted [File={0}]".format(repr(event_file)))
+                except Exception as e:
+                    print("Error deleting event file. [File={0}] [Exception={1}]".format(repr(event_file), repr(e)))
```
kjohn-msft, line 87: Since this is swallowing exceptions, what's the recovery/continuation strategy if there are rampant failures (like all event files can't be deleted)? At what point is it a terminal exception?

src/extension/src/TelemetryWriter.py
```diff
@@ -13,75 +13,107 @@
...
+
+        except Exception as err:
+            print("Error deleting older event files. [EventsFolderPath={0}] [Exception={1}]".format(repr(self.events_folder_path), repr(err)))
```
kjohn-msft, line 90: I don't understand this -- who is throwing the exception that is being caught here?

kjohn-msft (commented):
Comments inline

src/extension/src/TelemetryWriter.py
rane-rajasi: Changed to throw a terminal exception is no files were deleted or the directory size exceeds limit even after deletion attempts

src/extension/src/TelemetryWriter.py
rane-rajasi: removed

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -419,7 +419,7 @@ def __write_record(self, operation, code, output, delay, timestamp=None):
 
         try:
             record = {
-                "timestamp": str(timestamp) if timestamp is not None else datetime.datetime.strptime(str(datetime.datetime.utcnow()).split(".")[0], "%Y-%m-%dT%H:%M:%SZ"), #WRONG
+                "timestamp": str(timestamp) if timestamp is not None else datetime.datetime.strptime(str(datetime.datetime.utcnow()).split(".")[0], Constants.DATE_FORMAT), #WRONG
```
kjohn-msft, line 422: nit: UTC_DATETIME_FORMAT

src/core/src/core_logic/ExecutionConfig.py
```diff
@@ -41,6 +41,7 @@ def __init__(self, env_layer, composite_logger, execution_parameters):
         self.log_folder = self.environment_settings[Constants.EnvSettings.LOG_FOLDER]
         self.config_folder = self.environment_settings[Constants.EnvSettings.CONFIG_FOLDER]
         self.status_folder = self.environment_settings[Constants.EnvSettings.STATUS_FOLDER]
+        self.events_folder = self.environment_settings[Constants.EnvSettings.EVENTS_FOLDER]
```
kjohn-msft, line 44: Events folder is not guaranteed to be there in environment settings. -- note to self for looping back in review.

Please include tasks for a test plan for testing on an agent that doesn't support it and testing on an agent that supports it.

src/core/src/bootstrap/Constants.py
```diff
@@ -163,13 +167,23 @@ class PatchOperationErrorCodes(EnumBackport):
 
     ERROR_ADDED_TO_STATUS = "Error_added_to_status"
 
-    # Telemetry Categories
-    TELEMETRY_OPERATION_STATE = "State"
-    TELEMETRY_CONFIG = "Config"
-    TELEMETRY_PACKAGE = "PackageInfo"
-    TELEMETRY_ERROR = "Error"
-    TELEMETRY_INFO = "Info"
-    TELEMETRY_DEBUG = "Debug"
```
kjohn-msft, line 172: Please retain all of this - the fundamental expectation was enabling telemetry was a pipeline to carry information already destined for telemetry. Issues such as this should not be coming up late in development. Expectations are usually clarified upfront in design reviews. 

To avoid this in the future, please ensure a full engineering lifecycle is followed:
- design review with a documented design that's signed off
- translation of spec into code, with any revisions found to be required triggering a re-review of the specific changes required.

I didn't think to press this for telemetry as I'd incorrectly assumed the expectations were clear. But let's ensure this happens for everything going forward.

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -419,7 +419,7 @@ def __write_record(self, operation, code, output, delay, timestamp=None):
 
         try:
             record = {
-                "timestamp": str(timestamp) if timestamp is not None else datetime.datetime.strptime(str(datetime.datetime.utcnow()).split(".")[0], "%Y-%m-%dT%H:%M:%SZ"), #WRONG
+                "timestamp": str(timestamp) if timestamp is not None else datetime.datetime.strptime(str(datetime.datetime.utcnow()).split(".")[0], Constants.UTC_DATETIME_FORMAT), #WRONG
```
kjohn-msft, line 422: In a different PR (specifically for this), please apply this constant wherever applicable elsewhere.

src/core/src/local_loggers/CompositeLogger.py
```diff
@@ -22,20 +22,26 @@
...
+        self.telemetry_msg_truncated_note = " This message will be truncated in telemetry as it exceeds the size limit."
 
-    def log(self, message):
+    def log(self, message, message_type=Constants.TelemetryEventLevel.Informational):
         """log output"""
         message = self.__remove_substring_from_message(message, Constants.ERROR_ADDED_TO_STATUS)
         message = message.strip()
+        if self.telemetry_writer is not None and self.telemetry_writer.events_folder_path is not None:
+            self.telemetry_writer.write_event(self.telemetry_task, message, message_type)
+            message = self.__add_telemetry_error_note(message)
```
kjohn-msft, line 44: Telemetry is not best effort.. The design tenet here is:
- Does the agent support telemetry? No = don't allow execution to proceed. Write the error to CRP through the status file asking the customer to update their agent.
- If the agent supports telemetry, but telemetry is somehow not working (in a way we can detect), we need to stop execution again with the same process.

The only exception to this would be to allow developer validation at our end.

src/core/src/local_loggers/CompositeLogger.py
```diff
@@ -47,26 +53,34 @@ def log_error(self, message):
...
+        self.log(message, message_type=Constants.TelemetryEventLevel.Error)
 
     def log_warning(self, message):
         """log warning"""
         message = self.__remove_substring_from_message(message, Constants.ERROR_ADDED_TO_STATUS)
         message = self.WARNING + (self.NEWLINE_REPLACE_CHAR.join(message.split(os.linesep))).strip()
-        self.log(message)
+        self.log(message, message_type=Constants.TelemetryEventLevel.Warning)
 
     def log_debug(self, message):
```
kjohn-msft, line 64: Include a one-off log level just for the telemetry writer, invoked only from the telemetry writer where it may report any rare issue with itself. You'd have to work around circular dependencies but it's doable.

src/core/src/local_loggers/CompositeLogger.py
```diff
@@ -77,3 +91,9 @@ def __remove_substring_from_message(message, substring=Constants.ERROR_ADDED_TO_
             message = message.replace("[{0}]".format(Constants.ERROR_ADDED_TO_STATUS), "")
         return message
 
+    def __add_telemetry_error_note(self, message):
+        """ Adding telemetry error messages to the original message, so as to log telemetry errors in other logs """
+        if len(message.encode('utf-8')) > Constants.TELEMETRY_MSG_SIZE_LIMIT_IN_BYTES:
+            message = message + self.telemetry_msg_truncated_note
+        return message
```
kjohn-msft, line 98: Composite logger is holding knowledge/responsibility of telemetry it shouldn't have and creating a stronger coupling between the two classes. Telemetry should report such an issue back to composite logger using the aforementioned one-off log level for telemetry.

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -26,8 +26,8 @@ class AptitudePackageManager(PackageManager):
     """Implementation of Debian/Ubuntu based package management operations"""
 
     # For more details, try `man apt-get` on any Debian/Ubuntu based box.
-    def __init__(self, env_layer, execution_config, composite_logger, telemetry_writer, status_handler):
-        super(AptitudePackageManager, self).__init__(env_layer, execution_config, composite_logger, telemetry_writer, status_handler)
+    def __init__(self, env_layer, execution_config, composite_logger, status_handler):
+        super(AptitudePackageManager, self).__init__(env_layer, execution_config, composite_logger, status_handler)
```
kjohn-msft, line 30: This functionality should be retained here and in other places.

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -75,15 +75,13 @@ def invoke_package_manager(self, command):
                                             'Patch Management cannot proceed successfully. Before the next Patch Operation, please run the following '
                                             'command and perform any configuration steps necessary on the machine to return it to a healthy state: '
                                             'sudo dpkg --configure -a')
-            self.telemetry_writer.send_execution_error(command, code, out)
```
kjohn-msft, line 78: Retain these functions and make them capable of utilizing the telemetry channel.

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -262,33 +259,28 @@ def is_package_version_installed(self, package_name, package_version):
                 if 'Package: ' in line:
                     if package_name in line:
                         composite_found_flag = composite_found_flag | 1
-                    else:  # should never hit for the way this is invoked, hence telemetry
```
kjohn-msft, line 265: Same comment all across -- telemetry exists independently. Composite logger chooses to utilize it.

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -14,131 +14,123 @@
...
+
+                if os.path.exists(event_file):
+                    os.remove(event_file)
+                    print("Deleted event file. [File={0}]".format(repr(event_file)))
+            except Exception as e:
+                print("Error deleting event file. [File={0}] [Exception={1}]".format(repr(event_file), repr(e)))
+
+        if self.__get_events_dir_size() >= Constants.TELEMETRY_DIR_SIZE_LIMIT_IN_BYTES:
+            raise Exception("Older event files were not deleted. Current event will not be sent to telemetry as events directory size exceeds maximum limit")
+
```
kjohn-msft, line 94: Please introduce a telemetry writer only log level in composite logger (that doesn't loop back to telemetry) - all of the code here (and below) are running blind aside from exceptions.

Important: refrain from using print in general but especially inside telemetry logger. The file mirror catch is to only capture things we have inadvertently missed or it's early in the bootstrap process.

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -14,131 +14,123 @@
...
-            payload = json.dumps(payload)[0:4095]
-            for transport in self.data_transports:
-                transport.write(payload)
-            return ""  # for consistency
+            if os.path.exists(file_path):
+                # if file_size exceeds max limit, sleep for 1 second, so the event can be written to a new file
+                file_size = self.get_file_size(file_path)
+                if file_size >= Constants.TELEMETRY_EVENT_FILE_SIZE_LIMIT_IN_BYTES:
+                    time.sleep(1)
+                    file_path = self.__get_event_file_path(folder_path)
```
kjohn-msft, line 105: What's the significance of one second? (not more, not less?) It's a code-smell for potential future timing issues, hence the ask.

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -14,131 +14,123 @@
...
+                else:
+                    prev_events = self.__fetch_events_from_previous_file(file_path)
+
+            prev_events.append(data)
+            with tempfile.NamedTemporaryFile(mode, dir=os.path.dirname(file_path), delete=False) as tf:
+                json.dump(prev_events, tf, default=data.__str__())
+                tempname = tf.name
+            shutil.move(tempname, file_path)
```
kjohn-msft, line 113: This protects against read/write concurrency issues. But how does this protect against concurrency between processes we own that can get spun up at any time?

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -14,131 +14,123 @@
...
-    def get_disk_type(self):
-        """ Retrieve disk info """
-        cmd = "cat /sys/block/sda/queue/rotational"
-        code, out = self.env_layer.run_command_output(cmd, False, False)
-        if "1" in out:
-            return "Hard drive"
-        elif "0" in out:
-            return "SSD"
-        else:
-            return "Unknown"
```
kjohn-msft, line 143: Please retain this and ensure it gets pushed out into the new channel.

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -14,131 +14,123 @@
...
-    # end region
+            raise Exception("Unable to write to {0}. Error: {1}.".format(str(file_path), repr(error)))
+
+    def set_operation_id(self, operation_id):
+        self.operation_id = operation_id
```
kjohn-msft, line 118: Why is this delayed set in Core? We shouldn't be writing out difficult to attribute events..

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -14,131 +14,123 @@
...
+
+    @staticmethod
+    def get_file_size(file_path):
+        return os.path.getsize(file_path)
+
+    @staticmethod
+    def __fetch_events_from_previous_file(file_path):
+        with open(file_path, 'r') as file_handle:
+            file_contents = file_handle.read()
+            return json.loads(file_contents)
```
kjohn-msft, line 135: Be very careful about concurrency issues -- please actually have a design session (put a doc together please) covering concerns you've predicted, anything raised in this review and how they are addressed.

src/extension/src/Constants.py
```diff
@@ -27,6 +27,9 @@ def __iter__(self):
                     if item == self.__dict__[item]:
                         yield item
 
+    # Extension version (todo: move to a different file)
+    EXT_VERSION = "1.6.12"
+
```
kjohn-msft, line 32: You can set this to 3.0.[%exec_sub_ver%] for anything that gets built using packager. There needs to be a better story here. Discuss in design requested previously...

src/extension/src/EnableCommandHandler.py
```diff
@@ -49,8 +49,12 @@ def execute_handler_action(self):
             self.ext_output_status_handler.read_file(self.seq_no)
 
             config_settings = self.ext_config_settings_handler.read_file(self.seq_no)
-            operation = config_settings.__getattribute__(self.config_public_settings.operation)
 
+            # set activity_id in telemetry
+            if self.logger.telemetry_writer is not None:
+                self.logger.telemetry_writer.set_operation_id(config_settings.__getattribute__(self.config_public_settings.activity_id))
+
+            operation = config_settings.__getattribute__(self.config_public_settings.operation)
```
kjohn-msft, line 57: Consider setting activity id exactly once in non-enable scenarios to a single timestamp that can be used to correlate better to all events that came out of that entire operation.

src/extension/tests/helpers/RuntimeComposer.py
kjohn-msft: I don't see the constant to turn telemetry_writer on and off? Expectation is it should feed into config factory into the t_w initialization (arming to on or off). With the possibility to be set to off with either that or when the Current env = Constants.Dev (not mandatory). 

Also please update open source contributor instructions with this to explain how to turn it off.

src/extension/tests/TestTelemetryWriter.py
```diff
@@ -0,0 +1,134 @@
+import json
+import os
```
kjohn-msft, line 2: Pending review once product code changes are set.

src/extension/tests/TestExtEnvHandler.py
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 #
```
kjohn-msft, line 14: All-up -- ensure tests mock running on both older agents without telemetry and those with.

kjohn-msft (commented):
Please review comments -- select comments with background I'm e-mailing out as checklist.

src/extension/src/ProcessHandler.py
```diff
@@ -79,8 +79,8 @@ def start_daemon(self, seq_no, config_settings, ext_env_handler):
         process = subprocess.Popen(command, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         if process.pid is not None:
             self.logger.log("New shell process launched successfully. [Process ID (PID)={0}]".format(str(process.pid)))
-            # Wait for 5 seconds
-            time.sleep(5)
+            # Wait for 2 seconds
+            time.sleep(2)
```
kjohn-msft, line 83: How was this number arrived at?

src/extension/tests/helpers/RuntimeComposer.py
rane-rajasi: Handled in PR #56 for core and PR #55 in handler

rane-rajasi: follow up on implementation on #56

src/core/src/core_logic/ExecutionConfig.py
rane-rajasi: Testing with 2 agent version, one with telemetry support and one without

kjohn-msft:
Please close this PR since the equivalent code has gone in newer PRs.

PR #49, opened 2021-02-02
Fix for rebootPending not set correctly during installation in RHEL and CentOS

rane-rajasi:
Logs for a test run on a RHEL 7.9 machine with the bug fix, added in the task: 9114160

Bug: rebootPending was being set to null.

Issue:
InstallPatches operation checks if reboot is required on the VM, which is then set in rebootPending flag. 
YumPackageManager, after checking if reboot is required was not returning any value, if reboot was not required, making the function return 'None', which is why rebootPending was set to null in the status.

Fix: Ensuring the function do_processes_require_restart() always returns a value.

### DIFF
--- a/src/core/src/package_managers/YumPackageManager.py
+++ b/src/core/src/package_managers/YumPackageManager.py
@@ -400,5 +400,5 @@ class YumPackageManager(PackageManager):
                 process_count += 1
                 process_list_verbose += process_details[1] + " (" + process_details[0] + "), "  # process name and id
 
-            self.composite_logger.log(" - Processes requiring restart (" + str(process_count) + "): [" + process_list_verbose + "<eol>]")
-            return process_count != 0  # True if there were anys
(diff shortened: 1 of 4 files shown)


src/core/src/package_managers/YumPackageManager.py
```diff
@@ -400,5 +400,5 @@ def do_processes_require_restart(self):
                 process_count += 1
                 process_list_verbose += process_details[1] + " (" + process_details[0] + "), "  # process name and id
 
-            self.composite_logger.log(" - Processes requiring restart (" + str(process_count) + "): [" + process_list_verbose + "<eol>]")
-            return process_count != 0  # True if there were anys
+        self.composite_logger.log(" - Processes requiring restart (" + str(process_count) + "): [" + process_list_verbose + "<eol>]")
+        return process_count != 0  # True if there were anys
```
mirichmo, line 404: "anys" seems like a typo

mirichmo (changes requested):

kjohn-msft (approved):

mirichmo (approved):

PR #50, opened 2021-02-10
Adding telemetry to handler

rane-rajasi:

2. Event is a json structure

Tested locally by manually enabling agent's telemetry pipeline on a VM


kjohn-msft:
This file is from Core:
src/core/tests/TestTelemetryWriter.py

rane-rajasi:
> This file is from Core:
> src/core/tests/TestTelemetryWriter.py

Yeah, in the original commit, I had updated TelemetryWriter and TestTelemetryWriter in core as well (even though telemetry writer wasn't used anywhere). Which is why in cherry-pick these files were added here. But in this version, I'm not making any changes to core, only adding telemetry to handler

src/extension/src/ActionHandler.py
```diff
@@ -58,27 +58,40 @@ def determine_operation(self, command):
...
 
+    def setup_telemetry(self):
+        # check if events folder exists, if it does init telemetry, if events folder does not exist, log that telemetry is not supported by agent since events folder does not exist
+        events_folder = self.ext_env_handler.events_folder
+        if events_folder is None or not os.path.exists(events_folder):
+            self.logger.log_warning("Telemetry is not supported by Linux Guest Agent as no events folder location specified in Handler Environment")
+        else:
+            self.logger.log("Telemetry is supported by Linux Guest Agent. Extension will send messages to telemetry")
+            self.logger.telemetry_writer.events_folder_path = events_folder
+
```
kjohn-msft, line 91: Something important from the full review:
Telemetry is not best effort.. The design tenet here is:

- Does the agent support telemetry? No = don't allow execution to proceed. Write the error to CRP through the status file asking the customer to update their agent.
- If the agent supports telemetry, but telemetry is somehow not working (in a way we can detect), we need to stop execution again with the same process.

Important: tests on the final code too must happen against an older agent version for graceful execution stop with clear message going up to the API through status. Please plan for this right now as new VMs created are expected to all get the new agent version later.

kjohn-msft: Important for public preview:
events_folder = self.ext_env_handler.events_folder is not sufficient..

It must check events_folder, then events_folder_preview and then finally fail.

src/extension/src/Constants.py
```diff
@@ -27,6 +27,9 @@ def __iter__(self):
                     if item == self.__dict__[item]:
                         yield item
 
+    # Extension version (todo: move to a different file)
+    EXT_VERSION = "1.6.12"
+
```
kjohn-msft, line 32: Needs to be addressed in full PR.

src/extension/src/ProcessHandler.py
```diff
@@ -78,8 +78,8 @@ def start_daemon(self, seq_no, config_settings, ext_env_handler):
         process = subprocess.Popen(command, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         if process.pid is not None:
             self.logger.log("New shell process launched successfully. [Process ID (PID)={0}]".format(str(process.pid)))
-            # Wait for 5 seconds
-            time.sleep(5)
+            # Wait for 2 seconds
+            time.sleep(2)
```
kjohn-msft, line 82: Same question from original review -- how was this number arrived at? Why is it not 0, 1, 5, 10? Arbitrary numbers can cause timing issues in production on slower machines than we're used to or due to OS level resource contention.

src/extension/src/EnableCommandHandler.py
```diff
@@ -49,6 +49,11 @@ def execute_handler_action(self):
             self.ext_output_status_handler.read_file(self.seq_no)
 
             config_settings = self.ext_config_settings_handler.read_file(self.seq_no)
+
+            # set activity_id in telemetry
+            if self.logger.telemetry_writer is not None:
+                self.logger.telemetry_writer.set_operation_id(config_settings.__getattribute__(self.config_public_settings.activity_id))
+
```
kjohn-msft, line 56: This is not best-effort...

src/extension/src/__main__.py
```diff
@@ -28,14 +29,14 @@
...
-from extension.src.local_loggers.StdOutFileMirror import StdOutFileMirror
 from extension.src.Constants import Constants
 
 
 def main(argv):
     stdout_file_mirror = None
     file_logger = None
-    logger = Logger()
+    telemetry_writer = TelemetryWriter()
+    logger = Logger(telemetry_writer=telemetry_writer)
```
kjohn-msft, line 39: This is important, and relates to the ask to roll the PR back after Canary validation of basic functionality:

Telemetry stands independently of composite logger. 
-	Telemetry may be invoked independent of the logger without any of that showing up in the on-machine log.
-	The logger may write to telemetry but that’s because it’s using that functionality, not because this is the gatekeeper to accessing the telemetry functionality.
-	Additionally the main change was supposed to be changing the telemetry writer transparently to flow out telemetry that’s already set up. (Okay with composite logger utilizing telemetry as it used to in the past in the previous logging mechanism)

kjohn-msft (commented):
Please review and address the blockers and note anything else for the full review (most are also in the full review).

src/extension/src/ActionHandler.py
rane-rajasi: Few concerns here:

- Does the agent support telemetry? No = don't allow execution to proceed. Write the error to CRP through the status file asking the customer to update their agent
1. setup_telemetry is defined within ActionHandler, and used within setup for each handler action (i.e. Install, Enable, Update, etc). 
2. status file is based on seq no. We read/fetch/get seq no only within Enable command. 
3. So at this point in code, handler does not have the seq no and thereby no <seq no>.status file.
4. So the question for us now is, do we check for telemetry on all handler commands or only for enable? Because enable is the only action which writes to status file right now. 
5. If we want to write to status file then telemetry support needs to be checked from within enable handler

- If the agent supports telemetry, but telemetry is somehow not working (in a way we can detect), we need to stop execution again with the same process.
1. The only way for us to know if agent supports telemetry, is to check if HandlerEnvironment.json file contains key 'eventsFolder' or 'eventsFolder_preview'. This check is done in ExtEnvHandler.py when it reads contents from HandlerEnvironment.json.
2. Agent will independently read events we write every 5 mins and there is not reporting mechanism on the VM for the status of these reads. Agent reports all issues in GuestAgentGenericLogs table. 
3. The one check I can think of here, in case an events folder is provided in HandlerEnvironment.json, validating the dir path and reporting a failure in case if the validation fails.


Like I mentioned earlier, the eventsFolder and eventsFolder_preview check is done in ExtEnvHandler. This reads contents of HandlerEnvironment.json and saves status folder, config folder, log folder and event folder paths. In case of events folder, it looks for 'eventsFolder' key within the file first, if not found, looks for 'eventsFolder_preview' key. If neither of these are found, saves None in the ext_env_handler.events_folder
This is then checked in telemetry setup in ActionHandler (here), validating the dir path and reporting whether telemetry is supported

src/extension/src/Constants.py
rane-rajasi: To be sure, thi is teh part about moving this Constant to a different file right?

src/extension/src/ProcessHandler.py
rane-rajasi: This was based on my local check for now, We had 5 second wait earlier. What I observed with that was, since the core process for assessment generally completes within that time (5 seconds), the handler check on whether the core process is running fails and prints out some of the core logs in the stdout log statement here. The core log file does not have the initial log statements that were recorded within handler logs

src/extension/src/EnableCommandHandler.py
rane-rajasi: What should we do here?

src/extension/src/ProcessHandler.py
kjohn-msft: Please change this to check immediately and in a loop for a fixed large timeout period if process hasn't started yet. The loop can exit when timeout (10 seconds?) expires or the process is created. Intervals between checks should be no larger than the shortest possible legitimate execution time for Core.

src/extension/src/Constants.py
kjohn-msft: Make it get populated as part of build/compilation from a different file that may write it across any source that needs it. Just as a note for full PR, may be resolved here. Non-issue as this whole PR needs to be reverted after validation.

src/extension/src/ActionHandler.py
kjohn-msft: Meta-comment: These are things that need to go into a design doc and reviewed prior to implementation. Please ensure all of these are covered in a design doc and reviewed prior to further reviews on the full PR.

Specifically here: For *canary testing only*, we can just stop execution when the agent is known not to support telemetry and write it to the log file. This necessitates testing against a VM with an older agent version. The pre-requisite is the conversation that needs to happen with the agent team on how to best prep test environment faithfully for that.

src/extension/src/__main__.py
kjohn-msft: As shared offline - ^ this is a must for the full PR. Okay for one-off validation and then roll back of the code.

src/extension/src/EnableCommandHandler.py
kjohn-msft: Doesn't need to be addressed in this PR if we're stopping execution as in the other comment early and just writing to the log file.

kjohn-msft (commented):
Additional comments inline.

src/extension/src/ActionHandler.py
rane-rajasi: Updated the code to stop extension flow, logging an error in handler log

src/extension/src/ActionHandler.py
```diff
@@ -58,27 +58,42 @@ def determine_operation(self, command):
...
+    def setup_telemetry(self):
+        # check if events folder exists, if it does init telemetry, if events folder does not exist, log that telemetry is not supported by agent since events folder does not exist
+        events_folder = self.ext_env_handler.events_folder
+        if events_folder is None or not os.path.exists(events_folder):
+            err_msg = "Telemetry is not supported by Linux Guest Agent as no events folder location specified in Handler Environment. Extension will not continue further. Please update Linux Agent on the VM.\n"
+            self.logger.log_error(err_msg)
+            exit(Constants.ExitCode.HandlerFailed)
+        else:
+            self.logger.log("Telemetry is supported by Linux Guest Agent. Extension will send messages to telemetry")
+            self.logger.telemetry_writer.events_folder_path = events_folder
```
kjohn-msft, line 92: I felt like I left this in one of the PRs -- it needs to check in this order:
- events_folder -- which will fail this time around
- events_folder_preview -- which will work for canary at this point
- then fail if either of the two are not found.

rane-rajasi: This is addressed (mentioned in my earlier comment as well):

the eventsFolder and eventsFolder_preview check is done in ExtEnvHandler. This reads contents of HandlerEnvironment.json and saves status folder, config folder, log folder and event folder paths. In case of events folder, it looks for 'eventsFolder' key within the file first, if not found, looks for 'eventsFolder_preview' key. If neither of these are found, saves None in the ext_env_handler.events_folder
This is then checked in telemetry setup in ActionHandler (here), looking for either none or an invalid path to determine whether  telemetry is supported

kjohn-msft (approved):

PR #52, opened 2021-02-18
Adding telemetry to handler similar to commit cfe345f7 while also addressing issues and feedback on the earlier commit

rane-rajasi:

Write events (i.e. logs) to json file/s
Event is a json structure
One file can have multiple events. There can be multiple event files. It is up to the extension on how to write event and event files. We are creating new event files/ms
Size restrictions as defined in Constants.py. Agent will check for these limitations, but good to check them at our end as well.
Agent will pick up the latest event files/5 mins. (Latest to avoid missing recent data in case the events exceed limit and some have to be deleted)

Includes:

- Adding telemetry to handler as was done in commit cfe345f7with a few changes
- All handler actions (install, enable, uninstall, etc) will not return an error if telemetry is not available. Will log a message stating telemetry not supported by agent in the extension's log files and continue execution as usual.
- In both telemetry available and not available scenarios, enable action will, as mentioned above, log a message and continue execution and invoke core with all config settings. Core is responsible to stop execution if telemetry is not available (This will be ensured when telemetry is added in Core)
- Addressed PR feedback on updating process run check used to verify if core process has launched successfully


src/extension/src/ProcessHandler.py
```diff
@@ -78,15 +78,8 @@ def start_daemon(self, seq_no, config_settings, ext_env_handler):
         process = subprocess.Popen(command, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         if process.pid is not None:
             self.logger.log("New shell process launched successfully. [Process ID (PID)={0}]".format(str(process.pid)))
-            # Wait for 5 seconds
-            time.sleep(5)
```
rane-rajasi, line 82: Addressing this feedback from the previous PR: https://github.com/Azure/LinuxPatchExtension/pull/50#discussion_r574215666 

NOTE: kept the wait time as 5 seconds since that was the original value working so far.

Copying comments from the other PR for easy access
kjohn-msft 8 days ago Member
Same question from original review -- how was this number arrived at? Why is it not 0, 1, 5, 10? Arbitrary numbers can cause timing issues in production on slower machines than we're used to or due to OS level resource contention.

 
@rane-rajasi rane-rajasi 7 days ago Author Member
This was based on my local check for now, We had 5 second wait earlier. What I observed with that was, since the core process for assessment generally completes within that time (5 seconds), the handler check on whether the core process is running fails and prints out some of the core logs in the stdout log statement here. The core log file does not have the initial log statements that were recorded within handler logs

 
@kjohn-msft kjohn-msft 7 days ago Member
Please change this to check immediately and in a loop for a fixed large timeout period if process hasn't started yet. The loop can exit when timeout (10 seconds?) expires or the process is created. Intervals between checks should be no larger than the shortest possible legitimate execution time for Core.

src/extension/src/ActionHandler.py
```diff
@@ -58,27 +58,41 @@ def determine_operation(self, command):
...
             self.stdout_file_mirror.stop()
         if self.file_logger is not None:
             self.file_logger.close()
 
+    def setup_telemetry(self):
+        # check if events folder exists, if it does init telemetry, if events folder does not exist, log that telemetry is not supported by agent since events folder does not exist
+        events_folder = self.ext_env_handler.events_folder
+        if events_folder is None or not os.path.exists(events_folder):
+            err_msg = "Telemetry is not supported by Linux Guest Agent as no events folder location specified in Handler Environment. Extension will not perform the requested operation unless Linux Agent is updated to a higher version that supports telemetry. Please update Linux Agent on the VM.\n"
+            self.logger.log_error(err_msg)
```
kjohn-msft, line 88: "The minimum Azure Linux Agent version prerequisite for the Linux patching was not met . Please update the Azure Linux Agent on this machine." -- we know what this is so in this one case, it doesn't need to be more explicit.

src/extension/src/ActionHandler.py
```diff
@@ -58,27 +58,41 @@ def determine_operation(self, command):
...
+        else:
+            self.logger.log("Telemetry is supported by Linux Guest Agent. Extension will send messages to telemetry")
```
kjohn-msft, line 90: "The minimum Azure Linux Agent version prerequisite for Linux patching was met."

src/extension/src/TelemetryWriter.py
```diff
@@ -13,75 +13,123 @@
...
+    def write_event_using_temp_file(self, folder_path, data, mode='w'):
+        """ Writes to a temp file in a single operation and then moves/overrides the original file with the temp """
+        file_path = self.__get_event_file_path(folder_path)
+        prev_events = []
+        try:
+            if os.path.exists(file_path):
+                # if file_size exceeds max limit, sleep for 1 second, so the event can be written to a new file
+                file_size = self.get_file_size(file_path)
+                if file_size >= Constants.TELEMETRY_EVENT_FILE_SIZE_LIMIT_IN_BYTES:
+                    time.sleep(1)
```
kjohn-msft, line 103: time.sleep is a little bit suspect here.. if there's a good justification, add it as a comment in code. This question will come up for all time.sleeps I see in any PR.

kjohn-msft (changes requested):
Minor comments

src/extension/src/TelemetryWriter.py
rane-rajasi: This is to ensure we don't exceed the agent limit for max # of events in an event file.

We are writing multiple events within a each event file. Since the event file name is based on the current timestamp, in case we have reached the limit for events that can exist in an event file, the handler would wait for sometime, so the next attempt to write the current event will be in a new file. Sleep is to ensure a new file is created since the file name is a timestamp (agent requirement).

Will add a comment in code

rane-rajasi: Comment updated on line 101

kjohn-msft (approved):

PR #53, opened 2021-02-19
Adding telemetry in core

rane-rajasi:

Write events (i.e. logs) to json file/s
Event is a json structure
One file can have multiple events. There can be multiple event files. It is up to the extension on how to write event and event files. We are creating new event files/ms
Size restrictions as defined in Constants.py. Agent will check for these limitations, but good to check them at our end as well.
Agent will pick up the latest event files/5 mins. (Latest to avoid missing recent data in case the events exceed limit and some have to be deleted)

Includes:

- Adding telemetry to core
- Telemetry enabled checked within Patch Assessment and execution stopped in case it is not enabled with the corresponding error message added to status blob (validation performed within Patch Assessment since it is the first action for all requests)

### DIFF
--- a/src/core/src/CoreMain.py
+++ b/src/core/src/CoreMain.py
@@ -26,7 +26,7 @@ class CoreMain(object):
         file_logger = bootstrapper.file_logger
         composite_logger = bootstrapper.composite_logger
         stdout_file_mirror = bootstrapper.stdout_file_mirror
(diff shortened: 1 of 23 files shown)


rane-rajasi:
Not to be used, have a separate one for adding telemetry in core: https://github.com/Azure/LinuxPatchExtension/pull/56

PR #56, opened 2021-02-20
Adding telemetry to core

rane-rajasi:

2. Event is a json structure

- Also, contains a telemetry on/off switch at extension
- Adds a telemetry event counter at the end of every message written to telemetry

Tested locally by manually enabling agent's telemetry pipeline on a VM

ToDo:
- [x] Fix unit tests

### DIFF
--- a/src/core/src/CoreMain.py
+++ b/src/core/src/CoreMain.py
@@ -26,7 +26,8 @@ class CoreMain(object):
         file_logger = bootstrapper.file_logger
         composite_logger = bootstrapper.composite_logger
         stdout_file_mirror = bootstrapper.stdout_file_mirror
-        lifecycle_manager = telemetry_writer = status_handler = None
+        telemetry_writer = bootstrapper.telemetry_writer
+        lifecycle_manager = status_handler = None
 
         # Init operation statuses
         patch_operation_requested = Constants.UNKNOWN
@@ -37,8 +38,7 @@ class CoreMain(object):
             # Level 2 bootstrapping
             composite_logger.log_debug("Building out full container...")
             container = bootstrapper.build_out_container()
-            lifecycle_manager, telemetry_writer, status_handler = bootstrapper.build_core_components(container)
-            composite_logger.telemetry_writer = telemetry_writer  # Need to set telemetry_writer within logger to enable sending all logs to telemetry
+            lifecycle_manager, status_handler = bootstrapper.build_core_components(container)
             composite_logger.log_debug("Completed building out full container.\n\n")
 
             # Basic environment check
@@ -49,7 +49,7 @@ class CoreMain(object):
             # Execution config retrieval
             composite_logger.log_debug("Obtaining execution configuration...")
             execution_config = container.get('execution_config')
-            telemetry_writer.setup_telemetry(execution_config)
+            telemetry_writer.set_operation_id(execution_config.activity_id)
             patch_operation_requested = execution_config.operation.lower()
             patch_assessor = container.get('patch_assessor')
             package_manager = container.get('package_manager')
--- a/src/core/src/CoreMain.py
+++ b/src/core/src/CoreMain.py
@@ -38,7 +38,7 @@ class CoreMain(object):
             composite_logger.log_debug("Building out full container...")
             container = bootstrapper.build_out_container()
             lifecycle_manager, telemetry_writer, status_handler = bootstrapper.build_core_components(container)
-            composite_logger.telemetry_writer = telemetry_writer  # Need to set telemetry_writer within composite logger to enable all logs to be sent to telemetry
+            composite_logger.telemetry_writer = telemetry_writer  # Need to set telemetry_writer within logger to enable sending all logs to telemetry
             composite_logger.log_debug("Completed building out full container.\n\n")
 
             # Basic environment check
@@ -49,6 +49,7 @@ class CoreMain(object):
             # Execution config retrieval
             composite_logger.log_debug("Obtaining execution configuration...")
             execution_config = container.get('execution_config')
+            telemetry_writer.setup_telemetry(execution_config)
             patch_operation_requested = execution_config.operation.lower()
             patch_assessor = container.get('patch_assessor')
             package_manager = container.get('package_manager')
@@ -81,7 +82,6 @@ class CoreMain(object):
 
             composite_logger.log_debug("Safely completing required operations after exception...")
             if telemetry_writer is not None:
-                # telemetry_writer.send_error_info("EXCEPTION: " + repr(error))
                 telemetry_writer.write_event("EXCEPTION: " + repr(error), Constants.TelemetryEventLevel.Error)
             if status_handler is not None:
                 composite_logger.log_debug(' - Status handler pending writes flags [I=' + str(patch_installation_successful) + ', A=' + str(patch_assessment_successful) + ']')
--- a/src/core/src/CoreMain.py
+++ b/src/core/src/CoreMain.py
@@ -38,6 +38,7 @@ class CoreMain(object):
             composite_logger.log_debug("Building out full container...")
             container = bootstrapper.build_out_container()
             lifecycle_manager, telemetry_writer, status_handler = bootstrapper.build_core_components(container)
+            composite_logger.telemetry_writer = telemetry_writer  # Need to set telemetry_writer within composite logger to enable all logs to be sent to telemetry
             composite_logger.log_debug("Completed building out full container.\n\n")
 
             # Basic environment check
@@ -80,7 +81,8 @@ class CoreMain(object):
 
             composite_logger.log_debug("Safely completing required operations after exception...")
             if telemetry_writer is not None:
-                telemetry_writer.send_error_info("EXCEPTION: " + repr(error))
+                # telemetry_writer.send_error_info("EXCEPTION: " + repr(error))
+                telemetry_writer.write_event("EXCEPTION: " + repr(error), Constants.TelemetryEventLevel.Error)
             if status_handler is not None:
                 composite_logger.log_debug(' - Status handler pending writes flags [I=' + str(patch_installation_successful) + ', A=' + str(patch_assessment_successful) + ']')
                 if patch_operation_requested == Constants.INSTALLATION.lower() and not patch_installation_successful:
@@ -103,8 +105,7 @@ class CoreMain(object):
             if lifecycle_manager is not None:
                 lifecycle_manager.update_core_sequence(completed=True)
 
-            telemetry_writer.send_runbook_state_info("Succeeded.")
-            telemetry_writer.close_transports()
+            telemetry_writer.write_event("Completed Linux Patch core operation.", Constants.TelemetryEventLevel.Informational)
 
             stdout_file_mirror.stop()
             file_logger.close(message_at_close="<End of output>")
--- a/src/core/src/bootstrap/Bootstrapper.py
+++ b/src/core/src/bootstrap/Bootstrapper.py
@@ -23,6 +23,7 @@ from core.src.bootstrap.ConfigurationFactory import ConfigurationFactory
 from core.src.bootstrap.Constants import Constants
 from core.src.bootstrap.Container import Container
 from core.src.local_loggers.StdOutFileMirror import StdOutFileMirror
+from core.src.service_interfaces.TelemetryWriter import TelemetryWriter
 
 
 class Bootstrapper(object):
--- a/src/core/src/bootstrap/Bootstrapper.py
+++ b/src/core/src/bootstrap/Bootstrapper.py
@@ -30,12 +30,12 @@ class Bootstrapper(object):
         # Environment awareness
         self.current_env = self.get_current_env()
         self.argv = argv
-        self.log_file_path, self.real_record_path = self.get_log_file_and_real_record_paths(argv)
+        self.log_file_path, self.real_record_path, self.events_folder = self.get_path_to_log_files_and_telemetry_dir(argv)
         self.recorder_enabled, self.emulator_enabled = self.get_recorder_emulator_flags(argv)
 
         # Container initialization
         print("Building bootstrap container configuration...")
-        self.configuration_factory = ConfigurationFactory(self.log_file_path, self.real_record_path, self.recorder_enabled, self.emulator_enabled)
+        self.configuration_factory = ConfigurationFactory(self.log_file_path, self.real_record_path, self.recorder_enabled, self.emulator_enabled, self.events_folder)
         self.container = Container()
         self.container.build(self.configuration_factory.get_bootstrap_configuration(self.current_env))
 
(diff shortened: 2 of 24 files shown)


src/core/src/CoreMain.py
```diff
@@ -38,6 +38,7 @@ def __init__(self, argv):
             composite_logger.log_debug("Building out full container...")
             container = bootstrapper.build_out_container()
             lifecycle_manager, telemetry_writer, status_handler = bootstrapper.build_core_components(container)
+            composite_logger.telemetry_writer = telemetry_writer  # Need to set telemetry_writer within composite logger to enable all logs to be sent to telemetry
             composite_logger.log_debug("Completed building out full container.\n\n")
```
rane-rajasi, line 42: update comment: # Need to set telemetry_writer within logger to enable sending all logs to telemetry

src/core/src/core_logic/PatchAssessor.py
```diff
@@ -77,11 +78,19 @@ def start_assessment(self):
...
             error_msg = "The minimum Azure Linux Agent version prerequisite for Linux patching was not met. Please update the Azure Linux Agent on this machine."
             self.composite_logger.log_error(error_msg)
             raise Exception(error_msg)
 
-        #ToDo: Ensure telemetry is setup correctly at this point
         self.composite_logger.log("The minimum Azure Linux Agent version prerequisite for Linux patching was met.")
+        self.telemetry_writer.events_folder_path = self.execution_config.events_folder
+        self.telemetry_writer.set_operation_id(self.execution_config.activity_id)
+        self.telemetry_writer.telemetry_startup()
+        self.is_telemetry_setup = True
```
kjohn-msft, line 94: To much tight coupling going on here (actually the whole function). The innards of telemetry_writer are way too exposed here.

src/core/src/bootstrap/Bootstrapper.py
```diff
@@ -23,6 +23,7 @@
 from core.src.bootstrap.Constants import Constants
 from core.src.bootstrap.Container import Container
 from core.src.local_loggers.StdOutFileMirror import StdOutFileMirror
+from core.src.service_interfaces.TelemetryWriter import TelemetryWriter
```
kjohn-msft, line 28: **Overarching comment for code as in this iteration:**
I understand the reason telemetry writer is only fully initialized in PatchAssessor is so that the error can be reported.

Conceptually and in the specific implementation however, this creates a tight coupling between those classes, and also relies on a specific assumption that the order of operation execution will remain the same way forever (assess then patch (then assess)) creating rigidity in the code and a subsequent fragility. [Academic comment since this doesn't apply to what is requested below: telemetry_writer should be idempotent vs. assessor avoiding a second call into it]

Consider this (or similar instead):::
Initialize telemetry writer within bootstrapper.
- Telemetry writer should internally store in a variable if it failed to initialize because the agent doesn't support it.
      - The way it's now causes tight coupling (see comment elsewhere).
- This failure to initialize can later be consumed in both the Assessor and Installation classes to report into status.
- The highest catch block in Core Main whenever it sees a *first* (not second!) assessment failure on an installation operation, reports into the installation block in status also (covers all first assessment failures).

How to initialize within bootstrapper
- See line 68 for deserialized environment_settings.
- Pull it out similar to that from environment_settings, pass it back out to __init__ as a third variable telemetry_events_path
- Use that on line 51 to do as-much-as-possible initialization of telemetry_writer (obv. missing activity id etc).

Then, back in CoreMain, immediately after execution_config is initialized (line 50 or so), complete the full initialization of telemetry_writer.

When assessment *or* installation runs (both should have this logic, and while today only assessment will ever run it as the execution flow diverts away in older agent case, it preserves resiliency):
- a loosely coupled check of telemetry_writer.is_agent_compatible() or somethign similar returns false -- then write appropriately to status and raise an exception.

This is a broad swathes painting above - please clarify any details that I may have skimmed over to not add even more verbosity, fill in finer details, solidify the implementation and bring back any concerns that may arise.

kjohn-msft (commented):
Telemetry writer initialization flow comment -- needs adjustment.

src/core/src/bootstrap/ConfigurationFactory.py
```diff
@@ -130,10 +130,17 @@ def new_bootstrap_configuration(config_env, log_file_path, real_record_path, rec
...
                 'component_args': ['env_layer', 'file_logger'],
                 'component_kwargs': {
-                    'current_env': config_env
+                    'current_env': config_env,
+                    'telemetry_writer': None  # Has to be initialized without telemetry_writer to avoid running into a circular dependency loop. Telemetry writer within composite logger will be set later after telemetry writer has been initialized
+                }
+            },
+            'telemetry_writer': {
+                'component': TelemetryWriter,
+                'component_args': ['composite_logger', 'env_layer'],
```
kjohn-msft, line 139: Re-order this to env_layer, composite_logger -- there is a call-order established for consistency across classes in this file. See lines 163, 168, etc. for example.

src/core/src/bootstrap/Constants.py
```diff
@@ -29,6 +29,9 @@ def __iter__(self):
     GLOBAL_EXCLUSION_LIST = ""   # if a package needs to be blocked across all of Azure
     UNKNOWN = "Unknown"
 
+    # Extension version (todo: move to a different file)
+    EXT_VERSION = "1.6.15"
```
kjohn-msft, line 33: Please create backlog task and link task here to close this comment.

src/core/src/core_logic/PatchAssessor.py
```diff
@@ -27,14 +27,15 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
 
         self.composite_logger = composite_logger
         self.telemetry_writer = telemetry_writer
+        self.is_telemetry_setup = False
```
kjohn-msft, line 30: Why is this necessary? All calls to write telemetry are blind to whether telemetry is turned on.

Unclear about line 38 too. Will leave a comment there.

src/core/src/core_logic/PatchAssessor.py
```diff
@@ -27,14 +27,15 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
...
         self.status_handler = status_handler
 
         self.package_manager = package_manager
 
     def start_assessment(self):
         """ Start an update assessment """
         self.status_handler.set_current_operation(Constants.ASSESSMENT)
-        self.telemetry_setup()
+        self.telemetry_writer.is_agent_compatible()
```
kjohn-msft, line 38: This should be a condition check that returns true or false; if false, assessment writes it's own status entry (error) into status handler, and raises an exception to stop the run.

if not self.telemetry_writer.is_agent_compatible():
     #write out the error and raise exception as you had before (lines 81 to 83).

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -142,3 +112,120 @@ def get_disk_type(self):
...
+
+    def is_agent_compatible(self):
+        """ Verifies if telemetry is available. Stops execution if not available. """
+
+        if not self.__is_agent_compatible:
+            error_msg = "The minimum Azure Linux Agent version prerequisite for Linux patching was not met. Please update the Azure Linux Agent on this machine."
+            self.composite_logger.log_telemetry_module_error(error_msg)
+            raise Exception(error_msg)
+
+        self.composite_logger.log_telemetry_module("The minimum Azure Linux Agent version prerequisite for Linux patching was met.")
```
kjohn-msft, line 230: This was only meant to return a boolean, as the function name itself represents. 

Not what's to be done, but if this were the function body, the function would be self-descriptive in its name like: raise_if_agent_incompatible()

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -87,35 +76,16 @@ def send_machine_config_info(self):
...
-        return self.send_config_info(machine_info, 'machine_config')
+        return self.write_config_info(machine_info, 'machine_config')
 
-    def send_execution_error(self, cmd, code, output):
+    def write_execution_error(self, cmd, code, output):
         # Expected to log any errors from a cmd execution, including package manager execution errors
         error_payload = {
             'cmd': str(cmd),
             'code': str(code),
             'output': str(output)[0:3072]
```
kjohn-msft, line 86: Note that this is a string truncation to avoid excessive payload sizes in V1. Don't leave this as an artefact behind - you've got mechanisms for this, so have it centralized.

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -13,72 +13,61 @@
...
+
 from core.src.bootstrap.Constants import Constants
 
 
 class TelemetryWriter(object):
     """Class for writing telemetry data to data transports"""
 
-    def __init__(self, env_layer, execution_config):
-        self.data_transports = []
+    def __init__(self, composite_logger, env_layer, events_folder_path):
```
kjohn-msft, line 30: Fix ordering to match other class layouts -- env_layer, composite_logger, events_folder_path

src/core/src/local_loggers/CompositeLogger.py
```diff
@@ -47,29 +52,53 @@ def log_error(self, message):
...
         """log verbose"""
         message = self.__remove_substring_from_message(message, Constants.ERROR_ADDED_TO_STATUS)
+        if self.telemetry_writer is not None and self.telemetry_writer.events_folder_path is not None:
+            self.telemetry_writer.write_event(message, Constants.TelemetryEventLevel.Verbose)
         if self.file_logger is not None:
             self.file_logger.write("\n\t" + self.VERBOSE + " " + "\n\t".join(message.strip().splitlines()).strip())
 
+    def log_telemetry_module_error(self, message):
+        """Used exclusively by telemetry writer to log any errors raised within it's operation"""
+        message = self.__remove_substring_from_message(message, Constants.ERROR_ADDED_TO_STATUS)
```
kjohn-msft, line 84: No telemetry related error should be reported to the customer facing status. Given that's the case, why are we attempting to remove such a string?

src/core/src/local_loggers/CompositeLogger.py
```diff
@@ -47,29 +52,53 @@ def log_error(self, message):
...
+        message = (self.NEWLINE_REPLACE_CHAR.join(message.split(os.linesep))).strip()
+        if self.file_logger is not None:
+            timestamp = self.env_layer.datetime.timestamp()
+            self.file_logger.write("\n" + timestamp + "> " + self.TELEMETRY_ERROR + message.strip(), fail_silently=False)
+        else:
+            print(self.TELEMETRY_ERROR + " " + message)
+
+    def log_telemetry_module(self, message):
+        """Used exclusively by telemetry writer to log messages from it's operation"""
+        message = self.__remove_substring_from_message(message, Constants.ERROR_ADDED_TO_STATUS)
```
kjohn-msft, line 94: Same comment here..

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -142,3 +112,120 @@ def get_disk_type(self):
...
+        if not self.__is_agent_compatible or not Constants.TELEMETRY_ENABLED_AT_EXTENSION:
+            return
+
+        self.__delete_older_events()
+
+        event = self.__new_event_json(event_level, message, task_name)
+        if len(json.dumps(event)) > Constants.TELEMETRY_EVENT_SIZE_LIMIT_IN_BYTES:
+            self.composite_logger.log_telemetry_module_error("Cannot send data to telemetry as it exceeded the acceptable data size. [Data not sent={0}]".format(json.dumps(message)))
+        else:
+            self.write_event_using_temp_file(self.events_folder_path, event)
```
kjohn-msft, line 153: Generic comment about telemetry_writer -- the only error from this that may show up in customer facing statuses, etc. should be something generic like 'Internal reporting error. Execution could not complete.' which is piped up the call stack with no other details (thus the technical requirement is a try catch that accomplishes that while logging the specific details within the log file.

We should never have a customer report that they saw such a message since it'd 99.99% be a bug in our code.

kjohn-msft: This may require try catches on most/all externally invokable methods.

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -142,3 +112,120 @@ def get_disk_type(self):
...
+            shutil.move(tempname, file_path)
+        except Exception as error:
+            raise Exception("Unable to write to telemetry. [Event File={0}] [Error={1}].".format(str(file_path), repr(error)))
+
+    def __get_events_dir_size(self):
+        return sum([os.path.getsize(os.path.join(self.events_folder_path, f)) for f in os.listdir(self.events_folder_path) if os.path.isfile(os.path.join(self.events_folder_path, f))])
+
+    @staticmethod
+    def __get_event_file_path(folder_path):
+        return os.path.join(folder_path, str(int(round(time.time() * 1000))) + ".json")
```
kjohn-msft, line 207: document example of return in function comment

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -142,3 +112,120 @@ def get_disk_type(self):
...
+
+    @staticmethod
+    def get_file_size(file_path):
+        return os.path.getsize(file_path)
```
kjohn-msft, line 211: method comment explaining this is for unit test mocking

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -142,3 +112,120 @@ def get_disk_type(self):
...
+            prev_events.append(data)
+            with tempfile.NamedTemporaryFile(mode, dir=os.path.dirname(file_path), delete=False) as tf:
+                json.dump(prev_events, tf, default=data.__str__())
+                tempname = tf.name
+            shutil.move(tempname, file_path)
+        except Exception as error:
+            raise Exception("Unable to write to telemetry. [Event File={0}] [Error={1}].".format(str(file_path), repr(error)))
+
+    def __get_events_dir_size(self):
+        return sum([os.path.getsize(os.path.join(self.events_folder_path, f)) for f in os.listdir(self.events_folder_path) if os.path.isfile(os.path.join(self.events_folder_path, f))])
```
kjohn-msft, line 203: please break this line apart into multiple lines of execution -- doesn't lend to long term readability/maintainability as it is. Clarity > consiseness.

kjohn-msft (commented):
Haven't finished reviewing but sharing comments so far to address.

src/core/src/bootstrap/Constants.py
rane-rajasi: https://msazure.visualstudio.com/One/_queries/edit/9388858/?triage=true

src/core/src/local_loggers/CompositeLogger.py
rane-rajasi: Something that got added during copy paste. Removed

kjohn-msft (changes requested):
Tentatively approved, pending adding counter on telemetry writes as discussed.

kjohn-msft (approved):
- Ensure utf-8 size limits are what agent is expecting.
- Test with max limits privately, and follow up after that.
- Maybe adjust telemetry count message size limit down after research.

PR #57, opened 2021-02-22
Adding dynamic throttling

rane-rajasi:
- Restrict core to write max of 60 event files per minute.
- This is on top of rarane-telemetry-core3 branch, PR #56, since the telemetry for core is not yet merged
- Throttling only added in core


src/core/src/bootstrap/Constants.py
```diff
@@ -175,6 +175,8 @@ class PatchOperationErrorCodes(EnumBackport):
     TELEMETRY_EVENT_FILE_SIZE_LIMIT_IN_BYTES = 4194304  # 4MB
     TELEMETRY_DIR_SIZE_LIMIT_IN_BYTES = 41943040  # 40MB
     TELEMETRY_BUFFER_FOR_DROPPED_COUNT_MSG_IN_BYTES = 80  # buffer for the bytes dropped text added at the end of the truncated telemetry message
+    TELEMETRY_MAX_EVENT_FILE_THROTTLE_COUNT = 60
+    TELEMETRY_MAX_TIME_FOR_EVENT_FILE_THROTTLE = 60
```
kjohn-msft, line 179: Time must have units in name. MAX_TIME_IN_SECONDS in this case. Applies generally to any quantity that needs a unit to make sense.

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -138,18 +141,22 @@ def __ensure_message_restriction_compliance(self, full_message):
 
     def write_event(self, message, event_level=Constants.TelemetryEventLevel.Informational, task_name=Constants.TELEMETRY_TASK_NAME):
         """ Creates and writes event to event file after validating none of the telemetry size restrictions are breached """
-        if self.events_folder_path is None or not os.path.exists(self.events_folder_path) or not Constants.TELEMETRY_ENABLED_AT_EXTENSION:
+        if self.events_folder_path is None or not os.path.exists(self.events_folder_path) or not Constants.TELEMETRY_ENABLED_AT_EXTENSION or not self.__is_telemetry_startup:
```
kjohn-msft, line 144: why is all of this being checked with every single write_event call? The first two parameters should be checked once at init and mapped to the agent_supports_telemetry() / equivalent function talked about on the call.

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -174,25 +181,41 @@ def __delete_older_events(self):
...
+            else:
+                all_events = self.__fetch_events_from_previous_file(file_path)  # fetches existing content within the file
+        all_events.append(data)
+        return file_path, all_events
+
+    def ensure_event_file_max_count_is_met(self):
+        """ Ensures the # of event files that can be written per time unit restriction is met. Returns False if the any updates are required after the restriction enforcement. For eg: file_name is a timestamp and should be modified if a wait is added here """
+        if (datetime.datetime.utcnow() - self.start_time_for_event_file_throttle_check).total_seconds() < Constants.TELEMETRY_MAX_TIME_FOR_EVENT_FILE_THROTTLE:
+            # If event file count limit reached before time period, wait out the remaining time
+            if self.event_file_count >= Constants.TELEMETRY_MAX_EVENT_FILE_THROTTLE_COUNT:
```
kjohn-msft, line 203: Log and write to telemetry if throttling happens.

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -174,25 +181,41 @@ def __delete_older_events(self):
...
+            # if file_size exceeds max limit, sleep for 1 second, so the event can be written to a new file since the event file name is a timestamp
+            if file_size >= Constants.TELEMETRY_EVENT_FILE_SIZE_LIMIT_IN_BYTES:
+                time.sleep(1)
+                file_path = self.__get_event_file_path(folder_path)
+            else:
+                all_events = self.__fetch_events_from_previous_file(file_path)  # fetches existing content within the file
+        all_events.append(data)
+        return file_path, all_events
+
+    def ensure_event_file_max_count_is_met(self):
```
kjohn-msft, line 199: throttle_telemetry_writes_if_required

src/core/src/service_interfaces/TelemetryWriter.py
rane-rajasi: This PR is on top of the telemetry core, which is why these changes appear here

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -174,25 +181,41 @@ def __delete_older_events(self):
...
+        """ Ensures the # of event files that can be written per time unit restriction is met. Returns False if the any updates are required after the restriction enforcement. For eg: file_name is a timestamp and should be modified if a wait is added here """
+        if (datetime.datetime.utcnow() - self.start_time_for_event_file_throttle_check).total_seconds() < Constants.TELEMETRY_MAX_TIME_FOR_EVENT_FILE_THROTTLE:
+            # If event file count limit reached before time period, wait out the remaining time
+            if self.event_file_count >= Constants.TELEMETRY_MAX_EVENT_FILE_THROTTLE_COUNT:
+                time.sleep((datetime.datetime.utcnow() - self.start_time_for_event_file_throttle_check).total_seconds())
```
kjohn-msft, line 204: Calculate the time to sleep, verify again it's not more than the upper limit of time to sleep, log it and then sleep. Gaps in execution time must be ascertainable in logs and telemetry.

kjohn-msft (commented):
Wait for core to complete with merge before working on this further, as there are overlaps.

src/core/src/service_interfaces/TelemetryWriter.py
rane-rajasi: Writing to telemetry in case of throttle: 
There is one caveat to this, the event can only be written after throttling, because

1. Throttle is caused because extension reached the max number of event files that can be added within a certain amount of time. Once the limit is reached, extension waits for the remainder of that time before writing anything to telemetry. Writing a throttle message to telemetry will involve an event file, so even if we do write a throttle message, it can only be after the wait time.

kjohn-msft: It may be written immediately if we throttle at one less than the max...

kjohn-msft (approved):

PR #61, opened 2021-03-15
Additional logging for telemetry writer events

rane-rajasi:
Adding additional logs in case of an exception within telemetry writer to help in debugging errors due to telemetry write.

### DIFF
--- a/src/core/src/service_interfaces/TelemetryWriter.py
+++ b/src/core/src/service_interfaces/TelemetryWriter.py
@@ -133,17 +133,22 @@ class TelemetryWriter(object):
         In case a message is truncated due to size restrictions, adds the count of chars dropped at the end.
         Adds a telemetry event counter at the end of every event, irrespective of truncation, which can be used in debugging operation flow. """
 
-        message_size_limit_in_chars = Constants.TELEMETRY_MSG_SIZE_LIMIT_IN_CHARS
-        formatted_message = re.sub(r"\s+", " ", str(full_message))
+        try:
+            message_size_limit_in_chars = Constants.TELEMETRY_MSG_SIZE_LIMIT_IN_CHARS
+            formatted_message = re.sub(r"\s+", " ", str(full_message))
 
-        if len(formatted_message.encode('utf-8')) + Constants.TELEMETRY_EVENT_COUNTER_MSG_SIZE_LIMIT_IN_CHARS > message_size_limit_in_chars:
-            self.composite_logger.log_telemetry_module("Data sent to telemetry will be truncated as it exceeds size limit. [Message={0}]".format(str(formatted_message)))
-            formatted_message = formatted_message.encode('utf-8')
-            chars_dropped = len(formatted_message) - message_size_limit_in_chars + Constants.TELEMETRY_BUFFER_FOR_DROPPED_COUNT_MSG_IN_CHARS + Constants.TELEMETRY_EVENT_COUNTER_MSG_SIZE_LIMIT_IN_CHARS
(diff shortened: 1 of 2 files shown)


src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -164,70 +169,84 @@ def write_event(self, message, event_level=Constants.TelemetryEventLevel.Informa
             else:
                 file_path, all_events = self.__get_file_and_content_to_write(self.events_folder_path, event)
                 self.__write_event_using_temp_file(file_path, all_events)
+
         except Exception:
             raise Exception("Internal reporting error. Execution could not complete.")
```
kjohn-msft, line 174: This is not good - the call stack is lost when an exception is raised like this. Log and clean raise, as you've done elsewhere is the correct pattern (unless in situations where there are specific situations not to - like to avoid non-user friendly messages from bubbling all the way up).

At the very least, log here before raising. Exercise discretion on how to raise in this instance.

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -164,70 +169,84 @@ def write_event(self, message, event_level=Constants.TelemetryEventLevel.Informa
...
+                else:
+                    all_events = self.__fetch_events_from_previous_file(file_path)  # fetches existing content within the file
+            all_events.append(data)
+            return file_path, all_events
+        except Exception as e:
+            self.composite_logger.log_telemetry_module_error("Error occurred while fetching event file to write the event to. [Error={0}]")
+            raise
 
     def __throttle_telemetry_writes_if_required(self, is_event_file_throttling_needed=True):
         """ Ensures the # of event files that can be written per time unit restriction is met. Returns False if the any updates are required after the restriction enforcement. For eg: file_name is a timestamp and should be modified if a wait is added here """
```
kjohn-msft, line 227: For future callers, document in a comment here who exactly is allowed to call with the throttling disabled and why.

kjohn-msft (commented):
Comments inline

src/core/src/service_interfaces/TelemetryWriter.py
rane-rajasi: This was based on an earlier feedback:
image

src/core/src/service_interfaces/TelemetryWriter.py
rane-rajasi: This was added in write_event, which passes this arg down to __throttle_telemetry_writes_if_required

line 152-153:
def write_event(self, message, event_level=Constants.TelemetryEventLevel.Informational, task_name=Constants.TELEMETRY_TASK_NAME, is_event_file_throttling_needed=True):
        """ Creates and writes event to event file after validating none of the telemetry size restrictions are breached
        NOTE: is_event_file_throttling_needed is used to determine if event file throttling is required and as such should always be True.
        The only scenario where this is False is when throttling is taking place and we write to telemetry about it. i.e. only from within __throttle_telemetry_writes_if_required()"""

src/core/src/service_interfaces/TelemetryWriter.py
kjohn-msft: The original guidance is correct - this is is a case where the exception to the pattern applies on the raise. But without logging prior to the raise, detailed data about the issue is implicitly lost -- hence that ask to log to file (the original comment also carries the same ask: "the technical requirement is a try catch that accomplishes that while logging specific details within the log file").

src/core/src/service_interfaces/TelemetryWriter.py
rane-rajasi: Added a similar comment, as write_event(), in __throttle_telemetry_writes_if_required()

kjohn-msft (approved):

PR #64, opened 2021-03-30
Addressing configure patching requests

rane-rajasi:
Includes 

- Processing Configure patching requests in handler and passing down patchMode to Core
- Core reading patchMode from config settings
- Configure Patching will always run irrespective of the operation
- Handling configure patching requests
   Approach:
       - Get the current auto OS updates set on the machine and add it to ConfigurePatchingSummary substatus
       - If patchMode is AutomaticByPlatform and auto OS updates are enabled on the machine, disable those
       - Update ConfigurePatchingSummary substatus after disabling auto OS updates
 - If the uber-operation is not configure patching, continue with the uber-operation
    NOTE: in this case, any errors thrown by configure patching will be logged in configure patch summary substatus and the uber-operation will continue without any blocks. No error reporting for configure patching within other substatuses

Questions:
- Read this later in our common doc for Windows and Linux, about ConfigurePatchingSummary substatus: "/*This substatus will be shown in all operation to show the current configuration setting the machine*/ " ---- Does this mean even Assessment or Installation requests will have ConfigurePatchingSummary substatus?
   A. Yes, ConfigurePatchingSummary substatus will be shown for all operations, including Assessment and Installation
- In this approach, though auto OS updates will only be disabled for patchMode="AutomaticByPlatform", ConfigurePatchingSummary substatus will always be set, read for ImageDefault/DefaultForImage patchMode as well. Does this follow our expected proposal? 
  A. Yes

### DIFF
--- a/src/core/src/CoreMain.py
+++ b/src/core/src/CoreMain.py
@@ -53,22 +53,26 @@ class CoreMain(object):
             patch_operation_requested = execution_config.operation.lower()
             patch_assessor = container.get('patch_assessor')
             package_manager = container.get('package_manager')
+            configure_patching = container.get('configure_patching')
 
-            # if this is an auto patching installation request, log and disable (if enabled by default) the current auto OS update status. NOTE: log status in a separate file in config settings
-            if execution_config.maintenance_run_id is not None and patch_operation_requested == Constants.INSTALLATION.lower():
-                package_manager.disable_auto_os_update()
+            # # if this is an auto patching installation request, log and disable (if enabled by default) the current auto OS update status. NOTE: log status in a separate file in config settings
+            # if execution_config.maintenance_run_id is not None and patch_operation_requested == Constants.INSTALLATION.lower():
+            #     package_manager.disable_auto_os_update()
 
-            # Assessment happens no matter what
-            patch_assessment_successful = patch_assessor.start_assessment()
-
-            # Patching + additional assessment occurs if the operation is 'Installation'
-            if patch_operation_requested == Constants.INSTALLATION.lower():
-                # setting current operation here, to include patch_installer init within installation actions, ensuring any exceptions during patch_installer init are added in installation summary errors object
-                status_handler.set_current_operation(Constants.INSTALLATION)
-                patch_installer = container.get('patch_installer')
-                patch_installation_successful = patch_installer.start_installation()
-                patch_assessment_successful = False
+            if patch_operation_requested == Constants.CONFIGURE_PATCHING.lower():
+                configure_patching.start_configure_patching()
+            else:
+                # Assessment happens no matter what
                 patch_assessment_successful = patch_assessor.start_assessment()
+
+                # Patching + additional assessment occurs if the operation is 'Installation'
+                if patch_operation_requested == Constants.INSTALLATION.lower():
+                    # setting current operation here, to include patch_installer init within installation actions, ensuring any exceptions during patch_installer init are added in installation summary errors object
+                    status_handler.set_current_operation(Constants.INSTALLATION)
+                    patch_installer = container.get('patch_installer')
+                    patch_installation_successful = patch_installer.start_installation()
+                    patch_assessment_successful = False
+                    patch_assessment_successful = patch_assessor.start_assessment()
 
         except Exception as error:
             # Privileged operation handling for non-production use
@@ -85,12 +89,17 @@ class CoreMain(object):
                 telemetry_writer.write_event("EXCEPTION: " + repr(error), Constants.TelemetryEventLevel.Error)
             if status_handler is not None:
                 composite_logger.log_debug(' - Status handler pending writes flags [I=' + str(patch_installation_successful) + ', A=' + str(patch_assessment_successful) + ']')
-                if patch_operation_requested == Constants.INSTALLATION.lower() and not patch_installation_successful:
-                    status_handler.set_installation_substatus_json(status=Constants.STATUS_ERROR)
-                    composite_logger.log_debug('  -- Persisted failed installation substatus.')
-                if not patch_assessment_successful:
-                    status_handler.set_assessment_substatus_json(status=Constants.STATUS_ERROR)
-                    composite_logger.log_debug('  -- Persisted failed assessment substatus.')
+
+                if patch_operation_requested == Constants.CONFIGURE_PATCHING.lower():
+                    status_handler.set_configure_patching_substatus_json(status=Constants.STATUS_ERROR)
+                    composite_logger.log_debug('  -- Persisted failed configure patching substatus.')
+                else:
+                    if patch_operation_requested == Constants.INSTALLATION.lower() and not patch_installation_successful:
+                        status_handler.set_installation_substatus_json(status=Constants.STATUS_ERROR)
+                        composite_logger.log_debug('  -- Persisted failed installation substatus.')
+                    if not patch_assessment_successful:
+                        status_handler.set_assessment_substatus_json(status=Constants.STATUS_ERROR)
+                        composite_logger.log_debug('  -- Persisted failed assessment substatus.')
 
                 if Constants.ERROR_ADDED_TO_STATUS not in repr(error):
                     status_handler.add_error_to_status("Terminal exception {0}".format(repr(error)), Constants.PatchOperationErrorCodes.OPERATION_FAILED)
--- a/src/core/src/CoreMain.py
+++ b/src/core/src/CoreMain.py
@@ -45,6 +45,7 @@ class CoreMain(object):
 
             # Current operation in status handler is set to either assessment or installation when these operations begin. Setting it to assessment since that is the first operation that runs always.
             # This ensures all errors occurring before assessment starts are logged within the error objects of assessment substatus
+            #ToDo: check with Koshy on what this should be
             if status_handler.get_current_operation() is None:
                 status_handler.set_current_operation(Constants.ASSESSMENT)
 
--- a/src/core/src/CoreMain.py
+++ b/src/core/src/CoreMain.py
@@ -45,9 +45,8 @@ class CoreMain(object):
 
             # Current operation in status handler is set to either assessment or installation when these operations begin. Setting it to assessment since that is the first operation that runs always.
             # This ensures all errors occurring before assessment starts are logged within the error objects of assessment substatus
-            #ToDo: check on what this should be
             if status_handler.get_current_operation() is None:
-                status_handler.set_current_operation(Constants.ASSESSMENT)
+                status_handler.set_current_operation(Constants.ASSESSMENT)  # Needs to be revisited
 
             # Environment startup
             bootstrapper.bootstrap_splash_text()
--- a/src/core/src/CoreMain.py
+++ b/src/core/src/CoreMain.py
@@ -134,7 +134,7 @@ class CoreMain(object):
             composite_logger.log_debug('  -- Persisted failed installation substatus.')
         #todo: Q to reviewer/team, if configure patching fails for a patch installation request, do we just report patch installation summary as failed or also assessment summary?
         # Right now, it reports both assessment and installation summary as failed and adds error message in each that the failure is due to configure patching
-        if not patch_assessment_successful:
+        if not patch_assessment_successful and patch_operation_requested != Constants.CONFIGURE_PATCHING.lower():
             status_handler.set_current_operation(Constants.ASSESSMENT)
             if not configure_patching_successful:
                 status_handler.add_error_to_status("Assessment failed due to configure patching failure. Please refer the error details in configure patching substatus")
--- a/src/core/src/CoreMain.py
+++ b/src/core/src/CoreMain.py
@@ -31,6 +31,7 @@ class CoreMain(object):
 
         # Init operation statuses
         patch_operation_requested = Constants.UNKNOWN
+        configure_patching_successful = False
         patch_assessment_successful = False
         patch_installation_successful = False
         overall_patch_installation_operation_successful = False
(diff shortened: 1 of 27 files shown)


src/core/src/core_logic/PatchAssessor.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
         self.package_manager = package_manager
 
     def start_assessment(self):
-        """ Start an update assessment """
+        """ Start an patch assessment """
```
mirichmo, line 35: nit: a

src/core/src/CoreMain.py
```diff
@@ -53,23 +53,27 @@ def __init__(self, argv):
...
-            if patch_operation_requested == Constants.INSTALLATION.lower():
-                # setting current operation here, to include patch_installer init within installation actions, ensuring any exceptions during patch_installer init are added in installation summary errors object
-                status_handler.set_current_operation(Constants.INSTALLATION)
-                patch_installer = container.get('patch_installer')
-                patch_installation_successful = patch_installer.start_installation()
-                patch_assessment_successful = False
+            if patch_operation_requested == Constants.CONFIGURE_PATCHING.lower():
+                configure_patching.start_configure_patching()
+            else:
+                # Assessment happens no matter what
```
mirichmo, line 65: Even during cancelPatchInstallation? Or is this executing in the core process that will be stopped when cancelPatchInstallation arrives?

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -368,6 +368,38 @@ def get_package_size(self, output):
...
+    def get_current_auto_os_patch_state(self):
+        """ Gets the current auto OS update patch state on the machine """
+        self.composite_logger.log("Fetching the current automatic OS patch state on the machine...")
+        self.__get_current_auto_os_updates_setting_on_machine()
+        if self.unattended_upgrade_value != 0 or self.unattended_upgrade_value != 1:
+            return Constants.PATCH_STATE_UNKNOWN
+        elif self.unattended_upgrade_value:
+            return Constants.PATCH_STATE_ENABLED
+        else:
+            return Constants.PATCH_STATE_DISABLED
```
mirichmo, line 380: The logic may be a bit simpler if you re-order the checks:

if self.unattend_upgrade_value == 0
    return Constants.PATCH_STATE_DISABLED
elfi self.unattend_upgrade_value == 1
    return Constants.PATCH_STATE_ENABLED
else
    return Constants.PATCH_STATE_UNKNOWN

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -327,6 +327,10 @@ def get_package_size(self, output):
     # endregion
 
     # region auto OS updates
+    def get_current_auto_os_patch_state(self):
+        """ Gets the current auto OS update patch state on the machine """
+        pass
```
mirichmo, line 332: Is this a "pass" because it hasn't been implemented yet or because there is nothing to do? A comment explaining why would be useful here and in the other package manager files that define pass for the methods of the interface.

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -356,6 +363,36 @@ def __new_patch_metadata_for_healthstore_json(self, patch_version=Constants.PATC
             "shouldReportToHealthStore": report_to_healthstore
         }
 
+    def set_configure_patching_substatus_json(self, status=Constants.STATUS_TRANSITIONING, code=0, automatic_os_patch_state=Constants.PATCH_STATE_UNKNOWN):
```
mirichmo, line 366: To answer your question in the description:
Yes, configurePatchingSummary should be produced during all operations, as stated in the comment, since it is just reporting the current status of the system.

src/extension/src/EnableCommandHandler.py
```diff
@@ -67,12 +67,12 @@ def execute_handler_action(self):
             core_state_content = self.core_state_handler.read_file()
 
             # If ConfigurePatching is requested, do nothing, will be implemented in future
-            if operation == Constants.CONFIGURE_PATCHING:
-                self.logger.log("Received a configure patching request, no action will be taken as it is not supported for now. [Operation Sequence={0}]".format(str(self.seq_no)))
-                exit(Constants.ExitCode.Okay)
+            # if operation == Constants.CONFIGURE_PATCHING:
+            #     self.logger.log("Received a configure patching request, no action will be taken as it is not supported for now. [Operation Sequence={0}]".format(str(self.seq_no)))
+            #     exit(Constants.ExitCode.Okay)
```
mirichmo, line 72: This should be removed prior to merge.

mirichmo (changes requested):

src/core/src/CoreMain.py
```diff
@@ -53,23 +53,27 @@ def __init__(self, argv):
...
             patch_assessor = container.get('patch_assessor')
             package_manager = container.get('package_manager')
+            configure_patching = container.get('configure_patching')
 
-            # if this is an auto patching installation request, log and disable (if enabled by default) the current auto OS update status. NOTE: log status in a separate file in config settings
-            if execution_config.maintenance_run_id is not None and patch_operation_requested == Constants.INSTALLATION.lower():
-                package_manager.disable_auto_os_update()
+            # # if this is an auto patching installation request, log and disable (if enabled by default) the current auto OS update status. NOTE: log status in a separate file in config settings
+            # if execution_config.maintenance_run_id is not None and patch_operation_requested == Constants.INSTALLATION.lower():
+            #     package_manager.disable_auto_os_update()
```
kjohn-msft, line 60: Configure patching calls may not always get to the machine prior to install patches coming through. So this code needs to still remain, in the sense that if it's not disabled and an autopatching request is what's coming through, disable it.

src/core/src/bootstrap/Constants.py
```diff
@@ -62,20 +62,32 @@ class ConfigSettings(EnumBackport):
...
     PATCH_ASSESSMENT_SUMMARY = "PatchAssessmentSummary"
     PATCH_INSTALLATION_SUMMARY = "PatchInstallationSummary"
     PATCH_METADATA_FOR_HEALTHSTORE = "PatchMetadataForHealthStore"
+    CONFIGURE_PATCHING_SUMMARY = "ConfigurePatchingSummary"
 
     # patch versions for healthstore when there is no maintenance run id
     PATCH_VERSION_UNKNOWN = "UNKNOWN"
 
+    # Patch Modes for Configure Patching
+    DEFAULT_FOR_IMAGE = "DefaultForImage"  # todo: cross check with CRP
```
kjohn-msft, line 83: It's ImageDefault --- is DefaultForImage mentioned anywhere?? Please share if you've seen it like this somewhere..

src/core/src/bootstrap/Constants.py
```diff
@@ -62,20 +62,32 @@ class ConfigSettings(EnumBackport):
         PATCHES_TO_INCLUDE = 'patchesToInclude'
         PATCHES_TO_EXCLUDE = 'patchesToExclude'
         MAINTENANCE_RUN_ID = 'maintenanceRunId'
+        PATCH_MODE = 'patchMode'
 
-    # File to save default settings for auto OS updates
+        # File to save default settings for auto OS updates
```
kjohn-msft, line 67: Why is this nested?

kjohn-msft:
_Read this later in our common doc for Windows and Linux, about ConfigurePatchingSummary substatus: "/This substatus will be shown in all operation to show the current configuration setting the machine/ " ---- Does this mean even Assessment or Installation requests will have ConfigurePatchingSummary substatus?_   --- **Yes, it must always have it.**

src/core/src/CoreMain.py
kjohn-msft: Also, the summary must always be written out even when it's an assessment or installation call..

src/core/src/CoreMain.py
kjohn-msft: This is just a comment in the context of the core process.

src/core/src/CoreMain.py
```diff
@@ -53,23 +53,27 @@ def __init__(self, argv):
...
-
-            # Patching + additional assessment occurs if the operation is 'Installation'
-            if patch_operation_requested == Constants.INSTALLATION.lower():
-                # setting current operation here, to include patch_installer init within installation actions, ensuring any exceptions during patch_installer init are added in installation summary errors object
-                status_handler.set_current_operation(Constants.INSTALLATION)
-                patch_installer = container.get('patch_installer')
-                patch_installation_successful = patch_installer.start_installation()
-                patch_assessment_successful = False
+            if patch_operation_requested == Constants.CONFIGURE_PATCHING.lower():
+                configure_patching.start_configure_patching()
```
kjohn-msft, line 63: configure patching must always happen.
assessment happens if it's not configure patching.
installation happens if it's installation.

Reorder code accordingly - only lines 62 and above need to change prior to the exception handling.

src/core/src/CoreMain.py
```diff
@@ -85,12 +89,17 @@ def __init__(self, argv):
...
+                if patch_operation_requested == Constants.CONFIGURE_PATCHING.lower():
+                    status_handler.set_configure_patching_substatus_json(status=Constants.STATUS_ERROR)
+                    composite_logger.log_debug('  -- Persisted failed configure patching substatus.')
+                else:
+                    if patch_operation_requested == Constants.INSTALLATION.lower() and not patch_installation_successful:
+                        status_handler.set_installation_substatus_json(status=Constants.STATUS_ERROR)
+                        composite_logger.log_debug('  -- Persisted failed installation substatus.')
+                    if not patch_assessment_successful:
+                        status_handler.set_assessment_substatus_json(status=Constants.STATUS_ERROR)
+                        composite_logger.log_debug('  -- Persisted failed assessment substatus.')
```
kjohn-msft, line 102: This works okay if it's not configure patching that failed and threw an exception in a non-configurepatching operation.

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -316,11 +316,15 @@ def read_with_retry(self, file_path_or_handle):
...
-                return value
+                try:
+                    value = file_handle.read()
+                    if was_path:  # what was passed in was not a file handle, so close the handle that was init here
+                        file_handle.close()
+                    self.__write_record(operation, code=0, output=value, delay=0)
+                    return value
+                except Exception as error:
+                    print("Unable to read from {0} (retries exhausted). Error: {1}.".format(str(file_path_or_handle), repr(error)))
+                    return None
```
kjohn-msft, line 327: This function is not meant to swallow exceptions like this. The caller must be prepared to handle exceptions on their own. What this function is supposed to do, however, is to actually do retries for MAX_FILE_OPERATION_RETRY_COUNT before raising the exception up. Replicate the retry and exception handling+raising logic as in write_with_retry. What caller was affected by this? Additional comments can be provided based on that.

src/core/src/core_logic/ConfigurePatching.py
```diff
@@ -0,0 +1,72 @@
...
+            self.status_handler.set_configure_patching_substatus_json(status=Constants.STATUS_SUCCESS, automatic_os_patch_state=current_auto_os_patch_state)
+
+        except Exception as error:
+            error_msg = 'Error processing configure patching request: ' + repr(error)
+            self.composite_logger.log_error(error_msg)
+            self.status_handler.add_error_to_status(error_msg, Constants.PatchOperationErrorCodes.DEFAULT_ERROR)
+            if Constants.ERROR_ADDED_TO_STATUS not in repr(error):
+                error.args = (error.args, "[{0}]".format(Constants.ERROR_ADDED_TO_STATUS))
+            self.status_handler.set_configure_patching_substatus_json(status=Constants.STATUS_ERROR)
+            raise
```
kjohn-msft, line 63: Review what the Windows extension is doing in this case and expand the common spec accordingly before applying to Linux. The documentation is lacking.

src/core/src/core_logic/ConfigurePatching.py
```diff
@@ -0,0 +1,72 @@
...
+        self.composite_logger = composite_logger
+        self.telemetry_writer = telemetry_writer
+        self.status_handler = status_handler
+
+        self.package_manager = package_manager
+
+    def start_configure_patching(self):
+        """ Start configure patching """
+        self.status_handler.set_current_operation(Constants.CONFIGURE_PATCHING)
+        self.raise_if_agent_incompatible()
```
kjohn-msft, line 35: This needs to be considered further as a configurepatching request is rarely likely to get the attention of a customer. Bring it up for discussion at scrum.

src/core/src/package_managers/YumPackageManager.py
kjohn-msft: +1; A comment is necessary here as it's a short-term thing, and not a legitimate reason for a permanent pass.

The practice is to raise NotImplementedException() in all these cases. The counterpoint is the code cannot be deployed with those exceptions in place - while that's exactly by design, and in many ways desirable, it reduces agility in this case.

src/core/tests/TestCoreMain.py
```diff
@@ -293,6 +293,31 @@ def test_installation_operation_fail_due_to_no_telemetry(self):
...
+        # # check telemetry events
+        # self.__check_telemetry_events(runtime)
+        #
+        # # check status file
+        # with runtime.env_layer.file_system.open(runtime.execution_config.status_file_path, 'r') as file_handle:
+        #     substatus_file_data = json.load(file_handle)[0]["status"]["substatus"]
+        # self.assertEquals(len(substatus_file_data), 1)
+        # self.assertTrue(substatus_file_data[0]["name"] == Constants.CONFIGURE_PATCHING_SUMMARY)
+        # self.assertTrue(substatus_file_data[0]["status"] == Constants.STATUS_SUCCESS.lower())
+        # runtime.stop()
```
kjohn-msft, line 315: marking todo

kjohn-msft (commented):
Completed cursory review -- in-depth on iterations closer to final. Re-iterating on past feedback - create and review a plan to break down PRs into distinct logical chunks that can go in sequentially, well before getting to the PR stage. Time to review increases exponentially as the PR breadth grows, and cost usually remains at those levels with each subsequent iteration.

src/core/src/core_logic/PatchAssessor.py
rane-rajasi: Resolved

src/core/src/CoreMain.py
rane-rajasi: Is cancelPatchInstallation the same as NoOperation? If yes, NoOperation request is managed by the extension handler, where it cancels any previously running core processes.

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: Resolved

src/core/src/package_managers/YumPackageManager.py
rane-rajasi: Pass in this case is because it isn't implemented yet. Added a note for the same

src/extension/src/EnableCommandHandler.py
rane-rajasi: Resolved

src/core/src/service_interfaces/StatusHandler.py
rane-rajasi: Implemented as per this guide in the latest iteration

src/core/src/CoreMain.py
rane-rajasi: Modified the code as per this and the next comment's instructions

src/core/src/bootstrap/Constants.py
rane-rajasi: Updated. This was likely done when we were still using DefaultForImage in CRP.

src/core/src/bootstrap/Constants.py
rane-rajasi: Resolved

src/core/src/CoreMain.py
rane-rajasi: Resolved.

src/core/src/CoreMain.py
rane-rajasi: Updated the error logging instructions, there is a todo item marked on in line 135 of CoreMain in the latest iteration. Need to follow up on that

src/core/src/core_logic/ConfigurePatching.py
rane-rajasi: ToDo

src/core/src/bootstrap/EnvLayer.py
rane-rajasi: ToDo

src/core/src/core_logic/ConfigurePatching.py
rane-rajasi: ToDo

src/core/tests/TestCoreMain.py
rane-rajasi: Resolved. Removed todo comment and added tests for validation

src/core/src/bootstrap/EnvLayer.py
rane-rajasi: Resolved. Updated the function per feedback and modified places where this was invoked to handle any raised exceptions. We already were doing that in old code, just one new code path added in this PR was not handling the exception: __get_current_auto_os_updates_setting_on_machine() in AptitudePackageManager

src/core/src/core_logic/ConfigurePatching.py
rane-rajasi: Resolved. Following the implemetation in Windows extension, i.e.
1.	Log error details in Configure Patching substatus
2.	Continue with rest of the Installation

NOTE: Windows does not mark the uber-operation as failed, if configure patching fails. Following the same approach in Linux. 

Q: Do we need to change this to mark the uber-operation as failed?

src/core/src/core_logic/ConfigurePatching.py
rane-rajasi: Resolved. 

Windows does not check for telemetry support. Created a ADO task for it. 
With the approach mentioned in the comment above, we log this error in Configure Patching substatus and continue with the uber-operation. So essentially, this check will fail again in assessment, where the operation will stop

kjohn-msft (approved):

PR #67, opened 2021-05-13
Installing all patches on an Auto Patching request on CentOS

rane-rajasi:
Includes:

- Change to install all available patches on CentOS for an Auto Patching request
- Unit test to verify the functionality in CentOS
- Unit test to verify Redhat machines are not affected by this change.
- Logs validating this change against different distros added in the task

### DIFF
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -114,13 +114,12 @@ class Constants(object):
     MAX_ASSESSMENT_RETRY_COUNT = 5
     MAX_INSTALLATION_RETRY_COUNT = 3
 
-    # Package Classifications
-    PACKAGE_CLASSIFICATIONS = {
-        0: 'Unclassified',           # doesn't serve a functional purpose in bit mask, but stands in for 'All' in code
-        1: 'Critical',
-        2: 'Security',
-        4: 'Other'
-    }
+    class PackageClassification(EnumBackport):
+        UNCLASSIFIED = 'Unclassified'
+        CRITICAL = 'Critical'
+        SECURITY = 'Security'
+        OTHER = 'Other'
+
     PKG_MGR_SETTING_FILTER_CRITSEC_ONLY = 'FilterCritSecOnly'
     PKG_MGR_SETTING_IDENTITY = 'PackageManagerIdentity'
     PKG_MGR_SETTING_IGNORE_PKG_FILTER = 'IgnorePackageFilter'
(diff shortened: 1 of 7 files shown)


src/core/src/package_managers/YumPackageManager.py
```diff
@@ -56,6 +56,14 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
         self.set_package_manager_setting(Constants.PKG_MGR_SETTING_IDENTITY, Constants.YUM)
         self.STR_TOTAL_DOWNLOAD_SIZE = "Total download size: "
 
+        # if an Auto Patching request comes in on a CentOS machine with Security and/or Critical classifications selected, we need to install all patches
+        installation_included_classifications = [] if execution_config.included_classifications_list is None else execution_config.included_classifications_list
+        if execution_config.maintenance_run_id is not None and execution_config.operation.lower() == Constants.INSTALLATION.lower() \
+                and 'CentOS' in str(env_layer.platform.linux_distribution()) \
+                and ('Critical' in installation_included_classifications or 'Security' in installation_included_classifications):
```
kjohn-msft, line 63: Critical and Security (not or; as auto patching will have both)..

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -56,6 +56,14 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
...
+            self.composite_logger.log_debug("Updating classifications list to install all patches for the Auto Patching request since classification based patching is not available on CentOS machines")
+            execution_config.included_classifications_list = [Constants.PACKAGE_CLASSIFICATIONS.__getitem__(1), Constants.PACKAGE_CLASSIFICATIONS.__getitem__(2), Constants.PACKAGE_CLASSIFICATIONS.__getitem__(4)]
```
kjohn-msft, line 65: That object was originally created that way to translate an integer representing all classifications (in a bitmask) into the actual list of classifications. In V2, that object is not necessary as it was before and heavily reflects its origin more than what purpose needs to be served today.

kjohn-msft (changes requested):
Please address commits.

src/core/src/bootstrap/Constants.py
```diff
@@ -114,13 +114,12 @@ class ConfigSettings(EnumBackport):
...
 
-    # Package Classifications
-    PACKAGE_CLASSIFICATIONS = {
-        0: 'Unclassified',           # doesn't serve a functional purpose in bit mask, but stands in for 'All' in code
-        1: 'Critical',
-        2: 'Security',
-        4: 'Other'
-    }
+    class PackageClassification(EnumBackport):
+        UNCLASSIFIED = 'Unclassified'
```
kjohn-msft, line 118: I don't expect this is needed anymore (it had a specific reason to be in the older dictionary).

kjohn-msft (approved):
Nit comment left inline and approved.

src/core/src/bootstrap/Constants.py
rane-rajasi: It's used in a few tests, will have to modify those and ensure everything works fine. Will change it in the next PR.

PR #69, opened 2021-06-02
Bug Fix: sudo not working because tty is required

rane-rajasi:
Bug fix for sudo status check failing due to tty set to required in /etc/sudoers, mostly observed in RHEL 72 and CentOS 7.2

Implementation details
- check whether tty is set to required either for current user or for all (last config among the 2 superseeds the other)
- we write our overriding config to disable tty (always for current user only) in /etc/sudoers.d/linuxpatchextension
- check if tty is disabled for current user in our custom sudoers file (i.e. in /etc/sudoers.d/linuxpatchextension)
   - if yes, do nothing, since the custom sudoers file will be read last and it's config overrides others
   - if no, write config to disable tty in custom sudoers, if set to required in /etc/sudoers
- Added unit tests to test different scenarios

**Point to note**: This action can only be performed by root user, since regular user doesn't have access to /etc/sudoers or /etc/sudoers.d. Agent team confirmed that agent is run with root access only so the extensions should also run with the same access. Ideally we don't need 'sudo' in our commands if agent runs the extension in root, but keeping that part of the code as is for now


src/core/src/bootstrap/EnvLayer.py
```diff
@@ -200,6 +207,85 @@ def check_sudo_status(self, raise_if_not_sudo=True):
...
+                else:
+                    defaults_set_for = re.search('(.*)' + self.require_tty_setting, setting).group(1)
+                    if self.get_current_user() in defaults_set_for or 'Defaults' == defaults_set_for.strip():
+                        tty_set_to_required = True
+
+            return tty_set_to_required
+
+        except Exception as error:
+            print("Error occurred while fetching data from [FilePath={0}] [Exception={1}]".format(str(self.etc_sudoers_file_path), repr(error)))
+            raise
```
kjohn-msft, line 252: Checking basic tty belongs here but the rest of it (mitigation and checking for mitigation) doesn't logically fall here. There is a logical class of things that address issues in customer environments safely and as a separate unit. This belongs in such a class. However, due to how primitive this is in the execution flow, it may deserve an exception to sit directly inside bootstrapper (all of it) - since it may be considered a basic, self-contained portion of bootstrapping.

kjohn-msft: The check_sudo_status function may also be moved into Bootstrapper going by that logical convention. It doesn't belong where it is currently looking at all of it now.

src/core/src/bootstrap/Bootstrapper.py
```diff
@@ -122,6 +122,8 @@ def basic_environment_health_check(self):
         self.composite_logger.log("Python version: " + " ".join(sys.version.splitlines()))
         self.composite_logger.log("Linux distribution: " + str(self.env_layer.platform.linux_distribution()) + "\n")
 
+        self.env_layer.ensure_tty_not_required()
+
         # Ensure sudo works in the environment
         sudo_check_result = self.env_layer.check_sudo_status()
```
kjohn-msft, line 128: It may make sense for both of these to actually be in the handler so the entire execution flow benefits from it long term.

kjohn-msft: (Sudo check may be left behind, and possibly duplicated).

kjohn-msft (changes requested):
Comments inline.

src/core/src/bootstrap/EnvLayer.py
rane-rajasi: Resolved

src/core/src/bootstrap/Bootstrapper.py
rane-rajasi: Resolved. Added EnvLayer.py and EnvManager.py in handler. EnvManager has check_sudo_status, tty mitigation and checking for mitigation. EnvLayer contains check for current tty settings on the machine. Also, copied a few other functions, that are required, from EnvLayer in Core to Handler

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -55,6 +57,11 @@ def __init__(self, real_record_path=None, recorder_enabled=False, emulator_enabl
         self.file_system = self.FileSystem(recorder_enabled, emulator_enabled, self.__write_record, self.__read_record,
                                            emulator_root_path=os.path.dirname(self.__real_record_path))
 
+        # components for tty config
+        self.etc_sudoers_file_path = "/etc/sudoers"
+        self.etc_sudoers_linux_patch_extension_file_path = "/etc/sudoers.d/linuxpatchextension"
+        self.require_tty_setting = "requiretty"
+
```
kjohn-msft, line 64: Left behind with 2 additional imports..

src/core/tests/TestTelemetryWriter.py
```diff
@@ -122,6 +122,11 @@ def test_write_to_new_file_if_event_file_limit_reached(self):
         telemetry_get_event_file_size_backup = self.runtime.telemetry_writer.get_file_size
         self.runtime.telemetry_writer.get_file_size = self.mock_get_file_size
 
+        # forcing wait of 1 sec to ensure new file is created, since we have mocked time.sleep in RuntimeComposer
+        init_time = time.time()
+        while time.time() < init_time + 1:
+            pass
+
```
kjohn-msft, line 129: This spools up the CPU to 100%; since it's single-threaded, it'll only be one core but we shouldn't do this.

When the mock is created, the original is backed up right. Have that reference available and sleep using that.

src/extension/src/EnvManager.py
```diff
@@ -0,0 +1,73 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Requires Python 2.7+
+
+
+class EnvManager(object):
```
kjohn-msft, line 18: Try to make this more specific as naming influences whether this becomes a dumping ground for things over time (bad), vs. forcing creation of specific classes as needed (good). EnvHealthManager for example.. or EnvIssueMitigator (not suitable because it doesn't allude to detection) or any others..

src/extension/src/__main__.py
```diff
@@ -46,6 +48,8 @@ def main(argv):
         runtime_context_handler = RuntimeContextHandler(logger)
         json_file_handler = JsonFileHandler(logger)
         ext_env_handler = ExtEnvHandler(json_file_handler)
+        env_layer = EnvLayer()
+        env_manager = EnvManager(env_layer)
```
kjohn-msft, line 52: Ensure there is a backlog item to convert all this into dependency injection like Core does it.

kjohn-msft (commented):
Comments inline.

src/extension/src/__main__.py
rane-rajasi: Added task# 10114420

src/core/src/bootstrap/EnvLayer.py
rane-rajasi: Removed

src/core/tests/TestTelemetryWriter.py
rane-rajasi: Updated - Using the original sleep method

src/extension/src/EnvManager.py
rane-rajasi: Used EnvHealthManager. I thought about EnvIssueMitigator while writing this code but it didn't completely justify the actions, so had stuck to EnvManager. EnvHealthManager sounds good

src/extension/src/__main__.py
```diff
@@ -49,7 +49,7 @@ def main(argv):
         json_file_handler = JsonFileHandler(logger)
         ext_env_handler = ExtEnvHandler(json_file_handler)
         env_layer = EnvLayer()
-        env_manager = EnvManager(env_layer)
+        env_manager = EnvHealthManager(env_layer)
```
kjohn-msft, line 52: env_health_manager -- consistency for ease of maintainability.

src/extension/tests/helpers/RuntimeComposer.py
```diff
@@ -16,7 +16,7 @@ def __init__(self):
         self.utility = Utility(self.logger)
         self.json_file_handler = JsonFileHandler(self.logger)
         self.env_layer = EnvLayer()
-        self.env_manager = EnvManager(self.env_layer)
+        self.env_manager = EnvHealthManager(self.env_layer)
```
kjohn-msft, line 19: env_health_manager

kjohn-msft (approved):
Minor comments inline on variable naming to match class name change. Can merge after fix.

src/extension/src/__main__.py
rane-rajasi: Resolved

src/extension/tests/helpers/RuntimeComposer.py
rane-rajasi: Resolved

PR #70, opened 2021-06-02
Bug fixes and improvements

rane-rajasi:
Contains:

- Fix for bug where the last status read in CRP for an installation operation has availablePatchSummary marked as 'In Progress'.
  This was due to the status for PatchInstallationSummary in status blob. CRP considers this status as the terminal state. So once PatchInstallationSummary is marked as 'success', CRP stops polling substatus and does not wait for the implicit (or 2nd) assessment to complete, if that is pending. 
Changed the implementation to only mark PatchInstallationSummary as 'success' once the implicit assessment is completed.
code change includes: removing substatus set to success in PatchInstaller and adding that in CoreMain after implicit assessment has competed

- Updated error message text for when Linux agent does not support telemetry, to ensure the error object within status blob includes the complete message
Code change in Constants.py

- For an installation operation, adding error details in installation substatus also if assessment fails. The error within installation substatus just mentions that an error occurred due to assessment failure and to refer the assessment substatus for more details.
Code change in CoreMain: Added a new variable to track overall installation substatus (var: overall_patch_installation_operation_successful), which includes implicit assessment, and using this to determine if failure occurred during assessment, in function update_patch_substatus_if_pending()

- Fix for adding error details in substatus, when failure occurs during core setup. 
Code change in CoreMain: Moved execution config read above any actions and adding all setup errors in assessment substatus. For an installation operation, a generic error ~ install failed due to assessment fail will be added in installation substatus due to the change in point 3 above.

### DIFF
--- a/src/core/src/CoreMain.py
+++ b/src/core/src/CoreMain.py
@@ -33,6 +33,7 @@ class CoreMain(object):
         patch_operation_requested = Constants.UNKNOWN
         patch_assessment_successful = False
         patch_installation_successful = False
+        overall_patch_installation_operation_successful = False
 
         try:
             # Level 2 bootstrapping
@@ -41,16 +42,17 @@ class CoreMain(object):
             lifecycle_manager, status_handler = bootstrapper.build_core_components(container)
             composite_logger.log_debug("Completed building out full container.\n\n")
 
-            # Basic environment check
-            bootstrapper.bootstrap_splash_text()
-            bootstrapper.basic_environment_health_check()
-            lifecycle_manager.execution_start_check()      # terminates if this instance shouldn't be running (redundant)
-
             # Execution config retrieval
             composite_logger.log_debug("Obtaining execution configuration...")
             execution_config = container.get('execution_config')
             telemetry_writer.set_operation_id(execution_config.activity_id)
             patch_operation_requested = execution_config.operation.lower()
+
+            # Basic environment check
+            bootstrapper.bootstrap_splash_text()
+            bootstrapper.basic_environment_health_check()
+            lifecycle_manager.execution_start_check()  # terminates if this instance shouldn't be running (redundant)
+
             patch_assessor = container.get('patch_assessor')
             package_manager = container.get('package_manager')
 
@@ -70,6 +72,11 @@ class CoreMain(object):
                 patch_assessment_successful = False
                 patch_assessment_successful = patch_assessor.start_assessment()
 
+                # PatchInstallationSummary to be marked as completed successfully only after the implicit (i.e. 2nd) assessment is completed, as per CRP's restrictions
+                if patch_assessment_successful and patch_installation_successful:
+                    patch_installer.mark_installation_completed()
+                    overall_patch_installation_operation_successful = True
+
         except Exception as error:
             # Privileged operation handling for non-production use
             if Constants.EnvLayer.PRIVILEGED_OP_MARKER in repr(error):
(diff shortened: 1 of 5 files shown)


src/core/src/bootstrap/Constants.py
```diff
@@ -189,8 +189,9 @@ class TelemetryEventLevel(EnumBackport):
         LogAlways = "LogAlways"
 
     TELEMETRY_TASK_NAME = "ExtensionCoreLog"
-    TELEMETRY_AT_AGENT_NOT_COMPATIBLE_ERROR_MSG = "The minimum Azure Linux Agent version prerequisite for Linux patching was not met. Please update the Azure Linux Agent on this machine following instructions here: http://aka.ms/UpdateLinuxAgent"
-    TELEMETRY_AT_AGENT_COMPATIBLE_MSG = "The minimum Azure Linux Agent version prerequisite for Linux patching was met."
+
+    TELEMETRY_AT_AGENT_NOT_COMPATIBLE_ERROR_MSG = "Minimum Linux Agent version prerequisite not met. Resolution: http://aka.ms/UpdateLinuxAgent"
```
kjohn-msft, line 193: 'Azure Linux Agent' instead of 'Linux Agent' everywhere...

rane-rajasi: Adding Azure truncates the error message in status blob

src/core/src/CoreMain.py
```diff
@@ -41,16 +42,17 @@ def __init__(self, argv):
...
             # Execution config retrieval
             composite_logger.log_debug("Obtaining execution configuration...")
             execution_config = container.get('execution_config')
             telemetry_writer.set_operation_id(execution_config.activity_id)
             patch_operation_requested = execution_config.operation.lower()
+
+            # Basic environment check
+            bootstrapper.bootstrap_splash_text()
+            bootstrapper.basic_environment_health_check()
+            lifecycle_manager.execution_start_check()  # terminates if this instance shouldn't be running (redundant)
```
kjohn-msft, line 54: Bootstrapper splash needs to be first. Execution start check also needs to come immediately after - we don't want/need telemetry from there to be associated with activity id. 
- Don't want because it'd have spurious entries associated with activity id and we also want the least amount of code to run prior to stopping. 
- Don't need because the telemetry is associated with the machine and can be retrieved linearly in case it's needed.

The basic environment health check can come later after execution config retrieval... You'd need to section and comment the sections appropriately.

src/core/src/CoreMain.py
```diff
@@ -84,19 +91,22 @@ def __init__(self, argv):
...
-                if not patch_assessment_successful:
-                    status_handler.set_assessment_substatus_json(status=Constants.STATUS_ERROR)
-                    composite_logger.log_debug('  -- Persisted failed assessment substatus.')
+                composite_logger.log_debug(' - Status handler pending writes flags [I=' + str(overall_patch_installation_operation_successful) + ', A=' + str(patch_assessment_successful) + ']')
 
+                # Current operation is set to either assessment or installation when these operations begin.
+                # If None is set at this point that is an indication that something went wrong before the first assessment operation could start.
+                # Logging such errors under assessment substatus since, error details within installation will point to assessment errors
+                if status_handler.get_current_operation() is None:
+                    status_handler.set_current_operation(Constants.ASSESSMENT)
```
kjohn-msft, line 100: Good change..

kjohn-msft: Actually, why can't this be set straight up near the beginning? The current operation (at first) will always be assessment. So it can be set right after status handler init (explicitly though, vs. default status handler initing to that). 
Is there a reason this need to happen here vs. that?

src/core/src/CoreMain.py
```diff
@@ -109,3 +119,16 @@ def __init__(self, argv):
...
+    def update_patch_substatus_if_pending(patch_operation_requested, overall_patch_installation_operation_successful, patch_assessment_successful, status_handler, composite_logger):
+        if patch_operation_requested == Constants.INSTALLATION.lower() and not overall_patch_installation_operation_successful:
+            if not patch_assessment_successful:
+                status_handler.set_current_operation(Constants.INSTALLATION)
+                status_handler.add_error_to_status("Installation failed due to assessment failure. Please refer the error details in assessment substatus")
+            status_handler.set_installation_substatus_json(status=Constants.STATUS_ERROR)
+            composite_logger.log_debug('  -- Persisted failed installation substatus.')
+        if not patch_assessment_successful:
+            status_handler.set_assessment_substatus_json(status=Constants.STATUS_ERROR)
+            composite_logger.log_debug('  -- Persisted failed assessment substatus.')
```
kjohn-msft, line 133: This would have been best encapsulated within the separate assessor and installer classes and invoked as needed. But that can't be done neatly as they may not have been initialized correctly. For completeness of positioning, it shouldn't be inside status handler as it's a level of detail that is not for status handler. It's not for CoreMain as it's usurping some logic that belongs elsewhere.

Maybe not for this PR, but consider what can be done for this.

Consider what may be done with this

src/core/src/bootstrap/Constants.py
kjohn-msft: "Unsupported older Azure Linux Agent version. To resolve: http://aka.ms/UpdateLinuxAgent" - 87 chars vs. 92 in PR.
 
The compatible message is less relevant but is short anyway so can have Azure in there.

kjohn-msft (changes requested):
Comments inline.

src/core/src/bootstrap/Constants.py
```diff
@@ -189,8 +189,9 @@ class TelemetryEventLevel(EnumBackport):
         LogAlways = "LogAlways"
 
     TELEMETRY_TASK_NAME = "ExtensionCoreLog"
-    TELEMETRY_AT_AGENT_NOT_COMPATIBLE_ERROR_MSG = "The minimum Azure Linux Agent version prerequisite for Linux patching was not met. Please update the Azure Linux Agent on this machine following instructions here: http://aka.ms/UpdateLinuxAgent"
-    TELEMETRY_AT_AGENT_COMPATIBLE_MSG = "The minimum Azure Linux Agent version prerequisite for Linux patching was met."
+
+    TELEMETRY_AT_AGENT_NOT_COMPATIBLE_ERROR_MSG = "Unsupported older Azure Linux Agent version. To resolve: http://aka.ms/UpdateLinuxAgent"
+    TELEMETRY_AT_AGENT_COMPATIBLE_MSG = "Minimum Linux Agent version prerequisite met"
```
kjohn-msft, line 194: This line is missing 'Azure' Linux Agent

kjohn-msft (approved):
Approving with a minor change requested (please address).

PR #71, opened 2021-06-15
Bug fix for error in telemetry error- execution could not complete, internal error

rane-rajasi:
Error reported in our logs is "Internal reporting error. Execution could not complete". Debugging revealed this happens while the extension is trying to access a file that was recently deleted by the agent. 
NOTE: Agent reads all events every 5 mins and deletes them at the end of each run, independent from our extension's run

Added a file exists check in all the possible missing places before processing the event file in TelemetryWriter.

### DIFF
--- a/src/core/src/service_interfaces/TelemetryWriter.py
+++ b/src/core/src/service_interfaces/TelemetryWriter.py
@@ -14,6 +14,7 @@
 #
 # Requires Python 2.7+
 import datetime
+import errno
 import json
 import os
 import re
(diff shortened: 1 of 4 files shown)


src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -273,8 +273,8 @@ def __get_events_dir_size(self):
         """ Returns total size, in bytes, of the events folder """
         total_dir_size = 0
         for f in os.listdir(self.events_folder_path):
-            if os.path.isfile(os.path.join(self.events_folder_path, f)):
-                total_dir_size += os.path.getsize(os.path.join(self.events_folder_path, f))
+            file_path = os.path.join(self.events_folder_path, f)
+            total_dir_size += os.path.getsize(file_path) if os.path.exists(file_path) and os.path.isfile(file_path) else 0
```
mirichmo, line 277: Is the exists check necessary? https://docs.python.org/3/library/os.path.html#os.path.isfile may do it implicitly if I am reading the documentation correctly.

amohammad-msft (commented):
changes look thorough. Don't have any additional comments, except +1 to Mike's. I don't think we need an exists check for the file_path in TelemetryWriter, both for src/core and src/extension

src/core/src/service_interfaces/TelemetryWriter.py
rane-rajasi: More investigation required, because the error re-appeared while trying this suggestion

kjohn-msft: Neither solution (first fix or comment on removing the exists check) guarantees the underlying issue will go away. 

You'll always be subject to timing issues in a path check + operation (as it's not atomic) and the thing that guarantees safety to do is to wrap getsize in a try catch - log further if the exception is not related to the file missing; and continue. Don't add anything to the dir size if it throws.

kjohn-msft (changes requested):
comments inline

src/core/src/service_interfaces/TelemetryWriter.py
rane-rajasi: Resolved

kjohn-msft (approved):

mirichmo (approved):

PR #73, opened 2021-07-01
Fix for SSL Certificate expired in RedHat and some code refactoring

rane-rajasi:
Contains
- Bug fix for SSL Certificate expired in RedHat
- Refactored ConfigurePatching class name to ConfigurePatchProcessor
- Unit tests for the fix

### DIFF
--- a/src/core/src/bootstrap/ConfigurationFactory.py
+++ b/src/core/src/bootstrap/ConfigurationFactory.py
@@ -20,7 +20,7 @@ import os
 from core.src.bootstrap.Constants import Constants
 from core.src.bootstrap.EnvLayer import EnvLayer
 
-from core.src.core_logic.ConfigurePatching import ConfigurePatching
+from core.src.core_logic.ConfigurePatchProcessor import ConfigurePatchProcessor
 from core.src.core_logic.ExecutionConfig import ExecutionConfig
 from core.src.core_logic.MaintenanceWindow import MaintenanceWindow
 from core.src.core_logic.PackageFilter import PackageFilter
@@ -192,7 +192,7 @@ class ConfigurationFactory(object):
                 'component_kwargs': {}
             },
             'configure_patching': {
-                'component': ConfigurePatching,
+                'component': ConfigurePatchProcessor,
                 'component_args': ['env_layer', 'execution_config', 'composite_logger', 'telemetry_writer', 'status_handler', 'package_manager'],
                 'component_kwargs': {}
             },
--- a/src/core/src/bootstrap/ConfigurationFactory.py
+++ b/src/core/src/bootstrap/ConfigurationFactory.py
@@ -20,7 +20,7 @@ import os
 from core.src.bootstrap.Constants import Constants
 from core.src.bootstrap.EnvLayer import EnvLayer
 
-from core.src.core_logic.ConfigurePatchProcessor import ConfigurePatchProcessor
+from core.src.core_logic.ConfigurePatchingProcessor import ConfigurePatchingProcessor
 from core.src.core_logic.ExecutionConfig import ExecutionConfig
 from core.src.core_logic.MaintenanceWindow import MaintenanceWindow
 from core.src.core_logic.PackageFilter import PackageFilter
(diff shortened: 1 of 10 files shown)


src/core/src/bootstrap/ConfigurationFactory.py
```diff
@@ -20,7 +20,7 @@
 from core.src.bootstrap.Constants import Constants
 from core.src.bootstrap.EnvLayer import EnvLayer
 
-from core.src.core_logic.ConfigurePatching import ConfigurePatching
+from core.src.core_logic.ConfigurePatchProcessor import ConfigurePatchProcessor
```
jemex, line 23: Consider ConfigurePatchManager since you have RebootManager

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -64,6 +64,13 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
             self.composite_logger.log_debug("Updating classifications list to install all patches for the Auto Patching request since classification based patching is not available on CentOS machines")
             execution_config.included_classifications_list = [Constants.PackageClassification.CRITICAL, Constants.PackageClassification.SECURITY, Constants.PackageClassification.OTHER]
 
+        # Known errors and the corresponding action items
+        self.known_errors_and_fixes = {"SSL peer rejected your certificate as expired": self.fix_ssl_certificate_issue,
+                                       "Error: Cannot retrieve repository metadata (repomd.xml) for repository": self.fix_ssl_certificate_issue,
```
jemex, line 69: Since the function for all the 3 errors are the same, how about a dictionary <string , List<String>>
which is the key is the function and the value is the list of errors the function fixes.

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -359,6 +374,38 @@ def update_os_patch_configuration_sub_setting(self, patch_configuration_sub_sett
         pass
     # endregion
 
+    # region Handling known errors
+    def mitigate_issue(self, output):
```
jemex, line 378: How about renaming this function to "try_mitigate_issue",

jemex (approved):

kjohn-msft:
Please merge latest from master.

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -72,6 +79,14 @@ def invoke_package_manager(self, command):
...
         self.composite_logger.log_debug('\nInvoking package manager using: ' + command)
         code, out = self.env_layer.run_command_output(command, False, False)
+
+        # check for known issues in the output and fix. Run the command again to get results after fix
+        if "Error" in out or "Errno" in out:
+            issue_mitigated = self.mitigate_issue(out)
+            if issue_mitigated:
+                self.composite_logger.log_debug('\nPost mitigation, invoking package manager again using: ' + command)
+                code, out = self.env_layer.run_command_output(command, False, False)
+
```
kjohn-msft, line 89: send code and out to a function to detect issues (and return) - it may cover more cases in the future.
[normalize all functions around this into the Package Manager class for consistency in naming and usage across package managers]
- log if issues were detected
- check if the issue falls into known issues (log)
       - try to mitigate (log outcome)
- repeat check until no known issues remain (current implementation assume single issue)

kjohn-msft (changes requested):
Comments inline. Please clarify if the comments need it..

src/core/src/bootstrap/ConfigurationFactory.py
rane-rajasi: ConfigurePatchingProcessor

rane-rajasi: Done

src/core/src/package_managers/YumPackageManager.py
rane-rajasi: Resolved

src/core/src/package_managers/YumPackageManager.py
rane-rajasi: Resolved

src/core/src/package_managers/YumPackageManager.py
rane-rajasi: Not making any code changes to this logic, since 

- For option 1 i.e. Jemal's suggestion, we would be placing an identifier in the values part of a dictionary, and would then have to iterate over all values (i.e identifiers) in the dictionary to get the key (function that fixes that bug). This approach is helps with visually grouping all similar issues but increases computations
- For option 2, Koshy's suggestion of ["a", "b", "c"] => (), we retain the existing logic of error texts in dictionary keys, but have to iterate over the multiple keys in an item. Plus we have to search for (eg) key "a" in an original error text "axyz" (returned by the command output)

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -72,6 +79,14 @@ def invoke_package_manager(self, command):
         """Get missing updates using the command input"""
```
jemex, line 79: just for my knowledge, is this comment or logging?

rane-rajasi: comment

kjohn-msft (approved):

PR #74, opened 2021-07-06
Fix for patch install marked failed before implicit assessment

rane-rajasi:
Related to the 1st point mentioned in the description/1st comment in this PR: https://github.com/Azure/LinuxPatchExtension/pull/70

Change was made earlier to only update patch installation substatus if the second/implicit assessment is completed. We missed out  one place where the patch install status is marked 'Failed' before implicit assessment is complete. This again leads to the same issue at CRP where the availablePatchSummary remains 'InProgress' since the last status CRP reads is when Patch Install is marked Failed and while assessment is in progress.

Changes in this PR:
- Removed setting patch install to failed within PatchInstaller, so patch install remains in Transitioning state 
- Calling update status in the success path of CoreMain, so patch install is marked appropriately in case of failure after implicit assessment

### DIFF
--- a/src/core/src/core_logic/PatchInstaller.py
+++ b/src/core/src/core_logic/PatchInstaller.py
@@ -95,10 +95,6 @@ class PatchInstaller(object):
         # Combining maintenance
         overall_patch_installation_successful = bool(update_run_successful and not maintenance_window_exceeded)
 
-        if not overall_patch_installation_successful:
-            self.status_handler.set_installation_substatus_json(status=Constants.STATUS_ERROR)
-            # NOTE: For auto patching requests, no need to report patch metadata to healthstore in case of failure
-
         return overall_patch_installation_successful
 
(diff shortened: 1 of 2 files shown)


src/core/src/core_logic/PatchInstaller.py
```diff
@@ -95,10 +95,6 @@ def start_installation(self, simulate=False):
         # Combining maintenance
         overall_patch_installation_successful = bool(update_run_successful and not maintenance_window_exceeded)
 
-        if not overall_patch_installation_successful:
```
mirichmo, line 98: Can you add a comment here stating why it is not marked error at this point?

rane-rajasi: Resolved

kjohn-msft (approved):

mirichmo (approved):

PR #77, opened 2021-08-13
Add started by and fix reboot always

kjohn-msft:


src/core/src/core_logic/RebootManager.py
```diff
@@ -104,6 +104,11 @@ def start_reboot_if_required_and_time_available(self, current_time_available):
             self.composite_logger.log(" - There was no reboot pending detected. Reboot is being skipped as it's not required, as per patch installation configuration (" + str(Constants.REBOOT_IF_REQUIRED) + ").")
             return False
 
+        # prevent repeated reboots
+        if self.reboot_setting == Constants.REBOOT_ALWAYS and not reboot_pending and self.status_handler.get_installation_reboot_status() == Constants.RebootStatus.COMPLETED:
+            self.composite_logger.log(" - At least one reboot has occurred, and there's no reboot pending, so the conditions for the 'Reboot Always' setting is fulfilled and reboot won't be repeated.")
+            return False
```
rane-rajasi, line 110: Should we also update rebootPending value in this case?

src/core/tests/Test_RebootManager.py
```diff
@@ -74,6 +74,22 @@ def test_reboot_always_time_available(self):
...
+        argument_composer = ArgumentComposer()
+        argument_composer.reboot_setting = reboot_setting_in_api
+        runtime = RuntimeCompositor(argument_composer.get_composed_arguments(), True, Constants.YUM)
+        reboot_manager = runtime.reboot_manager
+
+        # mock completing the reboot once, with no reboot required
+        runtime.status_handler.set_installation_reboot_status(Constants.RebootStatus.REQUIRED)
+        runtime.status_handler.set_installation_reboot_status(Constants.RebootStatus.STARTED)
+        runtime.status_handler.set_installation_reboot_status(Constants.RebootStatus.COMPLETED)
+
```
rane-rajasi, line 88: Is there a way to ensure reboot_pending is True, just so that this test always hits that scenario?

src/core/src/core_logic/RebootManager.py
kjohn-msft: reboot_pending is from the source of truth on the VM.. we don't manipulate it.

src/core/tests/Test_RebootManager.py
kjohn-msft: The test has been modified to go through the first reboot scenario and then no reboot with the same setting

rane-rajasi (approved):

PR #79, opened 2021-08-20
Bugfix while testing

nikhim-um:
Found this bug late during testing. Fixing it

image

### DIFF
--- a/src/core/src/bootstrap/ConfigurationFactory.py
+++ b/src/core/src/bootstrap/ConfigurationFactory.py
@@ -278,7 +278,6 @@ class ConfigurationFactory(object):
         for i in range(0, Constants.MAX_IMDS_CONNECTION_RETRY_COUNT):
             try:
                 res = urlreq.urlopen(request, timeout=2)
-                print(res.get_code())
                 if(res.getcode() == 200):
                     print("Connection to IMDS end point successfully established. VMCloudType is Azure\n")
                     return Constants.VMCloudType.AZURE


src/core/src/bootstrap/ConfigurationFactory.py
```diff
@@ -278,7 +278,6 @@ def get_vm_cloud_type(self):
         for i in range(0, Constants.MAX_IMDS_CONNECTION_RETRY_COUNT):
             try:
                 res = urlreq.urlopen(request, timeout=2)
-                print(res.get_code())
```
rane-rajasi, line 281: why remove print statement instead of correcting the function name? It's always good to have logs on API response. Will help debug when needed on a machine

nikhim-um:
Test 1: when there is exception:
Exception_Case

Test 2: when there is no exception:
image

src/core/src/bootstrap/ConfigurationFactory.py
```diff
@@ -278,19 +278,20 @@ def get_vm_cloud_type(self):
         for i in range(0, Constants.MAX_IMDS_CONNECTION_RETRY_COUNT):
             try:
```
rane-rajasi, line 279: Add a log in here with text "Connecting to IMDS endpoint..." or "Querying IMDS endpoint..."

src/core/src/bootstrap/ConfigurationFactory.py
```diff
@@ -278,19 +278,20 @@ def get_vm_cloud_type(self):
...
                 if(res.getcode() == 200):
                     print("Connection to IMDS end point successfully established. VMCloudType is Azure\n")
                     return Constants.VMCloudType.AZURE
                 else:
                     raise
-            except:
+            except Exception as error:
                 """ Failed to connect to Azure IMDS endpoint. This is expected on Arc machine - but not expected on Azure machine."""
+                print('Exception from IMDS connection http request: ' + repr(error))
+                print("Failed to connect to IMDS end point. [Retry Count={0}].".format(str(i)))
```
rane-rajasi, line 290: Retry Count=0 looks may raise some queries like it did for me. Change this to 
print("Failed to connect to IMDS end point. [Trial ={0}].".format(str(i+1)))

So that should show up in the first attempt as
"Failed to connect to IMDS end point. [Trial =1] this will go all the way upto 4

Log for the last attempt, under else, is good

Some existing logs for reference: 
image

PR #81, opened 2021-08-27
Add zypper process tree telemetry

benank:
If there is a process that is locking the zypper package manager, then its details (such as process name and call tree) are shown in the logs. Call tree is used because sometimes the name of the process is not enough to determine where the call originated from.

### DIFF
--- a/src/core/src/package_managers/ZypperPackageManager.py
+++ b/src/core/src/package_managers/ZypperPackageManager.py
@@ -16,7 +16,6 @@
 
 """ZypperPackageManager for SUSE"""
 import re
-import os
 from core.src.package_managers.PackageManager import PackageManager
 from core.src.bootstrap.Constants import Constants
 
@@ -43,7 +42,6 @@ class ZypperPackageManager(PackageManager):
 
         # Package manager exit code(s)
         self.zypper_exitcode_ok = 0
-        self.zypper_exitcode_zypper_locked = 7
         self.zypper_exitcode_zypper_updated = 103
 
         # Support to check for processes requiring restart
@@ -51,6 +49,7 @@ class ZypperPackageManager(PackageManager):
 
         # Miscellaneous
         self.set_package_manager_setting(Constants.PKG_MGR_SETTING_IDENTITY, Constants.ZYPPER)
+        self.zypper_get_process_tree_cmd = 'ps --forest -o pid,cmd -g $(ps -o sid= -p {})'
 
     def refresh_repo(self):
         self.composite_logger.log("Refreshing local repo...")
@@ -67,10 +66,9 @@ class ZypperPackageManager(PackageManager):
             self.composite_logger.log_warning(" - Return code from package manager: " + str(code))
             self.composite_logger.log_warning(" - Output from package manager: \n|\t" + "\n|\t".join(out.splitlines()))
 
-            if code == self.zypper_exitcode_zypper_locked:
-                process_lock_tree = self.get_process_tree_from_package_manager_msg(out)
-                if process_lock_tree is not None:
-                    self.composite_logger.log_warning(" - Process tree of package manager locking process: \n{}".format(process_lock_tree))
+            process_lock_tree = self.get_process_tree_from_package_manager_msg(out)
+            if process_lock_tree is not None:
+                self.composite_logger.log_warning(" - Process tree of package manager locking process: \n{}".format(str(process_lock_tree)))
 
             self.telemetry_writer.write_execution_error(command, code, out)
             error_msg = 'Unexpected return code (' + str(code) + ') from package manager on command: ' + command
@@ -87,14 +85,47 @@ class ZypperPackageManager(PackageManager):
             self.set_package_manager_setting(Constants.PACKAGE_MGR_SETTING_REPEAT_PATCH_OPERATION, True)
         return out
 
+    """
+        When given a string containing a pid, returns a multi-line string with a list of parent/child processes for that pid.
+        See pattern matching inside the function for details on specific strings that are accepted.
+        
+        Example:
+            input: 
+                message (string): Output from package manager: | System management is locked by the application with pid 7914 (/usr/bin/zypper).
+                
+            returns (string):
+                  PID CMD
+                 7736 /bin/bash
+                 7912  \_ python3 package_test.py
+                 7913  |   \_ sudo LANG=en_US.UTF8 zypper --non-interactive update --dry-run bind-utils
+                 7914  |       \_ zypper --non-interactive update --dry-run bind-utils
+                 7982  |           \_ /usr/bin/python3 /usr/lib/zypp/plugins/urlresolver/susecloud
+                 7984  |               \_ /usr/bin/python3 /usr/bin/azuremetadata --api latest --subscriptionId --billingTag --attestedData --signature
+                 7986  \_ python3 package_test.py
+                 8298      \_ sudo LANG=en_US.UTF8 zypper --non-interactive update --dry-run grub2-i386-pc
+    """
     def get_process_tree_from_package_manager_msg(self, message):
-        # Find pid xxxxx within output string
+        """
+            First find pid xxxxx within output string.
+
+            Example:
+                'Output from package manager: | System management is locked by the application with pid 7914 (/usr/bin/zypper).'
+
+            pid_substr_search will contain: ' pid 7914 '
+        """
         regex = re.compile(' pid \d+ ')
         pid_substr_search = regex.search(message)
         if pid_substr_search is None:
             return None
 
-        # Extract just pid text from substring
+        """
+            Now extract just pid text from pid_substr_search.
+            
+            Example (pid_substr_search): 
+                ' pid 7914 '
+                
+            pid_search will contain: '7914'
+        """
         regex = re.compile('\d+')
         pid_search = regex.search(pid_substr_search.group())
         if pid_search is None:
@@ -102,9 +133,14 @@ class ZypperPackageManager(PackageManager):
 
         pid = pid_search.group()
         # Gives a process tree so the calling process name(s) can be identified
-        process_tree = os.system("ps --forest -o pid,cmd -g $(ps -o sid= -p {})".format(pid))
+        get_process_tree_cmd = self.zypper_get_process_tree_cmd.format(str(pid))
+        code, out = self.env_layer.run_command_output(get_process_tree_cmd, False, False)
 
-        return process_tree
+        # Failed to get process tree
+        if code is not 0:
+            return None
+
+        return out
 
     # region Classification-based (incl. All) update check
     def get_all_updates(self, cached=False):
--- a/src/core/src/package_managers/ZypperPackageManager.py
+++ b/src/core/src/package_managers/ZypperPackageManager.py
@@ -86,13 +86,9 @@ class ZypperPackageManager(PackageManager):
         return out
 
     def get_process_tree_from_pid_in_output(self, message):
-        """ When given a string containing a pid, returns a multi-line string with a list of parent/child processes for that pid.
-            See pattern matching inside the function for details on specific strings that are accepted.
-
+        """ Fetches pid from the error message by searching for the text 'pid' and returns the process tree with all details.
             Example:
-                input:
-                    message (string): Output from package manager: | System management is locked by the application with pid 7914 (/usr/bin/zypper).
-
+                input: message (string): Output from package manager: | System management is locked by the application with pid 7914 (/usr/bin/zypper).
                 returns (string):
                       PID CMD
                      7736 /bin/bash
@@ -102,26 +98,19 @@ class ZypperPackageManager(PackageManager):
                      7982  |           \_ /usr/bin/python3 /usr/lib/zypp/plugins/urlresolver/susecloud
                      7984  |               \_ /usr/bin/python3 /usr/bin/azuremetadata --api latest --subscriptionId --billingTag --attestedData --signature
                      7986  \_ python3 package_test.py
-                     8298      \_ sudo LANG=en_US.UTF8 zypper --non-interactive update --dry-run grub2-i386-pc
-        """
+                     8298      \_ sudo LANG=en_US.UTF8 zypper --non-interactive update --dry-run grub2-i386-pc """
 
         """ First find pid xxxxx within output string.
-
             Example: 'Output from package manager: | System management is locked by the application with pid 7914 (/usr/bin/zypper).'
-
-            pid_substr_search will contain: ' pid 7914'
-        """
+            pid_substr_search will contain: ' pid 7914' """
         regex = re.compile(' pid \d+')
         pid_substr_search = regex.search(message)
         if pid_substr_search is None:
             return None
 
         """ Now extract just pid text from pid_substr_search.
-            
-            Example (pid_substr_search): ' pid 7914'
-                
-            pid_search will contain: '7914'
-        """
+            Example (pid_substr_search): ' pid 7914'   
+            pid_search will contain: '7914' """
         regex = re.compile('\d+')
         pid_search = regex.search(pid_substr_search.group())
         if pid_search is None:
@@ -138,11 +127,7 @@ class ZypperPackageManager(PackageManager):
         code, out = self.env_layer.run_command_output(get_process_tree_cmd, False, False)
 
         # Failed to get process tree
-        if code != 0:
-            return None
-
-        # Also failed to get process tree
-        if len(out) == 0:
+        if code != 0 or len(out) == 0:
             return None
 
         return out
--- a/src/core/src/package_managers/ZypperPackageManager.py
+++ b/src/core/src/package_managers/ZypperPackageManager.py
@@ -66,9 +66,9 @@ class ZypperPackageManager(PackageManager):
             self.composite_logger.log_warning(" - Return code from package manager: " + str(code))
             self.composite_logger.log_warning(" - Output from package manager: \n|\t" + "\n|\t".join(out.splitlines()))
 
-            process_lock_tree = self.get_process_tree_from_package_manager_msg(out)
-            if process_lock_tree is not None:
-                self.composite_logger.log_warning(" - Process tree of package manager locking process: \n{}".format(str(process_lock_tree)))
+            process_tree = self.get_process_tree_from_pid_in_output(out)
+            if process_tree is not None:
+                self.composite_logger.log_warning(" - Process tree for the pid in output: \n{}".format(str(process_tree)))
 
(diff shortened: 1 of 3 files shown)


src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -64,6 +65,11 @@ def invoke_package_manager(self, command):
             self.composite_logger.log('[ERROR] Package manager was invoked using: ' + command)
             self.composite_logger.log_warning(" - Return code from package manager: " + str(code))
             self.composite_logger.log_warning(" - Output from package manager: \n|\t" + "\n|\t".join(out.splitlines()))
+
+            process_lock_tree = self.get_process_tree_from_package_manager_msg(out)
```
rane-rajasi, line 69: rename get_process_tree_from_package_manager_**msg** to get_process_tree_from_pid_in_output

rename process_lock_tree to **process_tree**

Variable name should reflect what it stores, or in this case what the function returns. In our example right now, a process is locked, but this is a generic function, merely searching for a pid in the output and returning additional details regarding that process. We don't only look for a locked process message in the output.

src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -64,6 +65,11 @@ def invoke_package_manager(self, command):
...
+            if process_lock_tree is not None:
+                self.composite_logger.log_warning(" - Process tree of package manager locking process: \n{}".format(str(process_lock_tree)))
```
rane-rajasi, line 71: Again, don't mention lock since the process can be for some other error message.

self.composite_logger.log_warning(" - Process tree for the pid in output: \n{}".format(str(process_tree)))

src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -79,6 +85,67 @@ def invoke_package_manager(self, command):
...
+                 7986  \_ python3 package_test.py
+                 8298      \_ sudo LANG=en_US.UTF8 zypper --non-interactive update --dry-run grub2-i386-pc
+    """
+    def get_process_tree_from_package_manager_msg(self, message):
+        """
+            First find pid xxxxx within output string.
+
+            Example:
+                'Output from package manager: | System management is locked by the application with pid 7914 (/usr/bin/zypper).'
+
```
rane-rajasi, line 113: remove extra line break

src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -79,6 +85,67 @@ def invoke_package_manager(self, command):
...
+                 7914  |       \_ zypper --non-interactive update --dry-run bind-utils
+                 7982  |           \_ /usr/bin/python3 /usr/lib/zypp/plugins/urlresolver/susecloud
+                 7984  |               \_ /usr/bin/python3 /usr/bin/azuremetadata --api latest --subscriptionId --billingTag --attestedData --signature
+                 7986  \_ python3 package_test.py
+                 8298      \_ sudo LANG=en_US.UTF8 zypper --non-interactive update --dry-run grub2-i386-pc
+    """
+    def get_process_tree_from_package_manager_msg(self, message):
+        """
+            First find pid xxxxx within output string.
+
```
rane-rajasi, line 110: remove line break

src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -79,6 +85,67 @@ def invoke_package_manager(self, command):
...
+                  PID CMD
+                 7736 /bin/bash
+                 7912  \_ python3 package_test.py
+                 7913  |   \_ sudo LANG=en_US.UTF8 zypper --non-interactive update --dry-run bind-utils
+                 7914  |       \_ zypper --non-interactive update --dry-run bind-utils
+                 7982  |           \_ /usr/bin/python3 /usr/lib/zypp/plugins/urlresolver/susecloud
+                 7984  |               \_ /usr/bin/python3 /usr/bin/azuremetadata --api latest --subscriptionId --billingTag --attestedData --signature
+                 7986  \_ python3 package_test.py
+                 8298      \_ sudo LANG=en_US.UTF8 zypper --non-interactive update --dry-run grub2-i386-pc
+    """
```
rane-rajasi, line 106: Ideally we shouldn't need the entire process tree info. Just details about the specific process mentioned in output should be good. For eg in this case just the process details for pid 7914. 

For now, let's keep this as is but add a todo comment in this function, stating we may want to revisit this later are reduce the result to specific process details only. 

Move the entire comment within the function

src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -79,6 +85,67 @@ def invoke_package_manager(self, command):
...
+            pid_substr_search will contain: ' pid 7914'
+        """
+        regex = re.compile(' pid \d+')
+        pid_substr_search = regex.search(message)
+        if pid_substr_search is None:
+            return None
+
+        """
+            Now extract just pid text from pid_substr_search.
+
```
rane-rajasi, line 123: remove line break

src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -79,6 +85,67 @@ def invoke_package_manager(self, command):
...
+        pid_substr_search = regex.search(message)
+        if pid_substr_search is None:
+            return None
+
+        """
+            Now extract just pid text from pid_substr_search.
+            
+            Example (pid_substr_search): 
+                ' pid 7914'
+
```
rane-rajasi, line 126: remove line break

src/core/tests/Test_ZypperPackageManager.py
```diff
@@ -183,5 +184,51 @@ def test_install_package_failure(self):
...
+
+        package_manager = self.container.get('package_manager')
+        self.assertIsNotNone(package_manager)
+
+        # Get this process' pid to test output with a real running process
+        test_process_pid = os.getpid()
+        self.assertIsNotNone(test_process_pid)
+
+        # Create example package manager message and include this process' pid
+        package_manager_msg = 'Output from package manager: | System management is locked by the application with pid {} (/usr/bin/zypper).'
```
rane-rajasi, line 198: package_manager_output instead of package_manager_msg

src/core/tests/Test_ZypperPackageManager.py
```diff
@@ -183,5 +184,51 @@ def test_install_package_failure(self):
...
         self.assertEquals(package_manager.install_update_and_dependencies('selinux-policy.noarch', '3.13.1-102.el7_3.16', simulate=True), Constants.FAILED)
 
+    def test_get_process_tree_from_package_manager_msg_success(self):
+        self.runtime.set_legacy_test_type('HappyPath')
+
+        package_manager = self.container.get('package_manager')
+        self.assertIsNotNone(package_manager)
+
+        # Get this process' pid to test output with a real running process
+        test_process_pid = os.getpid()
```
rane-rajasi, line 194: We never run unit tests on live examples. Always use mocks. Assume a pid and also the return message from the get_process-tree_cmd. See LegacyEnvLayerExtensions.py for examples, there are examples of mocking response for various commands in there

Also, make sure to run all tests within the tests folder in both /core/tests and /extension/tests with python 2.7 and any one version of python 3.x (I use python 3.8). Pycharm supports using multiple Python versions. All tests should pass (existing and new) in both versions

src/core/tests/Test_ZypperPackageManager.py
```diff
@@ -183,5 +184,51 @@ def test_install_package_failure(self):
...
+        self.assertIsNotNone(package_manager.get_process_tree_from_package_manager_msg(package_manager_msg))
+
+    def test_get_process_tree_from_package_manager_msg_failure_nonexistent_process(self):
+        self.runtime.set_legacy_test_type('HappyPath')
+
+        package_manager = self.container.get('package_manager')
+        self.assertIsNotNone(package_manager)
+
+        # Test with a nonexistent pid
+        test_process_pid = -1
```
rane-rajasi, line 211: Similar to the comment above, you need to mock the response on running get_process_tree_command

src/core/src/package_managers/ZypperPackageManager.py
benank: That makes sense, thanks. Renamed!

src/core/src/package_managers/ZypperPackageManager.py
benank: Agreed. There might be a better way that I didn't see to get detailed information about a specific process, but I found that the process tree was the best for now because the output would print `/usr/bin/zypper` for the process name no matter if zypper was executed from the python script or from the terminal.

Moved the comment and added a todo.

src/core/tests/Test_ZypperPackageManager.py
benank: Understood! I've made some changes that now use mocks based on the examples.

I've also run all the tests in both those locations with both versions and they've passed (except for the one unrelated one).

src/core/src/package_managers/ZypperPackageManager.py
rane-rajasi: By line break, I meant, remove the extra blank line. Remove blank line on line 91. Try to only have blank lines between different logic parts in code, and avoid additional blank lines in other places

src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -79,6 +85,68 @@ def invoke_package_manager(self, command):
...
         return out
 
+    def get_process_tree_from_pid_in_output(self, message):
+        """ When given a string containing a pid, returns a multi-line string with a list of parent/child processes for that pid.
+            See pattern matching inside the function for details on specific strings that are accepted.
+
+            Example:
+                input:
+                    message (string): Output from package manager: | System management is locked by the application with pid 7914 (/usr/bin/zypper).
+
```
rane-rajasi, line 95: remove blank line on line 95

src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -79,6 +85,68 @@ def invoke_package_manager(self, command):
...
+                     7913  |   \_ sudo LANG=en_US.UTF8 zypper --non-interactive update --dry-run bind-utils
+                     7914  |       \_ zypper --non-interactive update --dry-run bind-utils
+                     7982  |           \_ /usr/bin/python3 /usr/lib/zypp/plugins/urlresolver/susecloud
+                     7984  |               \_ /usr/bin/python3 /usr/bin/azuremetadata --api latest --subscriptionId --billingTag --attestedData --signature
+                     7986  \_ python3 package_test.py
+                     8298      \_ sudo LANG=en_US.UTF8 zypper --non-interactive update --dry-run grub2-i386-pc
+        """
+
+        """ First find pid xxxxx within output string.
+
```
rane-rajasi, line 109: remove blank line on line 109

src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -79,6 +85,68 @@ def invoke_package_manager(self, command):
...
+            Example: 'Output from package manager: | System management is locked by the application with pid 7914 (/usr/bin/zypper).'
+
```
rane-rajasi, line 111: remove blank line on line 111

src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -79,6 +85,68 @@ def invoke_package_manager(self, command):
...
+            pid_substr_search will contain: ' pid 7914'
+        """
+        regex = re.compile(' pid \d+')
+        pid_substr_search = regex.search(message)
+        if pid_substr_search is None:
+            return None
+
+        """ Now extract just pid text from pid_substr_search.
+
```
rane-rajasi, line 120: remove blank line on line 120

src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -79,6 +85,68 @@ def invoke_package_manager(self, command):
...
+        """
+        regex = re.compile(' pid \d+')
+        pid_substr_search = regex.search(message)
+        if pid_substr_search is None:
+            return None
+
+        """ Now extract just pid text from pid_substr_search.
+            
+            Example (pid_substr_search): ' pid 7914'
+
```
rane-rajasi, line 122: remove blank line on line 122

src/core/tests/Test_ZypperPackageManager.py
```diff
@@ -183,5 +184,42 @@ def test_install_package_failure(self):
...
+        self.runtime.set_legacy_test_type('HappyPath')
+
+        package_manager = self.container.get('package_manager')
+        self.assertIsNotNone(package_manager)
+
+        # Create example package manager message and include test pid
+        package_manager_output = 'Output from package manager: | System management is locked by the application with pid 7914 (/usr/bin/zypper).'
+
+        # Test to make sure a valid string was returned with process information
+        self.assertIsNotNone(package_manager.get_process_tree_from_pid_in_output(package_manager_output))
```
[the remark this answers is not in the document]
rane-rajasi: Enable compatibility checks on all python versions in PyCharm, to get indications of such compatibility issues:

In PyCharm, goto File -> Settings

image

src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -79,6 +85,68 @@ def invoke_package_manager(self, command):
...
+        # TODO: consider revisiting in the future to reduce the result to this pid only instead of entire tree
+        get_process_tree_cmd = self.zypper_get_process_tree_cmd.format(str(pid))
+        code, out = self.env_layer.run_command_output(get_process_tree_cmd, False, False)
+
+        # Failed to get process tree
+        if code != 0:
+            return None
+
+        # Also failed to get process tree
+        if len(out) == 0:
```
rane-rajasi, line 145: the 2 ifs can be combined into one:

if code !=0 or len(out) ==0:
   return None

src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -79,6 +85,68 @@ def invoke_package_manager(self, command):
             self.set_package_manager_setting(Constants.PACKAGE_MGR_SETTING_REPEAT_PATCH_OPERATION, True)
         return out
 
+    def get_process_tree_from_pid_in_output(self, message):
+        """ When given a string containing a pid, returns a multi-line string with a list of parent/child processes for that pid.
```
rane-rajasi, line 89: Change the description, to what the function does given the input. Something similar to:

""" Fetches pid from the error message, by searching for the text 'pid' and returns the process tree with all details. """

Remove this statement: "See pattern matching inside the function for details on specific strings that are accepted."
as we are calling this function for all errored instances in zypper. Instead, since we are only searching for the text 'pid' in the input, above statement should suffice

src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -79,6 +87,25 @@ def invoke_package_manager(self, command):
...
+    def get_process_tree_from_package_manager_msg(self, message):
+        # Find pid xxxxx within output string
+        regex = re.compile(' pid \d+ ')
+        pid_substr_search = regex.search(message)
+        if pid_substr_search is None:
+            return None
+
+        # Extract just pid text from substring
+        regex = re.compile('\d+')
+        pid_search = regex.search(pid_substr_search.group())
```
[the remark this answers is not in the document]
rane-rajasi: Since group returns a match, shouldn't it be used after we run either a match or search on the string? re.compile just prepares the pattern we need to search/match. regex.search() is actually going to find that pattern in the string. 

regex.search(pid_substr_search.group())
group used before regex.search is not achieving anything, since it doesn't try to match any pattern with the string during group and before search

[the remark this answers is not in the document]
benank: Yep, it should be used after we run a match or search on the string, and it is in this case. `pid_substr_search` is the search result from using `regex.search` on the original string that was passed in, so group is needed to extract the string found.

Above this code segment are these these lines:
```py
regex = re.compile(' pid \d+')
pid_substr_search = regex.search(message)
```

src/core/tests/library/LegacyEnvLayerExtensions.py
```diff
@@ -215,6 +215,17 @@ def run_command_output(self, cmd, no_output=False, chk_err=True):
                                  "\n" + \
                                  "You may wish to restart these processes.\n" + \
                                  "See 'man zypper' for information about the meaning of values in the above table.\n"
+                    elif cmd.find('ps --forest -o pid,cmd -g $(ps -o sid= -p 7914)') > -1:
```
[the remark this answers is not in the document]
benank: Indentation looks good on my end
image

src/core/tests/Test_ZypperPackageManager.py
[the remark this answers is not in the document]
benank: Thanks, I've enabled compatibility inspection now. 

`assertIsNotNone()` is used in every test in that file - why can it not be used to validate? I can see that it's usually used to verify the package manager, but never used to validate actual results at the end of a test. Why is that?

src/core/src/package_managers/ZypperPackageManager.py
benank: Thanks, will do!

src/core/src/package_managers/ZypperPackageManager.py
benank: Got it, thanks!

src/core/tests/Test_ZypperPackageManager.py
```diff
@@ -183,5 +184,42 @@ def test_install_package_failure(self):
...
+
+    def test_get_process_tree_from_package_manager_output_failure_nonexistent_process(self):
+        self.runtime.set_legacy_test_type('SadPath')
```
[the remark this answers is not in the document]
benank: > Please run the 3 tests you have added for this log under debug mode, one at a time and make sure they follow the expected code path within the function

Thanks for this tip, I wasn't familiar with how to debug in pycharm. I've done so and have verified that they are following the expected code path in the function.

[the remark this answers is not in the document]
benank: > Don't see the mock output for get process tree command under SadPath

I didn't add any because it's expected to not have any output. However, after some testing, that actually won't be the case. The process tree command will always return some sort of process tree, even if the pid doesn't belong to a valid process. I've added a search for the pid in the process tree returned to verify that it actually contains the pid used in the command.

src/core/src/package_managers/ZypperPackageManager.py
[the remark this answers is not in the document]
benank: Resolved offline

rane-rajasi (approved):

PR #82, opened 2021-08-27
Multiple bug fixes identified in auto Assessment and a few syntax corrections

rane-rajasi:
Contains:

- Bug fix for summaries from previous operation not preserved after auto assessment in status file. For eg: if the operation order was ConfigurePatching followed by Auto Assessment, status file after Auto Assessment should contain PatchAssessmentSummary (refreshed by Auto Assessment) and ConfigurePatchingSummary.
Included unit tests to cover this usecase

- Bug fix for main/outer operation name overwritten to Assessment by Auto Assessment. For eg: if the operation order was ConfigurePatching followed by Auto Assessment, operation name in status file after Auto Assessment should be 'ConfigurePatching'
 Included unit test to cover this

- Bug fix for reading boolean string values correctly. bool(<str>) which was used earlier, will always return True, even for bool('False')

- Bug fix for reading 'processIds' from CoreState.json. Were using the wrong key earlier

- Syntax corrections in many places as per standard practice

- Tests were taking longer to execute due to IMDS connection retries and were initializing LifecycleManagerArc for Azure tests. Fixed by mocking connection to set cloud type as Azure

- Added additional logging in a few places

- Added unit test for ConfigurePatching with different configs for patchMode and assessmentMode

### DIFF
--- a/src/core/src/core_logic/ExecutionConfig.py
+++ b/src/core/src/core_logic/ExecutionConfig.py
@@ -36,7 +36,7 @@ class ExecutionConfig(object):
         self.sequence_number = self.__get_value_from_argv(self.execution_parameters, Constants.ARG_SEQUENCE_NUMBER)
         self.environment_settings = self.__get_decoded_json_from_argv(self.execution_parameters, Constants.ARG_ENVIRONMENT_SETTINGS)
         self.config_settings = self.__get_decoded_json_from_argv(self.execution_parameters, Constants.ARG_CONFIG_SETTINGS)
-        self.exec_auto_assess_only = True if (self.__get_value_from_argv(self.execution_parameters, Constants.ARG_AUTO_ASSESS_ONLY, False)).lower() == 'true' else False
+        self.exec_auto_assess_only = (self.__get_value_from_argv(self.execution_parameters, Constants.ARG_AUTO_ASSESS_ONLY, False)).lower() == 'true'
 
         # Environment Settings
         self.composite_logger.log_debug(" - Parsing environment settings...")
--- a/src/core/src/core_logic/ExecutionConfig.py
+++ b/src/core/src/core_logic/ExecutionConfig.py
@@ -97,7 +97,7 @@ class ExecutionConfig(object):
         if default_value == Constants.DEFAULT_UNSPECIFIED_VALUE:
             raise Exception("Unable to find key {0} in core arguments: {1}.".format(key, str(argv)))
         else:
-            return default_value
+            return str(default_value)
 
     def __get_decoded_json_from_argv(self, argv, key):
         """ Discovers and decodes the JSON body of a specific base64 encoded JSON object in input arguments. """
--- a/src/core/src/core_logic/ExecutionConfig.py
+++ b/src/core/src/core_logic/ExecutionConfig.py
@@ -36,7 +36,7 @@ class ExecutionConfig(object):
         self.sequence_number = self.__get_value_from_argv(self.execution_parameters, Constants.ARG_SEQUENCE_NUMBER)
         self.environment_settings = self.__get_decoded_json_from_argv(self.execution_parameters, Constants.ARG_ENVIRONMENT_SETTINGS)
         self.config_settings = self.__get_decoded_json_from_argv(self.execution_parameters, Constants.ARG_CONFIG_SETTINGS)
-        self.exec_auto_assess_only = bool(self.__get_value_from_argv(self.execution_parameters, Constants.ARG_AUTO_ASSESS_ONLY, False))
+        self.exec_auto_assess_only = True if (self.__get_value_from_argv(self.execution_parameters, Constants.ARG_AUTO_ASSESS_ONLY, False)).lower() == 'true' else False
 
         # Environment Settings
         self.composite_logger.log_debug(" - Parsing environment settings...")
@@ -78,7 +78,6 @@ class ExecutionConfig(object):
 
     def __transform_execution_config_for_auto_assessment(self):
         self.composite_logger.log_debug("Setting execution configuration values for auto assessment.")
-        self.operation = Constants.AUTO_ASSESSMENT
         self.activity_id = str(uuid.uuid4())
         self.included_classifications_list = self.included_package_name_mask_list = self.excluded_package_name_mask_list = []
         self.maintenance_run_id = None
--- a/src/core/src/service_interfaces/LifecycleManagerArc.py
+++ b/src/core/src/service_interfaces/LifecycleManagerArc.py
@@ -65,7 +65,7 @@ class LifecycleManagerArc(LifecycleManager):
 
             # attempted sequence number is same as recorded core sequence - expected
             if int(self.execution_config.sequence_number) == int(core_sequence['number']):
-                if core_sequence['completed'].lower() == 'false':
+                if core_sequence['completed'].lower() == 'true':
                     self.composite_logger.log_debug("Auto-assessment is safe to start. Existing sequence number marked as completed.")
                     self.update_core_sequence(completed=False)  # signalling core restart with auto-assessment as its safe to do so
                 else:
--- a/src/core/src/service_interfaces/LifecycleManagerArc.py
+++ b/src/core/src/service_interfaces/LifecycleManagerArc.py
@@ -30,9 +30,9 @@ class LifecycleManagerArc(LifecycleManager):
         # Handshake file paths
         self.ext_state_file_path = os.path.join(self.execution_config.config_folder, Constants.EXT_STATE_FILE)
         self.core_state_file_path = os.path.join(self.execution_config.config_folder, Constants.CORE_STATE_FILE)
-        #Writing to log
+        # Writing to log
         self.composite_logger.log_debug("Initializing LifecycleManagerArc")
-        #Variables
+        # Variables
         self.arc_root = "/var/lib/waagent/"
         self.arc_core_state_file_name = "CoreState.json"
         self.arc_extension_folder_pattern = "Microsoft.SoftwareUpdateManagement.LinuxOsUpdateExtension-*"
@@ -50,7 +50,7 @@ class LifecycleManagerArc(LifecycleManager):
             self.update_core_sequence(completed=True)   # forced-to-complete scenario | extension wrapper will be watching for this event
             self.env_layer.exit(0)
 
-        if self.execution_config.operation == Constants.AUTO_ASSESSMENT:
+        if self.execution_config.exec_auto_assess_only:
             # newer sequence number has been observed, do not run
             if int(self.execution_config.sequence_number) < int(extension_sequence['number']) \
                     or int(self.execution_config.sequence_number) < int(core_sequence['number']):
@@ -65,12 +65,12 @@ class LifecycleManagerArc(LifecycleManager):
 
             # attempted sequence number is same as recorded core sequence - expected
             if int(self.execution_config.sequence_number) == int(core_sequence['number']):
-                if bool(core_sequence['completed']):
+                if core_sequence['completed'].lower() == 'false':
                     self.composite_logger.log_debug("Auto-assessment is safe to start. Existing sequence number marked as completed.")
                     self.update_core_sequence(completed=False)  # signalling core restart with auto-assessment as its safe to do so
                 else:
                     self.composite_logger.log_debug("Auto-assessment may not be safe to start yet as core sequence is not marked completed.")
-                    if len(self.identify_running_processes(core_sequence['process_ids'])) != 0:
+                    if len(self.identify_running_processes(core_sequence['processIds'])) != 0:
                         # NOT SAFE TO START
                         # Possible reasons: full core operation is in progress (okay), some previous auto-assessment is still running (bad scheduling, adhoc run, or process stalled)
                         self.composite_logger.log_warning("Auto-assessment is NOT safe to start yet. Existing core process(es) running. Exiting. [LastHeartbeat={0}][Operation={1}]".format(str(core_sequence['lastHeartbeat']), str(core_sequence['action'])))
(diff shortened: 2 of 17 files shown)


src/core/tests/Test_ConfigurePatchingProcessor.py
```diff
@@ -170,6 +172,91 @@ def test_operation_fail_for_configure_patching_request_for_apt(self):
...
+        self.__check_telemetry_events(runtime)
+
+        # check status file for configure patching patch mode
+        with runtime.env_layer.file_system.open(runtime.execution_config.status_file_path, 'r') as file_handle:
+            substatus_file_data = json.load(file_handle)[0]["status"]["substatus"]
+
+        # check status file for configure patching patch state
+        self.assertEquals(len(substatus_file_data), 1)
+        self.assertTrue(substatus_file_data[0]["name"] == Constants.CONFIGURE_PATCHING_SUMMARY)
+        self.assertTrue(substatus_file_data[0]["status"] == Constants.STATUS_SUCCESS.lower())
```
benank, line 206: nit: `.lower()` should be on the left hand side since it is a constant on the right

src/core/tests/Test_ConfigurePatchingProcessor.py
```diff
@@ -170,6 +172,91 @@ def test_operation_fail_for_configure_patching_request_for_apt(self):
...
+
+        # check status file for configure patching patch mode
+        with runtime.env_layer.file_system.open(runtime.execution_config.status_file_path, 'r') as file_handle:
+            substatus_file_data = json.load(file_handle)[0]["status"]["substatus"]
+
+        # check status file for configure patching patch state
+        self.assertTrue(runtime.package_manager.image_default_patch_configuration_backup_exists())
+        self.assertEquals(len(substatus_file_data), 1)
+        self.assertTrue(substatus_file_data[0]["name"] == Constants.CONFIGURE_PATCHING_SUMMARY)
+        self.assertTrue(substatus_file_data[0]["status"] == Constants.STATUS_SUCCESS.lower())
```
benank, line 249: nit: same as [here](https://github.com/Azure/LinuxPatchExtension/pull/82/files#r697770396)

src/core/tests/Test_CoreMain.py
```diff
@@ -432,6 +435,214 @@ def test_install_only_critical_and_security_packages_for_redhat_autopatching(sel
...
+        # check telemetry events
+        self.__check_telemetry_events(runtime)
+        # check status file
+        with runtime.env_layer.file_system.open(runtime.execution_config.status_file_path, 'r') as file_handle:
+            status_file_data = json.load(file_handle)[0]["status"]
+        self.assertTrue(status_file_data["operation"] == Constants.CONFIGURE_PATCHING)
+        substatus_file_data = status_file_data["substatus"]
+        self.assertEquals(len(substatus_file_data), 1)
+        self.assertTrue(substatus_file_data[0]["name"] == Constants.CONFIGURE_PATCHING_SUMMARY)
+        self.assertTrue(substatus_file_data[0]["status"] == Constants.STATUS_SUCCESS.lower())
```
benank, line 458: nit: same as [here](https://github.com/Azure/LinuxPatchExtension/pull/82/files#r697770396)

src/core/tests/Test_ConfigurePatchingProcessor.py
rane-rajasi: Do you mean it should be this way: 

self.assertTrue(Constants.STATUS_SUCCESS.lower() == substatus_file_data[0]["status"]) ? or something else?

If you mean something like this lower(Constants.STATUS_SUCCESS), that's is not a function in python.

src/core/src/core_logic/ExecutionConfig.py
```diff
@@ -36,7 +36,7 @@ def __init__(self, env_layer, composite_logger, execution_parameters):
         self.sequence_number = self.__get_value_from_argv(self.execution_parameters, Constants.ARG_SEQUENCE_NUMBER)
         self.environment_settings = self.__get_decoded_json_from_argv(self.execution_parameters, Constants.ARG_ENVIRONMENT_SETTINGS)
         self.config_settings = self.__get_decoded_json_from_argv(self.execution_parameters, Constants.ARG_CONFIG_SETTINGS)
-        self.exec_auto_assess_only = bool(self.__get_value_from_argv(self.execution_parameters, Constants.ARG_AUTO_ASSESS_ONLY, False))
+        self.exec_auto_assess_only = True if (self.__get_value_from_argv(self.execution_parameters, Constants.ARG_AUTO_ASSESS_ONLY, False)).lower() == 'true' else False
```
benank, line 39: May be less verbose to use 
```py
(self.__get_value_from_argv(self.execution_parameters, Constants.ARG_AUTO_ASSESS_ONLY, False)).lower() == 'true'
```
instead, but I can see why the current method might be more clear. Unless there is another reason why the current method has been used?

src/core/tests/Test_CoreMain.py
```diff
@@ -432,6 +435,214 @@ def test_install_only_critical_and_security_packages_for_redhat_autopatching(sel
...
+        self.__check_telemetry_events(runtime)
+        # check status file
+        with runtime.env_layer.file_system.open(runtime.execution_config.status_file_path, 'r') as file_handle:
+            status_file_data = json.load(file_handle)[0]["status"]
+        # verifying the original operation name is preserved
+        self.assertTrue(status_file_data["operation"] == Constants.CONFIGURE_PATCHING)
+        substatus_file_data = status_file_data["substatus"]
+        self.assertEquals(len(substatus_file_data), 2)
+        self.assertTrue(substatus_file_data[0]["name"] == Constants.PATCH_ASSESSMENT_SUMMARY)
+        self.assertTrue(substatus_file_data[0]["status"] == Constants.STATUS_SUCCESS.lower())
```
benank, line 487: nit: same as [here](https://github.com/Azure/LinuxPatchExtension/pull/82/files#r697770396)

src/core/tests/Test_CoreMain.py
```diff
@@ -432,6 +435,214 @@ def test_install_only_critical_and_security_packages_for_redhat_autopatching(sel
...
+        # check started by set to 'Platform'
+        self.assertTrue(json.loads(substatus_file_data[0]["formattedMessage"]["message"])['startedBy'], Constants.PatchAssessmentSummaryStartedBy.PLATFORM)
+        # verifying the older operation summary is preserved
+        self.assertTrue(substatus_file_data[1]["name"] == Constants.CONFIGURE_PATCHING_SUMMARY)
+        self.assertTrue(substatus_file_data[1]["status"] == Constants.STATUS_SUCCESS.lower())
```
benank, line 492: nit: same as [here](https://github.com/Azure/LinuxPatchExtension/pull/82/files#r697770396)

src/core/tests/Test_CoreMain.py
```diff
@@ -432,6 +435,214 @@ def test_install_only_critical_and_security_packages_for_redhat_autopatching(sel
...
+        message = json.loads(substatus_file_data[1]["formattedMessage"]["message"])
+        self.assertEqual(message["automaticOsPatchState"], Constants.AutomaticOsPatchStates.DISABLED)  # auto OS updates are disabled in RuntimeCompositor
+        # check status file for configure patching assessment state
+        message = json.loads(substatus_file_data[1]["formattedMessage"]["message"])
+        self.assertEqual(message["autoAssessmentStatus"]["autoAssessmentState"], Constants.AutoAssessmentStates.ENABLED)  # auto assessment is enabled
+
+        runtime.stop()
```
benank, line 499: Question: why is the `tearDown` method not used for this test class? I can see that most other test classes have it and use it

benank (approved):

src/core/tests/Test_ConfigurePatchingProcessor.py
benank: I was thinking something like this:

```py
self.assertTrue(substatus_file_data[0]["status"].lower() == Constants.STATUS_SUCCESS)
```

rane-rajasi: That wouldn't work because Constants.STATUS_SUCCESS is set to 'Success' and the one on left is already 'success'. 

With this: self.assertTrue(substatus_file_data[0]["status"].lower() == Constants.STATUS_SUCCESS)
the equation would be 'success' == 'Success', which will fail

But you've raised a good point. For equality we need to ensure both values are exactly same in terms of casing too. So we should have lower() on both sides just in case, to avoid any case mismatch

benank: Oh, I see, yep that makes sense. Having it on both sides sounds like a good idea to avoid any other case mismatches.

benank (approved):

src/core/tests/Test_ConfigurePatchingProcessor.py
```diff
@@ -170,6 +172,91 @@ def test_operation_fail_for_configure_patching_request_for_apt(self):
...
+        self.__check_telemetry_events(runtime)
+
+        # check status file for configure patching patch mode
+        with runtime.env_layer.file_system.open(runtime.execution_config.status_file_path, 'r') as file_handle:
+            substatus_file_data = json.load(file_handle)[0]["status"]["substatus"]
+
+        # check status file for configure patching patch state
+        self.assertTrue(runtime.package_manager.image_default_patch_configuration_backup_exists())
+        self.assertEquals(len(substatus_file_data), 1)
+        self.assertTrue(substatus_file_data[0]["name"] == Constants.CONFIGURE_PATCHING_SUMMARY)
```
mirichmo, line 248: Shouldn't these be `self.assertEquals(Constants.CONFIGURE_PATCHING_SUMMARY, substatus_file_data[0]["name"])`? Using the form self.assertEquals( expectedValue, actualValue )
Usually the built-in error messages are more helpfully stated such as "expected X does not equal actual Y" instead of "Does not equal true" (leads to the question, what was not true?)

Applies throughout the test code

src/core/tests/Test_CoreMain.py
```diff
@@ -432,6 +435,214 @@ def test_install_only_critical_and_security_packages_for_redhat_autopatching(sel
...
+        self.assertEquals(len(substatus_file_data), 2)
+        self.assertTrue(substatus_file_data[0]["name"] == Constants.PATCH_ASSESSMENT_SUMMARY)
+        self.assertTrue(substatus_file_data[0]["status"] == Constants.STATUS_SUCCESS.lower())
+        # check started by set to 'Platform'
+        self.assertTrue(json.loads(substatus_file_data[0]["formattedMessage"]["message"])['startedBy'], Constants.PatchAssessmentSummaryStartedBy.PLATFORM)
+        # verifying the older operation summary is preserved
+        self.assertTrue(substatus_file_data[1]["name"] == Constants.CONFIGURE_PATCHING_SUMMARY)
+        self.assertTrue(substatus_file_data[1]["status"] == Constants.STATUS_SUCCESS.lower())
+        # check status file for configure patching auto updates state
+        message = json.loads(substatus_file_data[1]["formattedMessage"]["message"])
```
mirichmo, line 560: Verifying that the lastModifiedDateTime hasn't changed would be good too

mirichmo (changes requested):

src/core/tests/Test_CoreMain.py
rane-rajasi: We traditionally use teardown to discard any configuration, env setup needed for test execution. Whenever the env setup/config is common between multiple tests in a file we use the setup and teardown create and discard setup for each test.

The runtime configuration in this test file varies between different tests (see values set in argument_composer), so there each test is responsible for setting and discarding it's own environment. See comment on line 33 under setUp()

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -523,15 +523,18 @@ def __load_status_file_components(self, initial_load=False):
...
-                errors = self.__installation_summary_json['errors']
-                if errors is not None and errors['details'] is not None:
-                    self.__installation_errors = errors['details']
-                    self.__installation_total_error_count = self.__get_total_error_count_from_prev_status(errors['message'])
+                if self.execution_config.exec_auto_assess_only:
+                    self.__installation_substatus_json = status_file_data['status']['substatus'][i]
+                else:
+                    message = status_file_data['status']['substatus'][i]['formattedMessage']['message']
+                    self.__installation_summary_json = json.loads(message)
+                    self.__installation_packages = self.__installation_summary_json['patches']
```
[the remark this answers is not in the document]
rane-rajasi: No, the else part here is used to preserve details, for non Auto Assessment operations, whenever and whichever properties, where required. For eg: on a reboot, that happens in between an Installation operation, we need to preserve previous state data such as any error details, reboot status, etc in PatchInstallationSummary, everything under PatchAssessmentSummary (which will be overwritten by implicit Assessment later), everything under ConfigurePatchSummary and so on. 

If part was added since we need to preserve all contents within these summaries from the previous iteration for the same seq.

src/core/src/core_logic/ExecutionConfig.py
rane-rajasi: Updated

src/core/tests/Test_ConfigurePatchingProcessor.py
rane-rajasi: This one applies to a lot of tests all across core and extension handler. Created a different task(Tracking ID: 10735197) to address this comment

src/core/tests/Test_CoreMain.py
```diff
@@ -17,7 +17,10 @@
 import json
 import os
 import re
+import time
```
benank, line 20: import is not used

src/core/tests/Test_CoreMain.py
benank: Gotcha, thanks for the info! That makes sense.

benank (approved):

src/core/src/service_interfaces/LifecycleManagerArc.py
```diff
@@ -65,12 +65,12 @@ def execution_start_check(self):
 
             # attempted sequence number is same as recorded core sequence - expected
             if int(self.execution_config.sequence_number) == int(core_sequence['number']):
-                if bool(core_sequence['completed']):
+                if core_sequence['completed'].lower() == 'false':
```
nikhim-um, line 68: Looks like logic is wrong here. it should be core_sequence['completed'].lower() == 'true'  -> which means previous operation on the same sequence number is completed and current autoassessment can proceed ( safe to start ) else Its not safe to start auto assessment

nikhim-um: same if statement is translated properly in LifecycleManagerAzure.py  line number 58

rane-rajasi: Changed. This is a good example of the importance of having unit tests along with the code change. I caught this issue from a UT for Azure and missed adding the same in Arc. A UT in Arc would have caught this.

benank (approved):

mirichmo (approved):

PR #89, opened 2021-09-14
Ensure events folder exists if folder path is in config

benank:
Please let me know if you would like me to add UTs for this change

### DIFF
--- a/src/core/src/service_interfaces/TelemetryWriter.py
+++ b/src/core/src/service_interfaces/TelemetryWriter.py
@@ -38,7 +38,10 @@ class TelemetryWriter(object):
         self.start_time_for_event_count_throttle_check = datetime.datetime.utcnow()
         self.event_count = 1
 
-        if events_folder_path is not None and os.path.exists(events_folder_path):
+        if events_folder_path is not None:
+            if not os.path.exists(events_folder_path):
+                os.mkdir(events_folder_path) # ensure events folder exists, even though telemetry should have created it
+
             self.events_folder_path = events_folder_path
             self.__is_agent_compatible = True
 
--- a/src/core/src/service_interfaces/TelemetryWriter.py
+++ b/src/core/src/service_interfaces/TelemetryWriter.py
@@ -38,10 +38,7 @@ class TelemetryWriter(object):
         self.start_time_for_event_count_throttle_check = datetime.datetime.utcnow()
         self.event_count = 1
 
-        if events_folder_path is not None:
-            if not os.path.exists(events_folder_path):
-                os.mkdir(events_folder_path) # ensure events folder exists, even though telemetry should have created it
-
+        if events_folder_path is not None and os.path.exists(events_folder_path):
             self.events_folder_path = events_folder_path
             self.__is_agent_compatible = True
 
--- a/src/extension/src/ActionHandler.py
+++ b/src/extension/src/ActionHandler.py
@@ -87,10 +87,14 @@ class ActionHandler(object):
     def setup_telemetry(self):
         # check if events folder exists, if it does init telemetry, if events folder does not exist, log that telemetry is not supported by agent since events folder does not exist
         events_folder = self.ext_env_handler.events_folder
-        if events_folder is None or not os.path.exists(events_folder):
+        if events_folder is None:
             err_msg = Constants.TELEMETRY_AT_AGENT_NOT_COMPATIBLE_ERROR_MSG
             self.logger.log_error(err_msg)
         else:
+            if not os.path.exists(events_folder):
+                os.mkdir(events_folder)
+                self.logger.log("Events folder path found in HandlerManifest but does not exist on disk. Creating now. [Path={0}]".format(str(events_folder)))
+            
             self.logger.log(Constants.TELEMETRY_AT_AGENT_COMPATIBLE_MSG)
             self.telemetry_writer.events_folder_path = events_folder
             # As this is a common function used by all handler actions, setting operation_id such that it will be the same timestamp for all handler actions, which can be used for identifying all events for an operation.
--- a/src/extension/src/ActionHandler.py
+++ b/src/extension/src/ActionHandler.py
@@ -93,7 +93,7 @@ class ActionHandler(object):
         else:
             if not os.path.exists(events_folder):
                 os.mkdir(events_folder)
-                self.logger.log("Events folder path found in HandlerManifest but does not exist on disk. Creating now. [Path={0}]".format(str(events_folder)))
+                self.logger.log("Events folder path found in HandlerEnvironment but does not exist on disk. Creating now. [Path={0}]".format(str(events_folder)))
             
             self.logger.log(Constants.TELEMETRY_AT_AGENT_COMPATIBLE_MSG)
             self.telemetry_writer.events_folder_path = events_folder
--- a/src/extension/src/TelemetryWriter.py
+++ b/src/extension/src/TelemetryWriter.py
@@ -66,8 +66,12 @@ class TelemetryWriter(object):
     def write_event(self, message, event_level=Constants.TelemetryEventLevel.Informational, task_name=Constants.TELEMETRY_TASK_NAME):
         """ Creates and writes event to event file after validating none of the telemetry size restrictions are breached """
         try:
-            if self.events_folder_path is None or not os.path.exists(self.events_folder_path) or not Constants.TELEMETRY_ENABLED_AT_EXTENSION:
+
+            if self.events_folder_path is None or not Constants.TELEMETRY_ENABLED_AT_EXTENSION:
                 return
+
+            if not os.path.exists(self.events_folder_path):
+                os.mkdir(self.events_folder_path)
 
             self.__delete_older_events()
 
--- a/src/extension/src/TelemetryWriter.py
+++ b/src/extension/src/TelemetryWriter.py
@@ -70,9 +70,6 @@ class TelemetryWriter(object):
             if self.events_folder_path is None or not Constants.TELEMETRY_ENABLED_AT_EXTENSION:
                 return
 
-            if not os.path.exists(self.events_folder_path):
-                os.mkdir(self.events_folder_path)
-
             self.__delete_older_events()
 
             event = self.__new_event_json(event_level, message, task_name)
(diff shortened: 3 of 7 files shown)


src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -38,7 +38,10 @@ def __init__(self, env_layer, composite_logger, events_folder_path):
         self.start_time_for_event_count_throttle_check = datetime.datetime.utcnow()
         self.event_count = 1
 
-        if events_folder_path is not None and os.path.exists(events_folder_path):
+        if events_folder_path is not None:
+            if not os.path.exists(events_folder_path):
+                os.mkdir(events_folder_path) # ensure events folder exists, even though telemetry should have created it
```
kjohn-msft, line 43: Please do this in extension handler (/src/ext/src) when the value is first read.

Core code remains the same, and should fail as it does today if the handler did not help tide over the agent issue.

kjohn-msft (changes requested):
Comment inline

src/core/src/service_interfaces/TelemetryWriter.py
kjohn-msft: Add quick UT in ext handler to ensure creation if missing.

benank: Thanks, moved to TelemetryWriter in extension.

benank:
Added UTs

src/extension/src/TelemetryWriter.py
```diff
@@ -66,9 +66,13 @@ def __ensure_message_restriction_compliance(self, full_message):
...
         """ Creates and writes event to event file after validating none of the telemetry size restrictions are breached """
         try:
-            if self.events_folder_path is None or not os.path.exists(self.events_folder_path) or not Constants.TELEMETRY_ENABLED_AT_EXTENSION:
+
+            if self.events_folder_path is None or not Constants.TELEMETRY_ENABLED_AT_EXTENSION:
                 return
 
+            if not os.path.exists(self.events_folder_path):
+                os.mkdir(self.events_folder_path)
+
```
kjohn-msft, line 75: Please sync with @rane-rajasi for a better place to put this. The write-event function is called a bunch of times during execution and this (folder existence and creation if necessary) should only be evaluated once during the start of each execution.

kjohn-msft (changes requested):
@rane-rajasi, please cover a review pass on this first.

src/extension/src/TelemetryWriter.py
rane-rajasi: Had a discussion over call, this should be moved to setup_telemetry() in ActionHandler.py. Please also test this on a VM

benank: Moved to `setup_telemetry()` in ActionHandler.py.

src/extension/tests/Test_ActionHandler.py
```diff
@@ -214,8 +214,14 @@ def test_telemetry_not_available(self):
...
-        # event folder is set within HandlerEnvironment but the directory path is invalid. i.e. Telemetry setup not valid
-        self.action_handler.ext_env_handler.events_folder = "testfolder"
+        self.action_handler.ext_env_handler.events_folder = events_folder_path_backup
+
+    def test_telemetry_available_events_folder_not_exists(self):
+        # handler actions will continue to execute after logging telemetry not suported message
+        events_folder_path_backup = self.action_handler.ext_env_handler.events_folder
+        
+        # events folder is set within HandlerEnvironment but the directory path is invalid, so Telemetry is setup but events folder wasn't created
+        self.action_handler.ext_env_handler.events_folder = os.path.join(tempfile.mkdtemp(), "testfolder")
```
kjohn-msft, line 224: informational to @rane-rajasi also -- the temp folder creation logic sprinkled around the test code results in folders getting created (with files) in the user's temp directory, and they're not being cleaned up by the tests concerned.

src/extension/src/TelemetryWriter.py
```diff
@@ -66,7 +66,8 @@ def __ensure_message_restriction_compliance(self, full_message):
     def write_event(self, message, event_level=Constants.TelemetryEventLevel.Informational, task_name=Constants.TELEMETRY_TASK_NAME):
         """ Creates and writes event to event file after validating none of the telemetry size restrictions are breached """
         try:
-            if self.events_folder_path is None or not os.path.exists(self.events_folder_path) or not Constants.TELEMETRY_ENABLED_AT_EXTENSION:
+
```
kjohn-msft, line 69: Excess new line added here.

src/extension/src/ActionHandler.py
```diff
@@ -87,10 +87,14 @@ def tear_down(self):
     def setup_telemetry(self):
         # check if events folder exists, if it does init telemetry, if events folder does not exist, log that telemetry is not supported by agent since events folder does not exist
         events_folder = self.ext_env_handler.events_folder
-        if events_folder is None or not os.path.exists(events_folder):
+        if events_folder is None:
             err_msg = Constants.TELEMETRY_AT_AGENT_NOT_COMPATIBLE_ERROR_MSG
             self.logger.log_error(err_msg)
         else:
+            if not os.path.exists(events_folder):
```
kjohn-msft, line 94: informational: elif: would have been functionally the same.

kjohn-msft (approved):

src/extension/src/ActionHandler.py
```diff
@@ -87,10 +87,14 @@ def tear_down(self):
...
+                os.mkdir(events_folder)
+                self.logger.log("Events folder path found in HandlerManifest but does not exist on disk. Creating now. [Path={0}]".format(str(events_folder)))
```
rane-rajasi, line 96: Correction: "Events folder path found in HandlerEnvironment.json but the folder itself does not exist on disk..."

HandlerManifest is a different file containing some other data.

src/extension/src/TelemetryWriter.py
```diff
@@ -66,7 +66,8 @@ def __ensure_message_restriction_compliance(self, full_message):
...
+            if self.events_folder_path is None or not Constants.TELEMETRY_ENABLED_AT_EXTENSION:
```
rane-rajasi, line 70: Keep the events_folder exists check as is.

src/extension/tests/Test_ActionHandler.py
rane-rajasi: We need to remove any temp directory created by UTs. All other parts of the UTs, where a temp dir is created, is followed by a cleanup at the end. 

For eg: the shutil.rmtree() call here
image

@kjohn-msft this part should delete all temp directories. I think only when a test fails, the temp directory cleanup is not carried out

src/extension/src/TelemetryWriter.py
benank: Will do.

src/extension/src/ActionHandler.py
benank: Thanks, fixed.

src/extension/src/TelemetryWriter.py
benank: Fixed.

src/extension/tests/Test_ActionHandler.py
benank: Most (if not all) existing tests do cleanup test folder creation. I changed this logic to clean up the test folder created.

src/extension/tests/Test_ActionHandler.py
```diff
@@ -214,8 +214,14 @@ def test_telemetry_not_available(self):
         with self.assertRaises(SystemExit) as sys_exit:
             self.action_handler.enable()
 
-        # event folder is set within HandlerEnvironment but the directory path is invalid. i.e. Telemetry setup not valid
-        self.action_handler.ext_env_handler.events_folder = "testfolder"
+        self.action_handler.ext_env_handler.events_folder = events_folder_path_backup
```
rane-rajasi, line 217: remove line 217. Earlier check was for the case when event folder path is set in HandlerEnvironment but does not contain a valid path. Which we are fixing by creating the events folder. So the test 'test_telemetry_not_available' will only validate when events folder path is not given.

Rename the new test below from 'test_telemetry_available_events_folder_not_exists' to 'test_telemetry_available_but_events_folder_not_exists'

src/extension/src/ActionHandler.py
benank: Not sure if I understand what you mean here. if the else was replaced with elif, it would only perform the logic inside if the events folder didn't exist, instead of performing the logic whether it existed or not.

src/extension/src/TelemetryWriter.py
```diff
@@ -66,6 +66,7 @@ def __ensure_message_restriction_compliance(self, full_message):
     def write_event(self, message, event_level=Constants.TelemetryEventLevel.Informational, task_name=Constants.TELEMETRY_TASK_NAME):
         """ Creates and writes event to event file after validating none of the telemetry size restrictions are breached """
         try:
+
```
rane-rajasi, line 69: Still has the excess new line

src/extension/src/ActionHandler.py
kjohn-msft: _Re: it would only perform the logic inside if the events folder didn't exist, instead of performing the logic whether it existed or not._
-- Which is exactly the same as what's happening with line 95 and 96 now. If there were other statements outside of the 'if' block in what is line 94 in your PR, what you're concerned about would be true.

src/extension/tests/Test_ActionHandler.py
benank: Gotcha, fixed and updated

src/extension/src/TelemetryWriter.py
benank: Fixed

src/extension/src/ActionHandler.py
rane-rajasi: Is it the ones on line 98 and 99? Which are in else but will not be in "if not os.path.exists(events_folder):"

kjohn-msft: Ignore -- indentation was misunderstood for 98, 99.

kjohn-msft (approved):

rane-rajasi (approved):

PR #92, opened 2021-09-20
Add retries for zypper package manager repo_refresh

benank:
- Fixed off-by-one issues in retry conditionals


src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -54,7 +55,23 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
...
+                    self.composite_logger.log_warning("Exception on package manager refresh repo. [Exception={0}] [RetryCount={1}]".format(repr(error), str(i)))
+                    time.sleep(i + 1)
+                else:
+                    error_msg = "Unable to refresh repo (retries exhausted). [{0}] [RetryCount={1}]".format(repr(error), str(i))
+                    self.composite_logger.log_warning(error_msg)
+                    self.status_handler.add_error_to_status(error_msg, Constants.PatchOperationErrorCodes.PACKAGE_MANAGER_FAILURE)
+
+                    # Reboot if possible and not already done (TODO: find a way to detect if a reboot has been done already)
+                    self.composite_logger.log_warning("Setting force_reboot flag to True.")
+                    self.force_reboot = True
```
benank, line 73: Unsure of how to check if we have already rebooted from a past flag set here.

I am thinking that something like this could work:
```py
if self.status_handler.get_installation_reboot_status() == Constants.RebootStatus.COMPLETED:
```
but am unsure if other factors could affect it, such as other things causing a reboot before this reboot is triggered.

src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -72,8 +89,7 @@ def invoke_package_manager(self, command):
 
             self.telemetry_writer.write_execution_error(command, code, out)
             error_msg = 'Unexpected return code (' + str(code) + ') from package manager on command: ' + command
-            self.status_handler.add_error_to_status(error_msg, Constants.PatchOperationErrorCodes.PACKAGE_MANAGER_FAILURE)
-            raise Exception(error_msg, "[{0}]".format(Constants.ERROR_ADDED_TO_STATUS))
+            raise Exception(error_msg)
```
kjohn-msft, line 92: Why was this changed? This shouldn't be...

src/core/src/bootstrap/Constants.py
```diff
@@ -160,6 +160,7 @@ class AutoAssessmentStates(EnumBackport):
     MAX_ASSESSMENT_RETRY_COUNT = 5
     MAX_INSTALLATION_RETRY_COUNT = 3
     MAX_IMDS_CONNECTION_RETRY_COUNT = 5
+    MAX_REPO_REFRESH_RETRY_COUNT = 3
```
kjohn-msft, line 163: MAX_ZYPPER_REPO_REFRESH_RETRY_COUNT  since we are only doing it for one.

src/core/src/package_managers/ZypperPackageManager.py
benank: Line 75 was removed because I felt that the error should only be reported if all the retries failed.

Because line 75 was removed, the status handler needs to update it in the new spot where the exception is raised, so the `ERROR_ ADDED_TO_STATUS` flag is added in the new exception location on the new line 74. If this was kept as-is, it would still report an error even if a retry succeeded.

benank:
Some tests are failing now after the off by one fix so I am looking into those now.

src/core/src/bootstrap/Constants.py
benank: Done

src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -54,7 +55,23 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
...
         # self.invoke_package_manager(self.repo_clean)  # purges local metadata for rebuild - addresses a possible customer environment error
-        self.invoke_package_manager(self.repo_refresh)
+        for i in range(0, Constants.MAX_REPO_REFRESH_RETRY_COUNT):
+            try:
+                self.invoke_package_manager(self.repo_refresh)
+                return
+            except Exception as error:
+                if i < Constants.MAX_REPO_REFRESH_RETRY_COUNT - 1:
+                    self.composite_logger.log_warning("Exception on package manager refresh repo. [Exception={0}] [RetryCount={1}]".format(repr(error), str(i)))
+                    time.sleep(i + 1)
```
kjohn-msft, line 65: Make the back off longer - (2 ^ i) + 1, given the contention is likely to be longer here

src/core/src/package_managers/ZypperPackageManager.py
kjohn-msft: There's more than this code path calling this. It shouldn't be changed for everyone to accommodate one of the callers.

src/core/src/bootstrap/Constants.py
kjohn-msft: Set this to 5 too

src/core/src/package_managers/ZypperPackageManager.py
kjohn-msft: You can condition it on the proposed check.

src/core/src/core_logic/RebootManager.py
```diff
@@ -91,6 +91,7 @@ def start_reboot_if_required_and_time_available(self, current_time_available):
         """ Starts a reboot if required. Happens only at the end of the run if required. """
         self.composite_logger.log("\nReboot Management")
         reboot_pending = False if not self.status_handler else self.status_handler.is_reboot_pending
+        reboot_pending = self.package_manager.force_reboot or reboot_pending
```
kjohn-msft, line 94: Log again here if the reason why reboot is happening is to help mitigate an issue with the package manager.

kjohn-msft:
Ensure this code is tested on a SUSE machine (doesn't necessarily need to exhibit the issue).

src/core/src/package_managers/ZypperPackageManager.py
benank: I see what you are saying now. Will revert the changes and will look into how I can accommodate it without modifying this part.

src/core/src/core_logic/RebootManager.py
benank: Done.

src/core/src/package_managers/ZypperPackageManager.py
benank: Done

src/core/src/package_managers/ZypperPackageManager.py
benank: Thanks, done.

src/core/src/bootstrap/Constants.py
benank: Done

src/core/src/bootstrap/Constants.py
```diff
@@ -160,6 +160,7 @@ class AutoAssessmentStates(EnumBackport):
     MAX_ASSESSMENT_RETRY_COUNT = 5
     MAX_INSTALLATION_RETRY_COUNT = 3
     MAX_IMDS_CONNECTION_RETRY_COUNT = 5
```
kjohn-msft, line 162: Make the file retry count 10 please. And have it edited in the extension src constants too.

kjohn-msft (changes requested):

src/core/src/bootstrap/Constants.py
benank: Done

kjohn-msft (approved):

PR #93, opened 2021-09-22
Disable auto OS updates in YUM

rane-rajasi:
Checking auto OS updates/patch state in all of these services

- yum-cron
- dnf-automatic
- PackageKit

Overall OSPatchState is computed from individual patch states for each of these services.

There were some updates necessary in the shared code between AptitudePackageManager and YumPackageManager. 
Changes in AptitudePackageManager are tested on Ubuntu 18.04


src/core/src/package_managers/YumPackageManager.py
```diff
@@ -347,26 +386,249 @@ def get_package_size(self, output):
...
+                elif os_auto_update_service == self.dnf_automatic:
+                    self.__init_auto_update_for_dnf_automatic()
+                    self.__get_current_auto_os_updates_setting_on_machine()
+                    if self.enable_on_reboot_value or self.apply_updates_value == 'yes':
+                        break
+                # elif os_auto_update_service == self.packagekit:
+                #     self.__init_auto_update_for_packagekit()
+                #     self.__get_current_auto_os_updates_setting_on_machine()
+                #     if self.enable_on_reboot_value or self.apply_updates_value == 'yes':
+                #         break
```
kjohn-msft, line 409: tagging commented out code

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -347,26 +386,249 @@ def get_package_size(self, output):
...
+            self.composite_logger.log_debug("Auto OS update is NOT installed on the machine")
+            return False
+
+    def install_auto_update_service(self, command):
+        self.composite_logger.log_debug("Installing auto updates using command: " + str(command))
+        code, out = self.env_layer.run_command_output(command, False, False)
+        if code != 0:
+            self.composite_logger.log('[ERROR] Package manager was invoked using: ' + command)
+            self.composite_logger.log_warning(" - Return code from package manager: " + str(code))
+            self.composite_logger.log_warning(" - Output from package manager: \n|\t" + "\n|\t".join(out.splitlines()))
```
kjohn-msft, line 620: Reduce to a single line to reduce log load. This will work because return code is just a number.

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -347,26 +386,249 @@ def get_package_size(self, output):
...
+            self.telemetry_writer.write_execution_error(command, code, out)
+            error_msg = 'Unexpected return code (' + str(code) + ') from package manager on command: ' + command
+            self.status_handler.add_error_to_status(error_msg, Constants.PatchOperationErrorCodes.PACKAGE_MANAGER_FAILURE)
+            raise Exception(error_msg, "[{0}]".format(Constants.ERROR_ADDED_TO_STATUS))
+        else:
+            self.composite_logger.log_debug("\n\n==[SUCCESS]===============================================================")
+            self.composite_logger.log_debug(" - Return code from package manager: " + str(code))
+            self.composite_logger.log_debug(" - Output from package manager: \n|\t" + "\n|\t".join(out.splitlines()))
```
kjohn-msft, line 628: Single line

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -347,26 +386,249 @@ def get_package_size(self, output):
...
+            self.composite_logger.log_debug("==========================================================================\n\n")
```
kjohn-msft, line 629: Things like this should only go to file logger. The composite logger log_verbose function can be used here. And the log_verbose function should not write to telemetry.

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -52,6 +53,44 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
...
+        # commands for DNF Automatic updates service
+        self.dnf_automatic_configuration_file_path = '/etc/dnf/automatic.conf'
+        self.dnf_automatic_install_check_cmd = 'systemctl list-unit-files --type=service | grep dnf-automatic.service'  # list-unit-files returns installed services, ref: https://www.freedesktop.org/software/systemd/man/systemctl.html#Unit%20File%20Commands
+        self.dnf_automatic_install_cmd = 'yum -y install yum-cron'
+        self.dnf_automatic_enable_on_reboot_check_cmd = 'systemctl is-enabled yum-cron'
+        self.dnf_automatic_disable_on_reboot_cmd = 'systemctl disable yum-cron'
+        self.dnf_automatic_config_pattern_match_text = ' = (no|yes)'
+        self.dnf_automatic_download_updates_identifier_text = 'download_updates'
+        self.dnf_automatic_apply_updates_identifier_text = 'apply_updates'
+        self.dnf_automatic_enable_on_reboot_identifier_text = "enable_on_reboot"
```
kjohn-msft: Initialize in a separate function and call from here for code organization. This code and related functions can be in a separate foldable code region for code organization. In the future, we may explore splitting up this class as it's become too long.

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -347,26 +386,249 @@ def get_package_size(self, output):
...
+                #     self.__get_current_auto_os_updates_setting_on_machine()
+                #     if self.enable_on_reboot_value or self.apply_updates_value == 'yes':
+                #         break
+
+        if self.apply_updates_value == 'no' and not self.enable_on_reboot_value:
+            return Constants.AutomaticOSPatchStates.DISABLED
+        elif self.apply_updates_value == 'yes' or self.enable_on_reboot_value:
+            return Constants.AutomaticOSPatchStates.ENABLED
+        else:
+            return Constants.AutomaticOSPatchStates.UNKNOWN
```
kjohn-msft, line 416: Log the values that result in this outcome. Similar for why we picked enabled or disabled. Make the selection, log the selection and then the reasons for that selection, prior to returning.

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -347,26 +386,249 @@ def get_package_size(self, output):
...
+            self.composite_logger.log_debug("Disabling auto OS updates in all identified services...")
+            for os_auto_update_service in self.auto_os_update_service_list:
+                # todo: Better way to implement this? Keeping functions separate since not all services have the same steps to disable and identifying the service specific commands is easier this way
+                if os_auto_update_service == self.yum_cron:
+                    self.disable_auto_os_update_for_yum_cron()
+                elif os_auto_update_service == self.dnf_automatic:
+                    self.disable_auto_os_update_for_dnf_automatic()
+                # elif os_auto_update_service == self.packagekit:
+                #     self.disable_auto_os_update_for_packagekit()
+                else:
```
kjohn-msft, line 446: There's no identification happening here - it's just processing a static list. Each function is called exactly once and the error is thrown once (in this implementation). It's cleaner to just call each function once.

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -347,26 +386,249 @@ def get_package_size(self, output):
...
+                    self.composite_logger.log_error("Disabling auto updates is not implemented for the service. Please check if this is required")
+            self.composite_logger.log_debug("Successfully disabled auto OS updates")
 
-    def is_image_default_patch_configuration_backup_valid(self, image_default_patch_configuration_backup):
-        return True
+        except Exception as error:
+            self.composite_logger.log_error("Could not disable auto OS updates. [Error={0}]".format(repr(error)))
+            raise
```
kjohn-msft, line 452: Please make sure that a failure to configure does not torpedo the execution. This is covered in the configure patching class but sanity-check.

src/core/src/package_managers/YumPackageManager.py
rane-rajasi: Resolved

src/core/src/package_managers/YumPackageManager.py
rane-rajasi: Resolved

src/core/src/package_managers/YumPackageManager.py
rane-rajasi: Resolved

src/core/src/package_managers/YumPackageManager.py
rane-rajasi: Resolved

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -284,24 +284,30 @@ def open(self, file_path, mode):
...
                         time.sleep(i + 1)
                     else:
-                        raise Exception("Unable to open {0} (retries exhausted). Error: {1}.".format(str(real_path), repr(error)))
+                        #todo: do not print this error trace when raise_if_not_found is False. Also do not retry if raise_if_not_found is False
+                        error_message = "Unable to open {0} (retries exhausted). Error: {1}.".format(str(real_path), repr(error))
+                        print(error_message)
+                        if raise_if_not_found:
+                            raise Exception(error_message)
+                        else:
+                            return None
```
kjohn-msft, line 293: Please change error message to the following for diagnosability:
"Unable to open file (retries exhausted). [File={0}][Error={1}][RaiseIfNotFound={2}].".format(str(real_path), repr(error),str(raise_if_not_found))

Sanity check expected behavior in all callers that come through to here.

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -313,7 +319,12 @@ def read_with_retry(self, file_path_or_handle):
                         if i < Constants.MAX_FILE_OPERATION_RETRY_COUNT - 1:
                             time.sleep(i + 1)
                         else:
-                            raise Exception("Unable to read from {0} (retries exhausted). Error: {1}.".format(str(file_path_or_handle), repr(error)))
+                            error_message = "Unable to read from {0} (retries exhausted). Error: {1}.".format(str(file_path_or_handle), repr(error))
```
kjohn-msft, line 322: Same comment about logging format and inclusion of the additional variable.

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -345,28 +371,331 @@ def get_package_size(self, output):
...
+                    if self.enable_on_reboot_value or self.apply_updates_value.lower() == 'true':
+                        break
+        else:
+            self.__get_current_auto_os_updates_setting_on_machine()  # Since auto update service is already identified, this will fetch config for that service
 
-    def disable_auto_os_update(self):
-        """ Disables auto OS updates on the machine only if they are enabled and logs the default settings the machine comes with """
-        # NOTE: Implementation pending
-        pass
+        if (self.apply_updates_value.lower() == 'no' or self.apply_updates_value.lower() == 'false') and not self.enable_on_reboot_value:
```
kjohn-msft, line 437: Assign this to a local variable - it may be evaluated as much as 4 times here without it.

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -345,28 +371,331 @@ def get_package_size(self, output):
...
+                    self.__init_auto_update_for_dnf_automatic()
+                    self.__get_current_auto_os_updates_setting_on_machine()
+                    if self.enable_on_reboot_value or self.apply_updates_value.lower() == 'yes':
+                        break
+                elif os_auto_update_service == self.packagekit:
+                    self.composite_logger.log_debug("Fetching current automatic OS patch state in packagekit service. This includes checks on current state and whether it is set to enable on reboot")
+                    self.__init_auto_update_for_packagekit()
+                    self.__get_current_auto_os_updates_setting_on_machine()
+                    if self.enable_on_reboot_value or self.apply_updates_value.lower() == 'true':
+                        break
```
kjohn-msft, line 433: Still don't get this for loop. Every condition in here is guaranteed to run at least once.

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -345,28 +371,331 @@ def get_package_size(self, output):
...
+                    if match is not None:
+                        self.apply_updates_value = match.group(1)
+
+            if self.download_updates_value == "":
+                self.composite_logger.log_debug("Machine did not have any value set for [Setting={0}]".format(str(self.download_updates_identifier_text)))
+
+            if self.apply_updates_value == "":
+                self.composite_logger.log_debug("Machine did not have any value set for [Setting={0}]".format(str(self.apply_updates_identifier_text)))
+
+            elif self.apply_updates_value == "yes":
```
kjohn-msft, line 618: Why is this check in some places with .lower() and some places like here without it?

It feels like all writes and reads to this property needs to go through an internal read and write function, with the property being fully internal.

src/core/tests/Test_YumPackageManager.py
```diff
@@ -13,7 +13,6 @@
 # limitations under the License.
 #
 # Requires Python 2.7+
-
```
kjohn-msft, line 16: Expected?

src/core/src/package_managers/YumPackageManager.py
kjohn-msft: Refer to a later comment - have a dedicated __get/__check  __set for this variable. And have the variable be internal too. Use the internal convention for all variables that are not expected to be read or handled from outside the class.

kjohn-msft (changes requested):
Thanks for resolving - some other comments inline.

src/core/src/bootstrap/EnvLayer.py
kjohn-msft: @benank - to take in a faster PR, and deploy to Canary for validation. May need additional changes.

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -345,28 +371,331 @@ def get_package_size(self, output):
...
+                self.composite_logger.log_debug("Auto updates are currently enabled")
+            elif self.apply_updates_value == "no":
+                self.composite_logger.log_debug("Auto updates are NOT currently enabled")
+
+        except Exception as error:
+            raise Exception("Error occurred in fetching default auto OS updates from the machine. [Exception={0}]".format(repr(error)))
```
benank, line 624: Might be more clear to have an exception message such as:

Error occurred in fetching **current** auto OS update **settings** from the machine.

src/core/src/package_managers/YumPackageManager.py
rane-rajasi: This part of the code no longer exists.

src/core/src/package_managers/YumPackageManager.py
rane-rajasi: Same as above, this code doesn't exist now

src/core/src/package_managers/YumPackageManager.py
rane-rajasi: As mentioned in other comments, this code doesn't exist now. Added a backlog item to address this comment in other places Task#11007289

src/core/src/bootstrap/EnvLayer.py
rane-rajasi: Resolved

src/core/src/bootstrap/EnvLayer.py
rane-rajasi: Resolved

src/core/src/package_managers/YumPackageManager.py
rane-rajasi: Resolved

src/core/src/package_managers/YumPackageManager.py
rane-rajasi: Resolved

src/core/src/package_managers/YumPackageManager.py
rane-rajasi: Resolved

src/core/tests/Test_YumPackageManager.py
rane-rajasi: Will address while adding UTs

src/core/src/package_managers/YumPackageManager.py
rane-rajasi: Resolved

src/core/src/bootstrap/Constants.py
```diff
@@ -112,6 +112,13 @@ class AutomaticOSPatchStates(EnumBackport):
         DISABLED = "Disabled"
         ENABLED = "Enabled"
 
+    # List of auto OS update services in Yum
+    # todo: Q for reviewer: Should this list be in YumPackageManager, since it is only used there?
```
kjohn-msft, line 116: yes, to prevent constants exploding, and potentially remaining as dead code later if implementations change

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -284,24 +284,29 @@ def open(self, file_path, mode):
                     if i < Constants.MAX_FILE_OPERATION_RETRY_COUNT:
                         time.sleep(i + 1)
                     else:
-                        raise Exception("Unable to open {0} (retries exhausted). Error: {1}.".format(str(real_path), repr(error)))
+                        error_message = "Unable to open file (retries exhausted). [File={0}][Error={1}][RaiseIfNotFound={2}].".format(str(real_path), repr(error), str(raise_if_not_found))
+                        print(error_message)
```
kjohn-msft, line 288: Shouldn't this only be in the 'return None' case?

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -345,28 +370,434 @@ def get_package_size(self, output):
...
+        self.composite_logger.log_debug("Fetching current automatic OS patch state in yum-cron service. This includes checks on whether the service is installed, current auto patch enable state and whether it is set to enable on reboot")
+        self.__init_auto_update_for_yum_cron()
+        is_service_installed, enable_on_reboot_value, download_updates_value, apply_updates_value = self.__get_current_auto_os_updates_setting_on_machine()
+
+        apply_updates = self.__get_extension_standard_value_for_apply_updates(apply_updates_value)
+
+        if apply_updates == self.apply_updates_enabled or enable_on_reboot_value:
+            return Constants.AutomaticOSPatchStates.ENABLED
+        # OS patch state is considered to be disabled: a) if it was successfully disabled or b) if the service is not installed
+        elif not is_service_installed or (apply_updates == self.apply_updates_disabled and not enable_on_reboot_value):
```
benank, line 442: `enable_on_reboot_value` will always be False here

benank: Same with the two other instances below

rane-rajasi: yes that is the check.

service is not installed or updates are disabled and enable on reboot is set to False, only then do we consider patch state to be disabled

src/core/src/bootstrap/EnvLayer.py
rane-rajasi: Resolved

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -345,28 +370,434 @@ def get_package_size(self, output):
...
+            return Constants.AutomaticOSPatchStates.DISABLED
+        else:
+            return Constants.AutomaticOSPatchStates.UNKNOWN
```
benank, line 477: I only see a one line difference between `__get_current_auto_os_patch_state_for_packagekit`, `__get_current_auto_os_patch_state_for_dnf_automatic`, and `__get_current_auto_os_patch_state_for_yum_cron`. Consider putting all the shared code in one place.

src/core/src/bootstrap/Constants.py
rane-rajasi: Will have to look into how to declare constants the packagemanager and refer it, added a note to address this in the next PR

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -345,28 +370,434 @@ def get_package_size(self, output):
...
+        """ Initializes all generic auto OS update variables with the config values for packagekit service """
+        self.os_patch_configuration_settings_file_path = self.packagekit_configuration_file_path
+        self.download_updates_identifier_text = self.packagekit_download_updates_identifier_text
+        self.apply_updates_identifier_text = self.packagekit_apply_updates_identifier_text
+        self.enable_on_reboot_identifier_text = self.packagekit_enable_on_reboot_identifier_text
+        self.installation_state_identifier_text = self.packagekit_installation_state_identifier_text
+        self.auto_update_config_pattern_match_text = self.packagekit_config_pattern_match_text
+        self.enable_on_reboot_check_cmd = self.packagekit_enable_on_reboot_check_cmd
+        self.install_check_cmd = self.packagekit_install_check_cmd
+        self.current_auto_os_update_service = Constants.YumAutoOSUpdateServices.PACKAGEKIT
```
benank, line 521: These 3 init methods could be merged into one that takes the `current_auto_os_update_service` constant as an argument. Instead of having each command/text as a separate variable, they could all be part of a dict using the `current_auto_os_update_service` constant as the first key, and then the name of the text to get as the second key.

eg. `self.auto_update_config[self.current_auto_os_update_service]["configuration_file_path"] ...`

I am not seeing any differences between the different auto update disabling methods to make use of separation like this.

kjohn-msft (approved):

src/core/src/package_managers/YumPackageManager.py
kjohn-msft: @rane-rajasi - please take a to do.

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -345,28 +370,434 @@ def get_package_size(self, output):
...
+    def disable_auto_os_update_for_packagekit(self):
+        """ Disables auto OS updates, using packagekit service, and logs the default settings the machine comes with """
+        self.composite_logger.log("Disabling auto OS updates using packagekit")
+        self.__init_auto_update_for_packagekit()
+
+        self.backup_image_default_patch_configuration_if_not_exists()
+        self.composite_logger.log_debug("Preemptively disabling auto OS updates using packagekit")
+        #todo: uncomment after finding the correct value
+        # self.update_os_patch_configuration_sub_setting(self.download_updates_identifier_text, "false", self.dnf_automatic_config_pattern_match_text)
+        self.update_os_patch_configuration_sub_setting(self.apply_updates_identifier_text, "false", self.dnf_automatic_config_pattern_match_text)
```
benank, line 585: This is using dnf instead of packagekit - intentional?

rane-rajasi: Resolved

src/core/src/package_managers/YumPackageManager.py
rane-rajasi: Will address in next PR

src/core/src/package_managers/YumPackageManager.py
rane-rajasi: Will address in next PR

PR #94, opened 2021-09-23
Revert off by one fix and update version to 1.6.26

benank:
Reverts off by one fix in #92 

New version includes:


kjohn-msft:
Please update manifest and Constant.py.

benank:
> Please update manifest and Constant.py with the versioning update.

Versioning update is here: #95 
But I can move those to this PR if you would like.

kjohn-msft (approved):

PR #95, opened 2021-09-23
Update version to 1.6.26

benank:
Contains:


benank:
Updating version in #94

PR #96, opened 2021-09-28
Add zypper package manager retries

benank:
- Add retries for all invocations to the package manager
- Ensure ZYPP_LOCK_TIMEOUT environment variable is set so it will be used
- Update version to 1.6.28

### DIFF
--- a/src/core/src/bootstrap/EnvLayer.py
+++ b/src/core/src/bootstrap/EnvLayer.py
@@ -83,7 +83,7 @@ class EnvLayer(object):
 
         return ret
 
-    def set_env_var(self, var_name, var_value=None):
+    def set_env_var(self, var_name, var_value=None, raise_if_not_success=False):
         """ Sets an environment variable with var_name and var_value in /etc/environment. If it already exists, it is overwriten. """
         try:
             environment_vars = self.file_system.read_with_retry(self.etc_environment_file_path)
@@ -124,9 +124,10 @@ class EnvLayer(object):
 
         except Exception as error:
             print("Error occurred while setting environment variable [Variable={0}] [Value={1}] [Exception={2}]".format(str(var_name), str(var_value), repr(error)))
-            raise
+            if raise_if_not_success:
+                raise
 
-    def get_env_var(self, var_name):
+    def get_env_var(self, var_name, raise_if_not_success=False):
         """ Returns the value of an environment variable with var_name in /etc/environment. Returns None if it does not exist. """
         try:
             environment_vars = self.file_system.read_with_retry(self.etc_environment_file_path)
@@ -152,7 +153,8 @@ class EnvLayer(object):
 
         except Exception as error:
             print("Error occurred while getting environment variable [Variable={0}] [Exception={1}]".format(str(var_name), repr(error)))
-            raise
+            if raise_if_not_success:
+                raise
 
     def run_command_output(self, cmd, no_output=False, chk_err=False):
         operation = "RUN_CMD_OUT"
--- a/src/core/src/bootstrap/EnvLayer.py
+++ b/src/core/src/bootstrap/EnvLayer.py
@@ -99,7 +99,7 @@ class EnvLayer(object):
                 regex = re.compile('{0}=.+'.format(var_name))
                 search = regex.search(environment_vars)
                 if search is None:
-                    return None
+                    return
 
                 group = search.group()
                 environment_vars = environment_vars.replace(group, '').replace("\n\n", "\n")
--- a/src/core/src/bootstrap/EnvLayer.py
+++ b/src/core/src/bootstrap/EnvLayer.py
@@ -91,9 +91,11 @@ class EnvLayer(object):
                 print("Error occurred while setting environment variable: File not found. [Variable={0}] [Value={1}] [Path={2}]".format(str(var_name), str(var_value), self.etc_environment_file_path))
                 return
 
+            environment_vars_lines = environment_vars.strip().split("\n")
+
             if var_value is None:
                 # remove environment variable
-                if not var_name in environment_vars:
+                if var_name not in environment_vars:
                     return
 
                 regex = re.compile('{0}=.+'.format(var_name))
@@ -107,15 +109,16 @@ class EnvLayer(object):
                 return
 
             formatted_env_var = "{0}={1}".format(var_name, str(var_value))
-            if not var_name in environment_vars:
+            if var_name not in environment_vars:
                 self.file_system.write_with_retry(self.etc_environment_file_path, "\n" + formatted_env_var)
             else:
                 # Update the value of the existing setting
-                for env_var in environment_vars:
+                for env_var in environment_vars_lines:
                     if var_name not in str(env_var):
                         continue
 
-                    environment_vars.replace(str(env_var), formatted_env_var)
+                    environment_vars = environment_vars.replace(str(env_var), formatted_env_var)
+                    break
 
                 self.file_system.write_with_retry(self.etc_environment_file_path, environment_vars, 'w')
 
@@ -133,7 +136,7 @@ class EnvLayer(object):
 
             settings = environment_vars.strip().split('\n')
 
-            if not var_name in settings:
+            if var_name not in environment_vars:
                 return None
             else:
                 # get specific environment variable value
@@ -142,8 +145,8 @@ class EnvLayer(object):
                         continue
 
                     env_var_split = env_var.split("=")
-                    if len(env_var_split) == 2:
-                        return env_var_split[1]
+                    if len(env_var_split) == 2 and env_var_split[0].strip() == var_name:
+                        return env_var_split[1].strip()
 
             return None
 
--- a/src/core/src/bootstrap/EnvLayer.py
+++ b/src/core/src/bootstrap/EnvLayer.py
@@ -19,6 +19,7 @@ import base64
 import datetime
 import json
 import os
+import re
 import platform
 import shutil
 import subprocess
@@ -82,22 +83,35 @@ class EnvLayer(object):
 
         return ret
 
-    def ensure_env_var_is_set(self, var_name, var_value):
-        """ Checks if an environment variable is set with var_name and var_value in /etc/environment. If not, the environment variable is set to var_value. """
+    def set_env_var(self, var_name, var_value=None):
+        """ Sets an environment variable with var_name and var_value in /etc/environment. If it already exists, it is overwriten. """
         try:
-            formatted_env_var = "{0}={1}".format(var_name, var_value)
             environment_vars = self.file_system.read_with_retry(self.etc_environment_file_path)
             if environment_vars is None:
                 print("Error occurred while setting environment variable: File not found. [Variable={0}] [Value={1}] [Path={2}]".format(str(var_name), str(var_value), self.etc_environment_file_path))
                 return
 
-            settings = environment_vars.strip().split('\n')
+            if var_value is None:
+                # remove environment variable
+                if not var_name in environment_vars:
+                    return
 
-            if not var_name in settings:
+                regex = re.compile('{0}=.+'.format(var_name))
+                search = regex.search(environment_vars)
+                if search is None:
+                    return None
+
+                group = search.group()
+                environment_vars = environment_vars.replace(group, '').replace("\n\n", "\n")
+                self.file_system.write_with_retry(self.etc_environment_file_path, environment_vars, "w")
+                return
+
+            formatted_env_var = "{0}={1}".format(var_name, str(var_value))
+            if not var_name in environment_vars:
                 self.file_system.write_with_retry(self.etc_environment_file_path, "\n" + formatted_env_var)
             else:
-                # ensure the setting is the proper value
-                for env_var in settings:
+                # Update the value of the existing setting
+                for env_var in environment_vars:
                     if var_name not in str(env_var):
                         continue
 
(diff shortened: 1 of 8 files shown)


benank:
Still working on tests.

Should I remove the retry code in `refresh_repo` and keep the reboot logic?

src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -44,13 +44,17 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
...
         self.zypper_exitcode_zypper_updated = 103
+        self.zypper_exitcode_zypp_locked = 7
 
         # Support to check for processes requiring restart
         self.zypper_ps = "sudo zypper ps -s"
 
         # Miscellaneous
         self.set_package_manager_setting(Constants.PKG_MGR_SETTING_IDENTITY, Constants.ZYPPER)
         self.zypper_get_process_tree_cmd = 'ps --forest -o pid,cmd -g $(ps -o sid= -p {})'
+        self.env_layer.ensure_env_var_is_set('ZYPP_LOCK_TIMEOUT', 5)
```
kjohn-msft, line 55: Read what the original value is first (and log). Do all of this as you enter invoke_package_manager and reset to original value when you exit from the function. Setting env. var is a low cost operation so the repeats are okay. We shouldn't leave permanent configuration changes on customer machines. There is still a possibility of failure where the value is not reset, but we can address that case later.

benank: Done. Split into get/set and moved to invoke_package_manager

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -55,6 +56,9 @@ def __init__(self, real_record_path=None, recorder_enabled=False, emulator_enabl
         self.file_system = self.FileSystem(recorder_enabled, emulator_enabled, self.__write_record, self.__read_record,
                                            emulator_root_path=os.path.dirname(self.__real_record_path))
 
+        # Constant paths
+        self.etc_environment_file_path = "/etc/environment"
```
kjohn-msft, line 60: We try to be OS agnostic in the environment layer - everything that is not, gets passed in usually. No need to block the PR on this; just FYI.

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -79,6 +83,77 @@ def get_package_manager(self):
...
+                        continue
+
+                    environment_vars = environment_vars.replace(str(env_var), formatted_env_var)
+                    break
+
+                self.file_system.write_with_retry(self.etc_environment_file_path, environment_vars, 'w')
+
+        except Exception as error:
+            print("Error occurred while setting environment variable [Variable={0}] [Value={1}] [Exception={2}]".format(str(var_name), str(var_value), repr(error)))
+            raise
```
kjohn-msft, line 127: This is best effort. 

Do not raise as things can succeed even without this. But because this is a shared class, define the behavior on whether to raise or not through a default variable in the function signature (see examples of 'raise_if_not_found' / similar elsewhere)

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -79,6 +83,77 @@ def get_package_manager(self):
...
+
+                    env_var_split = env_var.split("=")
+                    if len(env_var_split) == 2 and env_var_split[0].strip() == var_name:
+                        return env_var_split[1].strip()
+
+            return None
+
+        except Exception as error:
+            print("Error occurred while getting environment variable [Variable={0}] [Exception={1}]".format(str(var_name), repr(error)))
+            raise
```
kjohn-msft, line 155: Same comment about when to raise.

src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -85,30 +88,56 @@ def refresh_repo(self):
...
-
-        if code == self.zypper_exitcode_zypper_updated:
-            self.composite_logger.log_debug(" - Package manager update detected. Patch installation run will be repeated.")
-            self.set_package_manager_setting(Constants.PACKAGE_MGR_SETTING_REPEAT_PATCH_OPERATION, True)
-        return out
+
+        for i in range(0, self.package_manager_max_retries):
+            zypp_lock_timeout = None
+            try:
+                zypp_lock_timeout = self.env_layer.get_env_var('ZYPP_LOCK_TIMEOUT')
```
kjohn-msft, line 95: zypp_lock_timeout_backup

kjohn-msft (changes requested):
Comments inline

src/core/src/package_managers/ZypperPackageManager.py
benank: Done

src/core/src/bootstrap/EnvLayer.py
benank: Got it, thanks. Fixed and defaulted to not raising.

src/core/src/bootstrap/EnvLayer.py
benank: From what I have seen online, every Linux distro has its global environment variables stored here. But I understand your point and will keep it in mind in the future.

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -79,6 +83,79 @@ def get_package_manager(self):
...
+            if var_value is None:
+                # remove environment variable
+                if var_name not in environment_vars:
+                    return
+
+                regex = re.compile('{0}=.+'.format(var_name))
+                search = regex.search(environment_vars)
+                if search is None:
+                    return
+
```
rane-rajasi, line 105: Do we need these 2 ifs?
First if finds any part in the file with var_name, could be a comment too while the second one looks for the exact pattern match, which is what we are looking for.

If this is right, then I think just the second if should be okay

benank: Yep you're right, the first if would just save the execution of the regex logic. I'll remove the first if and leave it to the pattern matching.

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -79,6 +83,79 @@ def get_package_manager(self):
...
+                # Update the value of the existing setting
+                for env_var in environment_vars_lines:
+                    if var_name not in str(env_var):
+                        continue
+
+                    environment_vars = environment_vars.replace(str(env_var), formatted_env_var)
+                    break
+
+                self.file_system.write_with_retry(self.etc_environment_file_path, environment_vars, 'w')
+
```
rane-rajasi, line 121: Maybe in the next PR, if not enough time here, but the parts for writing when the env_var pre-exists in the the file and when it doesn't could be combined into one.

Refer https://github.com/Azure/LinuxPatchExtension/blob/aaa60bcb59f7a282ac8337a1cf31e9e120bb1e08/src/core/src/package_managers/AptitudePackageManager.py#L483 for a similar example where we write to a config file

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -79,6 +83,76 @@ def get_package_manager(self):
...
+                return None
+            else:
+                # get specific environment variable value
+                for env_var in settings:
+                    if var_name not in str(env_var):
+                        continue
+
+                    env_var_split = env_var.split("=")
+                    if len(env_var_split) == 2 and env_var_split[0].strip() == var_name:
+                        return env_var_split[1].strip()
```
rane-rajasi, line 147: Better to verify if env_var exists with a pattern match, as these files could contain comments with the variable name or the variable name in some other text format. Pattern match guarantees the part we fetch is exactly what we are looking for

src/core/tests/Test_CoreMain.py
```diff
@@ -278,7 +278,7 @@ def test_assessment_operation_fail(self):
         self.assertEquals(len(substatus_file_data), 2)
         self.assertTrue(substatus_file_data[0]["name"] == Constants.PATCH_ASSESSMENT_SUMMARY)
         self.assertTrue(substatus_file_data[0]["status"].lower() == Constants.STATUS_ERROR.lower())
-        self.assertEqual(len(json.loads(substatus_file_data[0]["formattedMessage"]["message"])["errors"]["details"]), 5)
+        self.assertEqual(len(json.loads(substatus_file_data[0]["formattedMessage"]["message"])["errors"]["details"]), 3)
```
rane-rajasi, line 281: Why did this value change?

src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -44,71 +44,96 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
...
+                    self.composite_logger.log_debug("==========================================================================\n\n")
+
+                if code == self.zypper_exitcode_zypper_updated:
+                    self.composite_logger.log_debug(" - Package manager update detected. Patch installation run will be repeated.")
+                    self.set_package_manager_setting(Constants.PACKAGE_MGR_SETTING_REPEAT_PATCH_OPERATION, True)
+                return out
+            except Exception as error:
+                error_msg = repr(error)
+                if Constants.ERROR_ADDED_TO_STATUS in error_msg:
+                    raise
```
rane-rajasi, line 125: Looks like we are depending on Constants.ERROR_ADDED_TO_STATUS to determine if retries are required, which does not look right.

 Constants.ERROR_ADDED_TO_STATUS is used to decide whether or not the error is to be added to error objects and should not be used for other cases. Better to have the usages independent, so that in future if the implementation for error objects change, it won't affect these retries.

src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -44,71 +44,96 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
...
-                    self.status_handler.add_error_to_status(error_msg, Constants.PatchOperationErrorCodes.PACKAGE_MANAGER_FAILURE)
+            # Reboot if not already done
+            if self.status_handler.get_installation_reboot_status() == Constants.RebootStatus.COMPLETED:
+                error_msg = "Unable to refresh repo (retries exhausted after reboot). [{0}]".format(repr(error))
+            else:
+                self.composite_logger.log_warning("Setting force_reboot flag to True.")
+                self.force_reboot = True
+                
+            self.composite_logger.log_warning(error_msg)
+            self.status_handler.add_error_to_status(error_msg, Constants.PatchOperationErrorCodes.PACKAGE_MANAGER_FAILURE)
```
rane-rajasi, line 78: Q: the error coming from command invocation, will be added to error objects from within package manager, either in line 110 or 132 in this file. Why do we need to add another error message?

Some context on error objects:
Error objects refers to the errors json within any summary substatus (i.e. PatchInstallationSummary or PatchAssesmentSummary, etc). Since it's added to a json file with possible size limitations, we only add contents actionable contents within it, which would help the user understand the cause of failure.

For eg: in this case, a message stating which command failed (from line 110 or 132). refresh_repo is a function defined in our code and adding a message "Unable to refresh repo (retries exhausted after reboot)" will not be of much help to the user. 

Add this error message to our logs but don't add it to the error objects, since we already have details on which command failed in there.

To do this, keep Constants.ERROR_ADDED_TO_STATUS in the error_message and simple use composite_logger (any log type you prefer), This will add the message to log files but not error objects in the status

benank: I added this extra error message so that it the source of error could be extra visible - in this case, the error message could say that retries were exhausted after reboot. But I understand now what you mean with the extra context, and I will put the error message from this area into a warning log and keep the Constants.ERROR_ADDED_TO_STATUS in the message.

src/core/src/package_managers/ZypperPackageManager.py
benank: Understood. I wanted some way to share the output code between the try/except blocks and opted for that. I have now just moved the code outside of those blocks so it's accessible to both.

src/core/tests/Test_CoreMain.py
benank: This value changed because of the retry logic. It was originally 2 (before any retries were added), but then set to 5 because of the 5 retries for refresh_repo and how it reported the errors. Now it is back to 2 because there are no more duplicate errors even with the new retries in the invocation.

src/core/src/bootstrap/EnvLayer.py
benank: Thanks, done. Also removed the loop since it can just use the pattern patching.

src/core/src/bootstrap/EnvLayer.py
benank: Got it, thanks.

rane-rajasi: God improvement to the logic but consider having a backlog item to further improve this, for instance we could have a single write_with_retry for 

 if var_name not in environment_vars:
                self.file_system.write_with_retry(self.etc_environment_file_path, "\n" + formatted_env_var)
            else:
                # Update the value of the existing setting

src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -44,71 +44,86 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
...
-                        self.composite_logger.log_warning("Setting force_reboot flag to True.")
-                        self.force_reboot = True
-                        
-                    self.composite_logger.log_warning(error_msg)
-                    self.status_handler.add_error_to_status(error_msg, Constants.PatchOperationErrorCodes.PACKAGE_MANAGER_FAILURE)
-
-                    raise Exception(error_msg)
+        try:
+            self.invoke_package_manager(self.repo_refresh)
+            return
```
rane-rajasi, line 63: minor: don't need return here, function should return if there is no exception and invoke_package_manager completes

rane-rajasi (approved):

src/core/src/package_managers/ZypperPackageManager.py
benank: Right, fixed

PR #97, opened 2021-09-29
Create updatecenter_troubleshooter.py

nikhim-um:
Added these checks:
Python check
Supported OS Check
Internet Connectivity check
IMDS connectivity check
Azcmagent installed and connected
Azcmagent  check if all services are active (systemctl list-units extd.service)
Check if extensions are available on the machine
Check of auto assessment service running or not
Check access to linux repositories

TODO:
azure agent connectivity, version checks pending


src/diagnostics/updatecenter_troubleshooter.py
```diff
@@ -0,0 +1,1337 @@
+#!/usr/bin/env python
```
kjohn-msft, line 1: Add copyright at the top

src/tools/diagnostics/updatecenter_troubleshooter.py
```diff
@@ -0,0 +1,1337 @@
+#!/usr/bin/env python
```
kjohn-msft, line 1: Consider README.md in the diagnostics folder explaining to say a customer / CSS how to use it. Overall:
- how to launch it
- how to capture the output

kjohn-msft (approved):

PR #100, opened 2021-10-19
Disabling auto updates in SUSE

rane-rajasi:


src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -15,12 +15,18 @@
...
+import json
+import os
 import re
 import time
 from core.src.package_managers.PackageManager import PackageManager
 from core.src.bootstrap.Constants import Constants
 
 
+class ZypperAutoOSUpdateServices(Constants.EnumBackport):
+    YAST2_ONLINE_UPDATE_CONFIGURATION = "yast2-online-update-configuration"
```
kjohn-msft, line 27: Can be nested into ZypperPackageManager and initialized there

src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -451,28 +467,203 @@ def get_package_size(self, output):
     # endregion
 
     # region auto OS updates
+    def __init_constants_for_yast2_online_update_configuration(self):
+        self.yast2_online_update_configuration_os_patch_configuration_settings_file_path = '/etc/sysconfig/automatic_online_update'
+        self.yast2_online_update_configuration_apply_updates_identifier_text = 'AOU_ENABLE_CRONJOB'
+        self.yast2_online_update_configuration_auto_update_config_pattern_match_text = '="(true|false)"'
+        self.yast2_online_update_configuration_installation_state_identifier_text = "installation_state"
```
kjohn-msft, line 474: Variable naming suggests need to enacapsulate into an object

src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -451,28 +467,203 @@ def get_package_size(self, output):
...
         """ Gets the current auto OS update patch state on the machine """
-        # NOTE: Implementation pending
-        pass
+        self.composite_logger.log("Fetching the current automatic OS patch state on the machine...")
+
+        current_auto_os_patch_state_for_yast2_online_update_configuration = self.__get_current_auto_os_patch_state_for_yast2_online_update_configuration()
+        self.composite_logger.log("OS patch state per auto OS update service: [yast2-online-update-configuration={0}]".format(str(current_auto_os_patch_state_for_yast2_online_update_configuration)))
+
+        current_auto_os_patch_state = current_auto_os_patch_state_for_yast2_online_update_configuration
+        self.composite_logger.log_debug("Overall Auto OS Patch State based on all auto OS update service states [OverallAutoOSPatchState={0}]".format(str(current_auto_os_patch_state)))
```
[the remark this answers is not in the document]
rane-rajasi: This is similar across Yum and Zypper, which have/may have multiple auto update services. 

Aptitude has this structure
image

benank, line 484: These two logging lines will log the same patch state variable - are both needed?

rane-rajasi: Right now it will log the same patch state since we have only one update service. This is written in a way to accommodate multiple update services. So the last log will give overall patch state based on different individual patch states

kjohn-msft:
Code organization and non-functional comments can be taken in a different PR.

kjohn-msft (approved):

src/core/src/package_managers/ZypperPackageManager.py
rane-rajasi: Resolved

PR #102, opened 2021-11-16
[Draft] Adding retries to zypper search

benank:


src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -362,7 +362,7 @@ def get_all_available_versions_of_package_ex(self, package_name, include_install
         package_versions = []
 
         self.composite_logger.log_debug("\nGetting all available versions of package '" + package_name + "' [Installed=" + str(include_installed) + ", Available=" + str(include_available) + "]...")
-        cmd = self.single_package_check_versions.replace('<PACKAGE-NAME>', package_name)
+        cmd = self.single_package_check_versions.replace('<PACKAGE-NAME>', package_name) 
         output = self.invoke_package_manager(cmd)
```
benank, line 366: These lines would be enclosed in retry logic with a maximum amount of retries and would only retry if the return code is 7.

kjohn-msft:
What is this PR about?

PR #103, opened 2021-11-19
Do not add error to status if it already exists

benank:


src/extension/tests/Test_ExtOutputStatusHandler.py
```diff
@@ -123,10 +123,35 @@ def test_add_error_to_status(self):
         self.assertEqual(updated_status_json[0][self.status_file_fields.status][self.status_file_fields.status_substatus][0][self.status_file_fields.status_name], Constants.PATCH_NOOPERATION_SUMMARY)
         self.assertNotEqual(json.loads(updated_status_json[0]["status"]["substatus"][0]["formattedMessage"]["message"])["errors"], None)
         self.assertEqual(json.loads(updated_status_json[0]["status"]["substatus"][0]["formattedMessage"]["message"])["errors"]["code"], 1)
-        self.assertTrue(len(json.loads(updated_status_json[0]["status"]["substatus"][0]["formattedMessage"]["message"])["errors"]["details"]), 5)
+        self.assertEqual(len(json.loads(updated_status_json[0]["status"]["substatus"][0]["formattedMessage"]["message"])["errors"]["details"]), 5)
```
benank, line 126: FYI: assertTrue here always succeeds, so I changed it to assertEqual

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -646,28 +646,28 @@ def add_error_to_status(self, message, error_code=Constants.PatchOperationErrorC
...
 
         if current_operation == Constants.ASSESSMENT:
-            self.__add_error(self.__assessment_errors, error_detail)
-            self.__assessment_total_error_count += 1
-            # retain previously set status and code for assessment substatus
-            if self.__assessment_substatus_json is not None:
-                self.set_assessment_substatus_json(status=self.__assessment_substatus_json["status"], code=self.__assessment_substatus_json["code"])
-            else:
-                self.set_assessment_substatus_json()
+            if self.__add_error(self.__assessment_errors, error_detail):
```
kjohn-msft, line 649: Rename the __add_error method to __try_add_error to make it immediately obvious that there are circumstances where the message add request won't be honored.

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -691,11 +691,16 @@ def __ensure_error_message_restriction_compliance(full_message):
     @staticmethod
     def __add_error(add_to, detail):
         """ Add formatted error object to given errors list """
+        for error_detail in add_to:
+            if detail["message"] == error_detail["message"]:
+                return False
+
```
kjohn-msft, line 697: Consider the case where this message [A] is already in the list, and message [B] comes along. Can you remove [A] from the list and replace it with the more "expressive" message [B] too?

[A] -- "I/O exception copying file"
[B] -- "Execution stopped due to failure. Exception: I/O exception copying file."

Having A in there while B gets reported next is sort of redundant.

This will mean you will also additionally need to to see if any entry in the list is also contained within the new message that is coming in.

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -691,11 +691,16 @@ def __ensure_error_message_restriction_compliance(full_message):
     @staticmethod
     def __add_error(add_to, detail):
```
kjohn-msft, line 692: __try_add_error as noted before

src/core/src/service_interfaces/StatusHandler.py
kjohn-msft: Also, note, when this is done, the reverse chronological order of message arrival should be preserved. Newest message is first in the list.

kjohn-msft (changes requested):
Look good. Requested additions for 'reverse encapsulation' of messages - where new messages are more verbose and expressive and encapsulates an older message.

src/core/src/service_interfaces/StatusHandler.py
benank: Good idea. Added logic to do this.

One thing that we might want to consider is the error codes of the messages - should they also be required to match when determining whether to add a message or not? I made the assumption that if the error message is similar enough (contains it), then the error is probably on the same severity level.

src/core/src/service_interfaces/StatusHandler.py
benank: Thanks, done

src/core/src/service_interfaces/StatusHandler.py
kjohn-msft: The updated implementation doesn't have the newest arriving message as first in the list if there is a message in can replace lower in the list. (Messages are reverse chronological.. )

B' => [X, A, B, C]   -- becomes -- > [B', X, A, C],  not [X, A, B', C]

Another approach, which is probably better is to do it only for the current first element. Less verbose messages may remain if they are older than the stored newest.

benank: Adjusted code to preserve reverse chronological ordering of error messages.

kjohn-msft (approved):

PR #104, opened 2021-12-01
Add additional logging when events folder path is not found

benank:


src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -41,6 +41,10 @@ def __init__(self, env_layer, composite_logger, events_folder_path):
         if events_folder_path is not None and os.path.exists(events_folder_path):
             self.events_folder_path = events_folder_path
             self.__is_agent_compatible = True
+        else:
+            self.write_event('VM has incompatible agent. Events folder path: {} (Exists: {})'
+                .format(str(events_folder_path), str(os.path.exists(events_folder_path)) if events_folder_path is not None else 'False'),
+                Constants.TelemetryEventLevel.Informational)
```
kjohn-msft, line 47: Unit test class initialization with non-existent path and review event

benank: There are UTs in place for this case already and I stepped through to ensure that the event is written correctly

kjohn-msft: Need to take a step back. What's the value of writing this event? If the events folder is not there, telemetry will not work. And then writing a telemetry event doesn't do anything. The write_event function will do nothing at all - it just immediately returns.

I'm not sure how any unit test showed this event as being written unless there was a flaw in the test setup somehow. Which test showed this working? Need to look at that closer.

You may want to write to the local log file using composite logger but it'll just be in the machine log, and that won't tell us anything the log doesn't already contain. Reviewing all of the above, is the change necessary?

kjohn-msft:
Keep investigating differences between core and handler to explain behavior.

PR #109, opened 2022-01-26
Guest agent telemetry improvements

benank:
Previously, the only method used to check whether or not telemetry was supported by the guest agent on a VM was to check for the existence of the "events" folder, used by the guest agent for telemetry.

This PR changes the telemetry check across the extension and core by changing the primary requirement to the following:
1. Get the value of the environment variable: `AZURE_GUEST_AGENT_EXTENSION_SUPPORTED_FEATURES`
2. Value should look something like this: `'[{  "Key": "ExtensionTelemetryPipeline", "Value": "1.0"}]'`
3. Check for the existence of the `ExtensionTelemetryPipeline` key within the list of key-value pairs of supported features. If this key exists, then the currently installed version of the guest agent supports telemetry for the extension. 

This approach can be seen in both the [extension](https://github.com/Azure/LinuxPatchExtension/blob/3ff25a3eb815bc851d3e722ab1de5df1ff67ab5b/src/extension/src/ActionHandler.py#L118) and [core](https://github.com/Azure/LinuxPatchExtension/blob/3ff25a3eb815bc851d3e722ab1de5df1ff67ab5b/src/core/src/service_interfaces/TelemetryWriter.py#L41).

There are a few more elements to consider with this approach:
 - If the events folder is not given then execution cannot continue, even if the guest agent is installed and supports telemetry. In this case, [additional version information is logged](https://github.com/Azure/LinuxPatchExtension/blob/3ff25a3eb815bc851d3e722ab1de5df1ff67ab5b/src/extension/src/ActionHandler.py#L134) about the guest agent to aid in investigation.
 - If telemetry is supported and the events folder is given but does not exist, then the events folder is created and execution continues as normal.

Additional logging about the installed agent version has also been added within core in the constructor for TelemetryWriter.

Verbose logs (in both ext and core) are no longer sent to telemetry and are only logged to file to reduce the amount of data sent through telemetry. Some locations with verbose log information have been changed to use `log_verbose` now.

### DIFF
--- a/src/core/src/service_interfaces/TelemetryWriter.py
+++ b/src/core/src/service_interfaces/TelemetryWriter.py
@@ -44,6 +44,7 @@ class TelemetryWriter(object):
 
         self.write_event('Started Linux patch core operation.', Constants.TelemetryEventLevel.Informational)
         self.write_machine_config_info()
+        self.log_agent_information()
 
     def write_config_info(self, config_info, config_type='unknown'):
         # Configuration info
--- a/src/core/src/service_interfaces/TelemetryWriter.py
+++ b/src/core/src/service_interfaces/TelemetryWriter.py
@@ -123,17 +123,21 @@ class TelemetryWriter(object):
         """ Returns True if the events folder path passed in is not None and exists on disk """
         return events_folder_path is not None and os.path.exists(events_folder_path)
 
-    @staticmethod
-    def __get_agent_supports_telemetry_from_env_var():
+    def __get_agent_supports_telemetry_from_env_var(self):
         """ Returns True if the env var AZURE_GUEST_AGENT_EXTENSION_SUPPORTED_FEATURES has a key of
             ExtensionTelemetryPipeline in the list. Value of the env var looks like this:
             '[{  "Key": "ExtensionTelemetryPipeline", "Value": "1.0"}]' """
         features_keyvalue_list_str = os.getenv(Constants.AZURE_GUEST_AGENT_EXTENSION_SUPPORTED_FEATURES_ENV_VAR)
         if features_keyvalue_list_str is None:
+            self.composite_logger.log_error('Failed to get guest agent supported features from env var. [Var=\'{}\']'.format(Constants.AZURE_GUEST_AGENT_EXTENSION_SUPPORTED_FEATURES_ENV_VAR))
             return False
 
         features_keyvalue_list = json.loads(features_keyvalue_list_str)
-        return any(kv_pair for kv_pair in features_keyvalue_list if kv_pair['Key'] == Constants.TELEMETRY_EXTENSION_PIPELINE_SUPPORTED_KEY)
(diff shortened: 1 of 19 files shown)


src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -117,6 +118,76 @@ def get_disk_type(self):
...
+        return version_search.group()
+
+    def log_agent_information(self):
+        """ Logs WALinuxAgent version information. """
+        agent_version = self.get_agent_version()
+        if agent_version is None:
+            self.composite_logger.log('WALinuxAgent is not installed.', Constants.TelemetryEventLevel.Informational)
+        else:
+            self.composite_logger.log('WALinuxAgent version: {}\nGoal state agent version: {}'.format(agent_version, self.get_goal_state_agent_version()), Constants.TelemetryEventLevel.Informational)
+
```
kjohn-msft, line 190: Code needs instrumentation. Is it possible to explain how the code executed and why it executed that way looking at a log file alone? 

For example, did the function at line 127 return False because the environment variable didn't exist or because the variable existed but the key didn't. On a related tangent, environment variables are case sensitive so take care around that.

Again for agent version search, did function fail because the agent returned nothing, or because it returned something that we weren't expecting to parse. Assume BVTs are all good and it's some customer coming and saying that our code didn't run and here's the log.

We must always be able to explain what code paths executed and why it executed that way from logs alone, else we'll be in this situation where we reactively need to push new code out and have a customer run the code again to understand what is actually happening. Which is a lot of overhead and will cause customer dissat.

kjohn-msft (changes requested):
Comment inline

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -145,8 +149,10 @@ def get_agent_version(self):
                 WALinuxAgent-2.2.49.2 running on sles 15.3
                 Python: 3.6.13
                 Goal state agent: 2.6.0.2 '''
-            return self.__agent_version_string_search(r'WALinuxAgent-\S+ running', out)
+            return self.__extract_agent_version_from_string(r'WALinuxAgent-\S+ running', out)
 
+        # Command failed, so log error and debugging information
+        self.composite_logger.log_error('Failed to execute command to get guest agent version. [Code={}] [Out=\'{}\']'.format(str(code), str(out)))
```
kjohn-msft, line 155: While non-indexed placeholders work, the pattern we follow in the code base is to index them for readability. Please adjust all the pre-format strings to align.

kjohn-msft (commented):
Minor comment inline

src/core/src/service_interfaces/TelemetryWriter.py
benank: Thanks. Added some more logging for clarity so that the code path can be traced later.

kjohn-msft (approved):

src/core/src/service_interfaces/TelemetryWriter.py
benank: Thanks, fixed!

PR #110, opened 2022-01-26
Add package manager return code 102 to zypper success codes

benank:
Adds return code 102 as a success code for zypper package manager.

> 102 - ZYPPER_EXIT_INF_REBOOT_NEEDED
Returned after a successful installation of a patch which requires reboot of computer.

Execution still continues normally, but the `force_reboot` flag on the package manager will be set to true so the machine will reboot when available.

Previously, this would be added as an error to the installation summary.

### DIFF
--- a/src/core/src/core_logic/RebootManager.py
+++ b/src/core/src/core_logic/RebootManager.py
@@ -94,7 +94,7 @@ class RebootManager(object):
         reboot_pending = self.package_manager.force_reboot or reboot_pending
 
         if self.package_manager.force_reboot:
-            self.composite_logger.log("Reboot is occurring to mitigate an issue with the package manager.")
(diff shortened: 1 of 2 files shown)


kjohn-msft:
Please additionally change this log in RebootManager on line 97 from:
self.composite_logger.log("Reboot is occurring to mitigate an issue with the package manager.")
to:
self.composite_logger.log("A reboot is pending as the package manager required it.")

kjohn-msft (changes requested):
Comment left on overall PR to change an additional file.

benank:
> Please additionally change this log in RebootManager on line 97 from: self.composite_logger.log("Reboot is occurring to mitigate an issue with the package manager.") to: self.composite_logger.log("A reboot is pending as the package manager required it.")

Updated.

kjohn-msft (approved):

PR #111, opened 2022-01-28
Fix SUSE error codes 6 and 106

benank:
Zypper package manager exit code 6 is `ZYPPER_EXIT_NO_REPOS`. This means that no repositories are defined. A refresh services command might be able to fix it: `sudo zypper refresh --services`. 

After failing with exit code 6, the services are refreshed, and the command is tried again. 

Exit code 106 is `ZYPPER_EXIT_INF_REPOS_SKIPPED`, which means that a repository failed to refresh for some reason and was skipped/disabled. This exit code has been added to the list of codes to retry since this might be a transient error.

Also includes:
 - Reorganization of zypper exit codes to be in ascending order for better readability
 - Moved `raise` in `refresh_repo` to inside the block that runs if it is still failing after a reboot. Previously, the reboot would never complete since an error was raised before trying to reboot

### DIFF
--- a/src/core/src/package_managers/ZypperPackageManager.py
+++ b/src/core/src/package_managers/ZypperPackageManager.py
@@ -107,11 +107,10 @@ class ZypperPackageManager(PackageManager):
             # Reboot if not already done
             if self.status_handler.get_installation_reboot_status() == Constants.RebootStatus.COMPLETED:
                 self.composite_logger.log_warning("Unable to refresh repo services (retries exhausted after reboot).")
+                raise
             else:
                 self.composite_logger.log_warning("Setting force_reboot flag to True after refreshing repo services.")
                 self.force_reboot = True
-                
-            raise
 
     # region Get Available Updates
     def invoke_package_manager(self, command):
--- a/src/core/src/package_managers/ZypperPackageManager.py
+++ b/src/core/src/package_managers/ZypperPackageManager.py
@@ -40,6 +40,7 @@ class ZypperPackageManager(PackageManager):
         # Repo refresh
         self.repo_clean = 'sudo zypper clean -a'
         self.repo_refresh = 'sudo zypper refresh'
+        self.repo_refresh_services = 'sudo zypper refresh --services'
 
         # Support to get updates and their dependencies
         self.zypper_check = 'sudo LANG=en_US.UTF8 zypper list-updates'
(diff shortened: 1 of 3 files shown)


src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -67,7 +70,7 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
         self.set_package_manager_setting(Constants.PKG_MGR_SETTING_IDENTITY, Constants.ZYPPER)
         self.zypper_get_process_tree_cmd = 'ps --forest -o pid,cmd -g $(ps -o sid= -p {})'
         self.package_manager_max_retries = 5
-        self.error_codes_to_retry = [self.zypper_exitcode_zypp_locked, self.zypper_exitcode_zypp_lib_exit_err]
+        self.error_codes_to_retry = [self.zypper_exitcode_zypp_locked, self.zypper_exitcode_zypp_lib_exit_err, self.zypper_exitcode_repos_skipped]
```
kjohn-msft, line 73: Wouldn't this be better location after line 64? And probably renamed to self.zypper_retriable_exit_codes?

src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -95,6 +98,21 @@ def refresh_repo(self):
...
+            self.invoke_package_manager(self.repo_refresh_services)
+        except Exception as error:
+            # Reboot if not already done
+            if self.status_handler.get_installation_reboot_status() == Constants.RebootStatus.COMPLETED:
+                self.composite_logger.log_warning("Unable to refresh repo services (retries exhausted after reboot).")
+            else:
+                self.composite_logger.log_warning("Setting force_reboot flag to True after refreshing repo services.")
+                self.force_reboot = True
+                
+            raise
```
kjohn-msft, line 114: Have you validated that the exception path out of here ensures that a reboot actually happens? (vs. terminating the execution completely with a failure)

kjohn-msft (changes requested):
Review comments

src/core/src/package_managers/ZypperPackageManager.py
benank: Yep, agreed!

src/core/src/package_managers/ZypperPackageManager.py
benank: It looks like it is terminating the execution completely with a failure. I'll look into making sure that it actually reboots.

benank: Adjusted so that the error is only thrown if the error persists after a reboot; otherwise, throwing the error before a reboot will prevent the machine from rebooting.

kjohn-msft (approved):

PR #114, opened 2022-03-03
Telemetry not supported incident mitigation

benank:
Mitigation + additional logging for telemetry not supported incident.

Guest Agent version not supported error message has been updated to include support error codes:

Example:
> Unsupported older Azure Linux Agent version. To resolve: http://aka.ms/UpdateLinuxAgent [AV:2.2.49.2, AGSV:2.6.0.2, EFE:0, EV:1, EVC: 0]

Code information:
- `AV`: Agent Version (Guest Agent)
  - `-1` if version not found, otherwise will be the version string
- `AGSV`: Agent Goal State Version (Guest Agent)
  - `-1` if version not found, otherwise will be the version string
- `EFE`: Events Folder Exists (on disk)
  - `0` if the events folder does not exist, `1` if it does exist
- `EV`: Env Var exists
  - `0` if the env var does not exist, `1` if it does exist
- `EVC`: Env Var Code (more detailed information on what went wrong when getting the env var)
  - See `__get_agent_supports_telemetry_from_env_var` for more information

### DIFF
--- a/src/core/src/service_interfaces/TelemetryWriter.py
+++ b/src/core/src/service_interfaces/TelemetryWriter.py
@@ -38,7 +38,7 @@ class TelemetryWriter(object):
         self.start_time_for_event_count_throttle_check = datetime.datetime.utcnow()
         self.event_count = 1
 
-        if self.__get_agent_supports_telemetry_from_env_var() and self.__get_events_folder_path_exists(events_folder_path):
+        if self.__get_events_folder_path_exists(events_folder_path):
             self.events_folder_path = events_folder_path
             self.__is_agent_compatible = True
 
--- a/src/core/src/service_interfaces/TelemetryWriter.py
+++ b/src/core/src/service_interfaces/TelemetryWriter.py
@@ -193,7 +193,7 @@ class TelemetryWriter(object):
 
         return version_search.group()
 
-    def get_telemetry_information(self):
+    def get_telemetry_diagnostics(self):
         """ Returns information about the guest agent telemetry for debugging purposes.
             Information message abbreviations:
                 AV: Agent Version (Guest Agent)
@@ -207,13 +207,13 @@ class TelemetryWriter(object):
         agent_goalstate_version = self.get_goal_state_agent_version()
         events_folder_exists = self.__get_events_folder_path_exists(self.events_folder_path)
         telemetry_env_var_supported = self.__get_agent_supports_telemetry_from_env_var()
-        return "AV: {0} AGSV: {1} EFE: {2} EV: {3} EVC: {4}".format(
-            str(agent_version) if agent_version is not None else "-1",
-            str(agent_goalstate_version) if agent_goalstate_version is not None else "-1",
-            "1" if events_folder_exists is True else "0",
-            "1" if telemetry_env_var_supported is True else "0",
-            str(self.agent_env_var_code)
-        )
+        return "AV:{0}, AGSV:{1}, EFE:{2}, EV:{3}, EVC:{4}".format(
+                str(agent_version) if agent_version is not None else "-1",
+                str(agent_goalstate_version) if agent_goalstate_version is not None else "-1",
+                "1" if events_folder_exists is True else "0",
+                "1" if telemetry_env_var_supported is True else "0",
+                str(self.agent_env_var_code)
+            )
 
     def __log_agent_information(self):
         """ Logs WALinuxAgent version information. """
--- a/src/core/src/service_interfaces/TelemetryWriter.py
+++ b/src/core/src/service_interfaces/TelemetryWriter.py
@@ -37,6 +37,7 @@ class TelemetryWriter(object):
         self.__telemetry_event_counter = 1  # will be added at the end of each event sent to telemetry to assist in tracing and identifying event/message loss in telemetry
         self.start_time_for_event_count_throttle_check = datetime.datetime.utcnow()
         self.event_count = 1
+        self.agent_env_var_code = 0  # code to give details on what went wrong when getting env var
 
         if self.__get_events_folder_path_exists(events_folder_path):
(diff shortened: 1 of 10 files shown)


src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -189,6 +193,28 @@ def __extract_agent_version_from_string(self, pattern, string):
...
+                EFE: Events Folder Exists (on disk)
+                EV: Env Var exists
+                EVC: Env Var Code (more detailed information on what went wrong when getting the env var)
+                    See __get_agent_supports_telemetry_from_env_var for more information
+        """
+        agent_version = self.get_agent_version()
+        agent_goalstate_version = self.get_goal_state_agent_version()
+        events_folder_exists = self.__get_events_folder_path_exists(self.events_folder_path)
+        telemetry_env_var_supported = self.__get_agent_supports_telemetry_from_env_var()
+        return "AV: {0} AGSV: {1} EFE: {2} EV: {3} EVC: {4}".format(
```
kjohn-msft, line 210: Minor: AV:{0}, AGSV:{1}, EFE:{2}, EV:{3}, EVC:{4}
for better delimiting

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -189,6 +193,28 @@ def __extract_agent_version_from_string(self, pattern, string):
 
         return version_search.group()
 
+    def get_telemetry_information(self):
```
kjohn-msft, line 196: get_telemetry_diagnostics

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -189,6 +193,28 @@ def __extract_agent_version_from_string(self, pattern, string):
...
+        agent_goalstate_version = self.get_goal_state_agent_version()
+        events_folder_exists = self.__get_events_folder_path_exists(self.events_folder_path)
+        telemetry_env_var_supported = self.__get_agent_supports_telemetry_from_env_var()
+        return "AV: {0} AGSV: {1} EFE: {2} EV: {3} EVC: {4}".format(
+            str(agent_version) if agent_version is not None else "-1",
+            str(agent_goalstate_version) if agent_goalstate_version is not None else "-1",
+            "1" if events_folder_exists is True else "0",
+            "1" if telemetry_env_var_supported is True else "0",
+            str(self.agent_env_var_code)
+        )
```
kjohn-msft, line 216: Check general indentation - the final ) is not supposed to be the same level as return on 210 for readability. 211 through 216 would benefit from one more level of nesting.

src/extension/src/ActionHandler.py
```diff
@@ -112,30 +112,43 @@ def tear_down(self):
...
+        if events_folder is not None:
             # Guest agent fully supports telemetry
-            ''' NOTE: unlike core, this code will run even if events_folder does not exist, 
-                since telemetry_writer.is_agent_compatible() only checks the env var. 
-                This ensures that the events_folder exists once core runs. '''
+            self.__log_telemetry_info(telemetry_supported=True)
+
             if not os.path.exists(events_folder):
                 os.mkdir(events_folder)
                 self.logger.log("Events folder path found in HandlerEnvironment but does not exist on disk. Creating now. [Path={0}]".format(str(events_folder)))
```
kjohn-msft, line 124: Please log the agent version where this was observed here too so this type of bug in the agent is easily measurable at scale from a single match (vs. going over multiple per machine).

src/extension/src/ActionHandler.py
```diff
@@ -112,30 +112,43 @@ def tear_down(self):
...
+        events_folder_str = str(events_folder) if events_folder is not None else ""
+        events_folder_exists = os.path.exists(events_folder) if events_folder is not None else False
+        env_var_supports_telemetry = self.telemetry_writer.is_agent_compatible()
+        telemetry_info = "Events folder: \'{0}\' - Events folder exists: {1} - Env var: {2}".format(
+            events_folder_str, str(events_folder_exists), env_var_supports_telemetry)
+
+        if env_var_supports_telemetry is True:
+            telemetry_info += " AgentVer: {0} GoalStateVer: {1}".format(self.telemetry_writer.get_agent_version(), self.telemetry_writer.get_goal_state_agent_version())
+        else:
+            telemetry_info += " AgentVer: Unknown GoalStateVer: Unknown"
```
kjohn-msft, line 145: The general structure we follow for logs is "Message. [Key1=Value1][Key2=Value2]...[Keyn=Valuen]" where key follows variable naming convention (so no spaces). Comment applies to 139, 143, 145.

kjohn-msft: And indirectly to 148, 150 because of the final composition outcome.

kjohn-msft (changes requested):
Comments inline.

src/extension/src/ActionHandler.py
benank: Agent version is included in `__log_telemetry_info` - do you mean a separate log with just the agent version?

src/extension/src/ActionHandler.py
benank: What about for the diagnostic codes (EV, AV, etc)? Should those be formatted similarly as well?

src/extension/src/ActionHandler.py
kjohn-msft: At the end of this log line -- "Creating now. [Path={0}][AgentVersion={1}]"

src/extension/src/ActionHandler.py
kjohn-msft: That's okay for status since it's customer facing.

kjohn-msft (approved):

PR #118, opened 2022-03-11
Fix RETRY - 1 issue

benank:
In some cases where retries are applied to an operation, the ending condition (used to log an error or throw an exception) is never reached because python ranges are exclusive of the ending bound. This PR fixes this issue and adds a -1 to the conditions so that the previously unreachable code is now used.

### DIFF
--- a/src/core/tests/Test_LifecycleManagerAzure.py
+++ b/src/core/tests/Test_LifecycleManagerAzure.py
@@ -51,7 +51,7 @@ class TestLifecycleManagerAzure(unittest.TestCase):
         # file open throws exception
         self.lifecycle_manager.ext_state_file_path = old_ext_state_file_path
         self.runtime.env_layer.file_system.open = self.mock_file_open_throw_exception
-        ext_state_json = self.lifecycle_manager.read_extension_sequence()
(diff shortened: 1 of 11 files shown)


src/core/tests/Test_LifecycleManagerAzure.py
```diff
@@ -61,9 +61,21 @@ def test_read_extension_sequence_success(self):
...
         # file open throws exception
+        backup_open = self.runtime.env_layer.file_system.open
         self.runtime.env_layer.file_system.open = self.mock_file_open_throw_exception
-        core_sequence_json = self.lifecycle_manager.read_core_sequence()
+        core_sequence_json = self.assertRaises(Exception, self.lifecycle_manager.read_core_sequence)
+        self.assertEquals(core_sequence_json, None)
+        # Unable to write to core state file (retries exhausted)
+
+        # json throws exception, but in a different area to test retries exhausted
+        import json
```
kjohn-msft, line 71: Please import at the top of the file

kjohn-msft:
Please ensure all unit tests are run in both core and ext, and pass as expected.

kjohn-msft (approved):

PR #120, opened 2022-03-11
Code coverage improvements

benank:
- Fixes code coverage comments in PRs (see below)
 - Blocks merging PRs if coverage goes down (any code added _must_ have UTs)
 - Adds Python 2.7 code coverage (unsure of how this is working right now - will require more time later to investigate)
 - Does not block merging if tests are failing - will need to investigate more on that later

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,15 +1,36 @@
 name: Workflow for Codecov
-on: [pull_request]
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
 jobs:
-  run:
+  codecov-python-3.9:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -1,15 +1,36 @@
 name: Workflow for Codecov
-on: [pull_request]
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
 jobs:
-  run:
+  codecov-python-39:
```
kjohn-msft, line 8: Why is 39 mapped to 2.7?

kjohn-msft (approved):

PR #121, opened 2022-03-14
Report 'CompletedWithWarnings' instead of 'Failed' if customer picked 'NeverReboots' and reboot is required

benank:
Previous behavior was to report 'Completed' in this scenario. Now it reports 'CompletedWithWarnings'.

### DIFF
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -197,6 +197,7 @@ class Constants(object):
         REQUIRED = "Required"
         STARTED = "Started"
         COMPLETED = "Completed"
+        COMPLETED_WITH_WARNINGS = "CompletedWithWarnings"
         FAILED = "Failed"
     
     # Enum for VM Cloud Type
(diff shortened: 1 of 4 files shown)


src/core/src/bootstrap/Constants.py
```diff
@@ -197,6 +197,7 @@ class RebootStatus(EnumBackport):
         REQUIRED = "Required"
         STARTED = "Started"
         COMPLETED = "Completed"
+        COMPLETED_WITH_WARNINGS = "CompletedWithWarnings"
```
kjohn-msft, line 200: The ask is on patch operation status, not on reboot status. Hold off on additional iterations until discussion is cleared.

kjohn-msft, line 200: The ask is on patch operation status, not on reboot status. Hold off on additional iterations until discussion is cleared.

kjohn-msft:
This PR can be abandoned. There was more context to the work item that would have helped deprioritize this sooner.

PR #122, opened 2022-03-14
Fix Zypper Package Manager Return Code 8

benank:
Fixes Zypper package manager return code 8. This error occurs when a package fails to install due to a conflict with another package.

### DIFF
--- a/src/core/src/package_managers/ZypperPackageManager.py
+++ b/src/core/src/package_managers/ZypperPackageManager.py
@@ -58,6 +58,7 @@ class ZypperPackageManager(PackageManager):
         self.zypper_exitcode_zypp_lib_exit_err = 4
         self.zypper_exitcode_no_repos = 6
         self.zypper_exitcode_zypp_locked = 7
+        self.zypper_exitcode_zypp_exit_err_commit = 8
         self.zypper_exitcode_reboot_required = 102
         self.zypper_exitcode_zypper_updated = 103
         self.zypper_exitcode_repos_skipped = 106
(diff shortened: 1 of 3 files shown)


src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -156,6 +166,17 @@ def invoke_package_manager(self, command):
                 self.force_reboot = True
             return out
 
+    def modify_upgrade_or_patch_cmd_to_replacefiles(self, cmd):
```
kjohn-msft, line 169: Minor: use the same variable names for the same entity even if it doesn't matter in a programmatic call stack: command instead of cmd here.

src/core/tests/library/LegacyEnvLayerExtensions.py
```diff
@@ -579,6 +603,9 @@ def run_command_output(self, cmd, no_output=False, chk_err=True):
                     if cmd.find('sudo zypper refresh') > -1:
                         code = 6
                         output = 'Warning: There are no enabled repositories defined. | Use \'zypper addrepo\' or \'zypper modifyrepo\' commands to add or enable repositories.'
+                    elif cmd.find('sudo zypper --non-interactive update') > -1:
```
kjohn-msft, line 606: This can be more specific to only catch the specific command instance where you want to return 8. (Use unique package details and incorporate here)

benank: Fixed + added a check to ensure that code 8 is returned in the UT.

PR #124, opened 2022-03-16
Add extra checks for Python path in startup script

benank:
Adds extra checks for Python in case it isn't in the path/environment variables.

Also changes exit code 51 to exit code 52, according to the [Extensions Best Practices](https://github.com/Azure/azure-marketplace/wiki/Extension-Build-Notes-Best-Practices#error-codes-and-messages-output-to-stderr) page. (missing dependency instead of unsupported OS)

### DIFF
--- a/src/extension/src/MsftLinuxPatchExtShim.sh
+++ b/src/extension/src/MsftLinuxPatchExtShim.sh
@@ -107,6 +107,7 @@ find_python PYTHON
 
 
 if [ -z "$PYTHON" ]; then
+   # Error codes: https://github.com/Azure/azure-marketplace/wiki/Extension-Build-Notes-Best-Practices#error-codes-and-messages-output-to-stderr
    echo "Python is required, but no Python interpreter was found on the box" >&2
    exit 52
 else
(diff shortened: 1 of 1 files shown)


src/extension/src/MsftLinuxPatchExtShim.sh
```diff
@@ -109,8 +107,8 @@ find_python PYTHON
 
 
 if [ -z "$PYTHON" ]; then
-   echo "No Python interpreter found on the box" >&2
-   exit 51 # Not Supported
+   echo "Python is required, but no Python interpreter was found on the box" >&2
+   exit 52
```
kjohn-msft, line 111: What's the error code mapping for 52?

kjohn-msft (changes requested):
Minor comment ask.

src/extension/src/MsftLinuxPatchExtShim.sh
benank: [Error code mapping for 52 is missing dependency](https://github.com/Azure/azure-marketplace/wiki/Extension-Build-Notes-Best-Practices#error-codes-and-messages-output-to-stderr); error code 51 is unsupported OS.

kjohn-msft (approved):

PR #126, opened 2022-03-16
Add Python 2.7 code coverage

benank:
Adds flags to separate Python 3.9 and 2.7 code coverage.


.github/workflows/ci.yml
```diff
@@ -38,4 +42,8 @@ jobs:
           coverage run -m unittest discover -s src
           coverage xml
       - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v2
\ No newline at end of file
+        uses: codecov/codecov-action@v2
+        with:
+          flags: python27
+          name: python-27
+          fail_ci_if_error: true
```
kjohn-msft, line 49: Not having new line at the end of the file getting flagged. Please include here and anywhere else.

Part of the reason for this is to avoid an unchanged line at the end of the file looking like it changed when more lines are added - incidentally like in this PR on the old line 41.

kjohn-msft (approved):

PR #136, opened 2022-04-11
Do not error if OS backup settings do not exist

benank:
Any new machine that installs the extension will encounter this error:
```
Extension does not have a valid backup of the default system configuration settings for auto OS updates.
```

This PR fixes this by checking if the backup file exists before checking if it is valid, which will prevent this error from occurring on new machines.

### DIFF
--- a/src/core/src/package_managers/AptitudePackageManager.py
+++ b/src/core/src/package_managers/AptitudePackageManager.py
@@ -477,7 +477,7 @@ class AptitudePackageManager(PackageManager):
             self.composite_logger.log_debug("Extension already has a valid backup of the default system configuration settings for auto OS updates.")
             return True
         else:
-            self.composite_logger.log_error("Extension does not have a valid backup of the default system configuration settings for auto OS updates.")
+            self.composite_logger.log_debug("Extension does not have a valid backup of the default system configuration settings for auto OS updates.")
             return False
 
     def update_os_patch_configuration_sub_setting(self, patch_configuration_sub_setting, value="0", patch_configuration_sub_setting_pattern_to_match=""):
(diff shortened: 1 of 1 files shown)


src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -477,7 +477,7 @@ def is_image_default_patch_configuration_backup_valid(self, image_default_patch_
             self.composite_logger.log_debug("Extension already has a valid backup of the default system configuration settings for auto OS updates.")
             return True
         else:
-            self.composite_logger.log_error("Extension does not have a valid backup of the default system configuration settings for auto OS updates.")
+            self.composite_logger.log_debug("Extension does not have a valid backup of the default system configuration settings for auto OS updates.")
```
kjohn-msft, line 480: I don't think there is a bug here. The problem is likely in the flow of the function "backup_image_default_patch_configuration_if_not_exists"

- It initializes an empty json
- It then proceeds to check if that empty json is a valid backup and the code logically reads like it's expecting a valid backup. Which is all wrong as its misleading and unnecessary, because of course the "backup" is not valid if it's just the initialized default. 

Consider fixing that because when this error is written (and it can legitimately be an error), it is because there is an actual problem with the data read, and not because it's a default value in code that's being inspected.

kjohn-msft (commented):
Some comments inline.

src/core/src/package_managers/AptitudePackageManager.py
benank: Thanks for the insight. I see the problem now and have made a few changes to fix this.

kjohn-msft (approved):

PR #137, opened 2022-04-14
Write to multiple statuses for ConfigurePatching on error

benank:
If ConfigurePatching fails with an error, the error is now also written to `autoAssessmentState` and `errors` sections in `autoAssessmentStatus`.


src/core/src/core_logic/ConfigurePatchingProcessor.py
```diff
@@ -140,7 +142,6 @@ def __raise_if_agent_incompatible(self):
             return
         if not self.telemetry_writer.is_agent_compatible():
             error_msg = "{0} [{1}]".format(Constants.TELEMETRY_AT_AGENT_NOT_COMPATIBLE_ERROR_MSG, self.telemetry_writer.get_telemetry_diagnostics())
-            self.composite_logger.log_error(error_msg)
```
benank, line 143: Line removed because `__report_consolidated_configure_patch_status` logs the error, so this results in duplicate messages (at least, without the duplicate checking code that exists).

kjohn-msft (approved):

PR #139, opened 2022-04-26
Reliability and monitoring improvements

kjohn-msft:
PR is in draft. Description will be updated.

### DIFF
--- a/src/core/src/core_logic/PatchAssessor.py
+++ b/src/core/src/core_logic/PatchAssessor.py
@@ -15,9 +15,12 @@
 # Requires Python 2.7+
 
 """ A patch assessment """
-import time
-from core.src.bootstrap.Constants import Constants
+import json
 import os
+import shutil
+import time
+import datetime
+from core.src.bootstrap.Constants import Constants
 
 class PatchAssessor(object):
     """ Wrapper class of a single patch assessment """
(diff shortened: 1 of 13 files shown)


src/core/src/core_logic/PatchAssessor.py
```diff
@@ -94,94 +103,109 @@ def raise_if_agent_incompatible(self):
...
-                    self.composite_logger.log_error("Unable to write to core state file (retries exhausted). [Exception={0}]".format(repr(error)))
+                    self.composite_logger.log_error("Unable to write to assessment state file (retries exhausted). [Exception={0}]".format(repr(error)))
                     raise
 
-        self.composite_logger.log_debug("Completed updating core sequence.")
-
+        self.composite_logger.log_debug("Completed updating assessment state.")
+
+    def get_seconds_since_epoch(self):
+        return int((datetime.datetime.now() - datetime.datetime(1970, 1, 1)).total_seconds())
```
Klaas-, line 182: I don't think that works like you think it works, my guess would be that datetime(1970, 1, 1) is in your local TZ? For example for me that resolves to -3600 epoch :)
But what you are trying to do is redundant anyways, epoch is literally the time in seconds since 1.1.1970, so you can just convert now into a timestamp. datetime.datetime.now().timestamp()

kjohn-msft:
Code was taken in piece-meal when I was away on parental leave. This PR will be abandoned after sanity-checking with master, after it is fully refreshed post all outstanding PRs.

PR #140, opened 2022-05-05
Add telemetrySupported field to environmentSettings

benank:
There are instances where inherited environment variables from the extension handler are not available in the extension core that should be. Specifically, anything passed in from the guest agent to the extension handler is not available in core in these rare cases. This may be due to the extension handler process exiting before it is able to pass all of its environment variables to the core process, but exact root cause is unknown.

This PR abstracts all references to agent in core by providing one environment setting "telemetrySupported" that is passed in by the handler if the proper environment variable exists. This should solve the problem by not relying on the flaky inherited environment variables and instead relying on the arguments passed into the core process by the handler.

It also re-adds some of the telemetry logic back to ensure that both the events folder is specified _and_ the environment variable exists from #109


kjohn-msft (commented):
As discussed, continue to maintain levels of abstraction between handler and core. So have the data go through environment settings and only just the required amount of information required by core.

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -31,21 +31,22 @@ class TelemetryWriter(object):
...
         self.event_count = 1
-        self.agent_env_var_code = Constants.AgentEnvVarStatusCode.AGENT_ENABLED  # code to give details on what went wrong when getting env var
 
         if self.__get_events_folder_path_exists(events_folder_path):
             self.events_folder_path = events_folder_path
-            self.__is_agent_compatible = True
 
-        self.write_event('Started Linux patch core operation.', Constants.TelemetryEventLevel.Informational)
-        self.write_machine_config_info()
-        self.__log_agent_information()
```
rane-rajasi, line 48: As discussed, keep this in. telemetrySupported can be added to TelemetryWriter such that it will be set during init
get_path_to_log_files_and_telemetry_dir() in Bootstrapper should be able to extract 'telemetrySupported' from environmentSettings which can be sent back to Bootstrapper, then used in ConfigurationFactory() init to be available as a init var for TelemetryWriter.

This ensures we have Telemetry from as early as possible and and perform telemetry exists check

src/core/tests/Test_ConfigurePatchingProcessor.py
```diff
@@ -129,12 +129,12 @@ def test_operation_success_for_installation_request_with_configure_patching(self
...
         runtime.stop()
 
-    def test_operation_fail_for_configure_patching_agent_incompatible(self):
+    def test_operation_fail_for_configure_patching_telemetry_not_supported(self):
         argument_composer = ArgumentComposer()
         argument_composer.operation = Constants.CONFIGURE_PATCHING
         argument_composer.patch_mode = Constants.PatchModes.AUTOMATIC_BY_PLATFORM
         argument_composer.events_folder = None
-        runtime = RuntimeCompositor(argument_composer.get_composed_arguments(), True, Constants.APT)
+        runtime = RuntimeCompositor(argument_composer.get_composed_arguments(env_settings=dict(telemetrySupported=False)), True, Constants.APT)
```
rane-rajasi, line 137: env_settings=dict(telemetrySupported=False) this will only set telemetrySupported in env_settings and no value for the rest (i.e. config_folder, status_folder, etc)

src/core/tests/library/RuntimeCompositor.py
```diff
@@ -101,8 +104,6 @@ def __init__(self, argv=Constants.DEFAULT_UNSPECIFIED_VALUE, legacy_mode=False,
         self.configure_patching_processor.auto_assess_service_manager.remove_service = self.mock_remove_service
         self.backup_remove_timer = self.configure_patching_processor.auto_assess_timer_manager.remove_timer
         self.configure_patching_processor.auto_assess_timer_manager.remove_timer = self.mock_remove_timer
-        self.backup_os_getenv = os.getenv
-        os.getenv = self.getenv_telemetry_enabled
```
rane-rajasi, line 107: ensure to remove restting os.getenv from self.backup_os_getenv also, if done

src/extension/tests/helpers/RuntimeComposer.py
```diff
@@ -13,7 +13,6 @@
 class RuntimeComposer(object):
     def __init__(self):
         self.backup_os_getenv = os.getenv
-        os.getenv = self.getenv_telemetry_enabled
```
rane-rajasi, line 16: remove self.backup_os_getenv = os.getenv also, if not used and the line which resets os.getenv from backup

src/extension/src/__main__.py
```diff
@@ -49,6 +49,7 @@ def main(argv):
         runtime_context_handler = RuntimeContextHandler(logger)
         json_file_handler = JsonFileHandler(logger)
         ext_env_handler = ExtEnvHandler(json_file_handler)
+        ext_env_handler.telemetry_supported = telemetry_writer.is_agent_compatible()
```
rane-rajasi, line 52: Add a test where telemetry is not supported because env var is not available from agent

src/core/tests/library/RuntimeCompositor.py
benank: Done - core tests don't reset os.getenv

src/extension/tests/helpers/RuntimeComposer.py
benank: This line was needed - added back

src/core/src/service_interfaces/TelemetryWriter.py
benank: Done!

src/core/src/core_logic/PatchInstaller.py
```diff
@@ -98,16 +98,16 @@ def start_installation(self, simulate=False):
 
         return overall_patch_installation_successful
 
-    def raise_if_agent_incompatible(self):
+    def raise_if_telemetry_unsupported(self):
         if self.lifecycle_manager.get_vm_cloud_type() == Constants.VMCloudType.ARC and self.execution_config.operation not in [Constants.ASSESSMENT, Constants.INSTALLATION]:
-            self.composite_logger.log("Skipping agent compatibility check for Arc cloud type when operation is not manual")
+            self.composite_logger.log("Skipping telemetry compatibility check for Arc cloud type when operation is not manual")
```
benank, line 103: It's impossible to hit these lines because assessment always runs before installation and also has this check, so it will fail the entire operation before getting to the installer. I added one test to mock a no-op assessment so that these lines can be covered by tests, but I had a question: How can this line be hit if the operation isn't installation? CoreMain specifically sets the operation to installation before starting the patch installer, so this should (?) always fail.

src/core/src/bootstrap/Constants.py
```diff
@@ -271,8 +270,3 @@ class EnvLayer(EnumBackport):
         PRIVILEGED_OP_REBOOT = PRIVILEGED_OP_MARKER + "Reboot_Exception"
         PRIVILEGED_OP_EXIT = PRIVILEGED_OP_MARKER + "Exit_"
 
-    # These are numbers instead of strings like extension because they are customer facing in the error message diagnostics
-    class AgentEnvVarStatusCode(EnumBackport):
```
rane-rajasi, line 275: nit: Ensure references to any of these codes in not left over in code

rane-rajasi, line 275: nit: Ensure references to any of these codes in not left over in code

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -80,7 +80,7 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
         self.__current_operation = None
 
         # If an error message is any of these strings, it ignores the length limit (STATUS_ERROR_MSG_SIZE_LIMIT_IN_CHARACTERS)
-        self.__ignore_error_message_restriction_compliance_strings = [Constants.TELEMETRY_AT_AGENT_NOT_COMPATIBLE_ERROR_MSG]
+        self.__ignore_error_message_restriction_compliance_strings = [Constants.TELEMETRY_NOT_COMPATIBLE_ERROR_MSG]
```
rane-rajasi, line 83: Why ignore the message restrictions? Afaik, these restriction are what the agent has on a status file, so we should follow those for any of our file writes

rane-rajasi, line 83: Why ignore the message restrictions? Afaik, these restriction are what the agent has on a status file, so we should follow those for any of our file writes

src/core/tests/Test_PatchInstaller.py
```diff
@@ -189,6 +189,56 @@ def healthstore_writes_helper(self, health_store_id, maintenance_run_id, expecte
...
+
+        runtime = RuntimeCompositor(argument_composer.get_composed_arguments(env_settings={"telemetrySupported": False}), True, Constants.YUM)
+        runtime.set_legacy_test_type('SuccessInstallPath')
+        self.assertRaises(Exception, runtime.patch_installer.raise_if_telemetry_unsupported)
+        runtime.stop()
+
+        runtime = RuntimeCompositor(argument_composer.get_composed_arguments(env_settings={"telemetrySupported": True}), True, Constants.YUM)
+        runtime.set_legacy_test_type('SuccessInstallPath')
+        runtime.patch_installer.lifecycle_manager.get_vm_cloud_type = lambda: Constants.VMCloudType.ARC
+        runtime.patch_installer.raise_if_telemetry_unsupported()
```
rane-rajasi, line 213: Is there any assert to check here? The function returns a value or sets something

rane-rajasi, line 213: Is there any assert to check here? The function returns a value or sets something

src/core/src/bootstrap/Constants.py
benank: Done, none left in core

benank: Duplicate comment? Or was this intended elsewhere

src/core/src/service_interfaces/StatusHandler.py
benank: This was originally added as an exception to ensure that the full error message with the link to resolve the agent issue was included in the message, along with the newly added diagnostic codes. I don't think this was a hard limit (as it's been working fine with this exception), but it was added as an exception to add on a constant-size extra string.

We can remove it now.

benank: Duplicate comment

src/core/src/bootstrap/Constants.py
benank: Duplicate comment

src/core/tests/Test_PatchInstaller.py
benank: It logs something, but the check here works because it will throw an exception if it fails, and this will fail the overall test. There isn't a way to check for something like `self.assertNotRaises` because an exception will fail the test anyway.

src/extension/tests/Test_ActionHandler.py
```diff
@@ -252,6 +253,9 @@ def test_update_command_fail(self):
         # Remove the directory after the test
         shutil.rmtree(test_dir)
 
+    def test_telemetry_available(self):
+        self.assertTrue(self.runtime.telemetry_writer.is_agent_compatible())
+
     def test_telemetry_available_env_var_not_exists(self):
         # agent env var is not set so telemetry is not supported
```
rane-rajasi, line 260: This test looks like is one for telemetry not supported scenario, but the assert on line 277 uses telemetry supported check

self.assertTrue(Constants.TELEMETRY_AT_AGENT_COMPATIBLE_MSG in file_contents)

rane-rajasi: Same comment for the test test_telemetry_available_env_var_key_not_exists

src/extension/src/__main__.py
rane-rajasi: Comments on the these tests left in TestActionHander

src/core/src/service_interfaces/StatusHandler.py
rane-rajasi: Did this exception exist just for the telemetry not supported message (without diagnostic codes) before? I vaguely remember a discussion about that

benank: This exception was added specifically for the diagnostic code additon. Previously, the telemetry not supported message fit into the character limit.

src/extension/tests/Test_ActionHandler.py
benank: You are right that this is not correct. I think that it was added like this because _previously_ we had the strict check (env var and events folder) and it was failing, so those changes were reverted and these tests were changed so that it checks for telemetry existing even though the env var is not there (but events folder is). 

I can change the code to rely on both events folder and env var (strict check).

src/core/tests/Test_PatchInstaller.py
```diff
@@ -189,6 +189,56 @@ def healthstore_writes_helper(self, health_store_id, maintenance_run_id, expecte
...
+        runtime.patch_installer.lifecycle_manager.get_vm_cloud_type = lambda: Constants.VMCloudType.ARC
+        runtime.patch_installer.execution_config.operation = Constants.CONFIGURE_PATCHING
+        runtime.patch_installer.raise_if_telemetry_unsupported()
+        runtime.stop()
+
+        runtime = RuntimeCompositor(argument_composer.get_composed_arguments(env_settings={"telemetrySupported": False}), True, Constants.YUM)
+        runtime.set_legacy_test_type('SuccessInstallPath')
+        runtime.patch_installer.lifecycle_manager.get_vm_cloud_type = lambda: Constants.VMCloudType.ARC
+        runtime.patch_installer.execution_config.operation = Constants.CONFIGURE_PATCHING
+        runtime.patch_installer.raise_if_telemetry_unsupported()
```
rane-rajasi, line 233: Shouldn't this raise an exception, since telemetry is not supported?

benank: Added clarification comment

rane-rajasi (approved):

PR #145, opened 2022-06-09
Only handle Zypper repeat exit code in one place

benank:
Addendum to #135. 

The fix for #135 did not work because there was one more location that was setting the package manager run to be repeated even on dry runs. This PR fixes it by removing that location and only having one location that handles any reboot/repeat exit codes (and now a message in the output as well)

### DIFF
--- a/src/core/src/package_managers/ZypperPackageManager.py
+++ b/src/core/src/package_managers/ZypperPackageManager.py
@@ -176,7 +176,7 @@ class ZypperPackageManager(PackageManager):
 
         if code == self.zypper_exitcode_zypper_updated or self.zypper_out_zypper_updated_msg in out:
             self.composite_logger.log_debug(
-                "One of the installed patches affects the package manager itself. Patch installation run will be repeated.")
+                " - One of the installed patches affects the package manager itself. Patch installation run will be repeated.")
             self.set_package_manager_setting(Constants.PACKAGE_MGR_SETTING_REPEAT_PATCH_OPERATION, True)
         elif code == self.zypper_exitcode_reboot_required:
             self.composite_logger.log_warning(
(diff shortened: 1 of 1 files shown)


src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -172,7 +175,7 @@ def __handle_zypper_updated_or_reboot_exit_codes(self, command, code):
                 str(code), command))
             return
 
-        if code == self.zypper_exitcode_zypper_updated:
+        if code == self.zypper_exitcode_zypper_updated or self.zypper_out_zypper_updated_msg in out:
             self.composite_logger.log_debug(
                 " - Package manager update detected. Patch installation run will be repeated.")
```
rane-rajasi, line 180: Update log to One of the installed patches affects the package manager. Patch installation run will be repeated

src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -172,7 +175,7 @@ def __handle_zypper_updated_or_reboot_exit_codes(self, command, code):
                 str(code), command))
```
rane-rajasi, line 175: Update logs on line 174/175. Remove the exit code reference

src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -159,11 +162,11 @@ def invoke_package_manager(self, command):
                 self.log_success_on_invoke(code, out)
 
             if code == self.zypper_exitcode_zypper_updated or code == self.zypper_exitcode_reboot_required:
```
rane-rajasi, line 164: Remove if statement. __handle_zypper_updated_or_reboot_exit_codes() will be called always irrespective of error code

rane-rajasi (approved):

PR #146, opened 2022-06-10
Differentiate logs from auto assessment and auto patching

benank:
We don't have any markers to differentiate between the extension telemetry events coming from an auto assessment operation and any of the non auto assessment operations. With auto assessment running in regular intervals, we should consider adding some marker to differentiate between these logs for easier investigation. This PR adds some markers to differentiate between logs.

### DIFF
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -126,6 +126,11 @@ class Constants(object):
         DISABLED = "Disabled"
         ENABLED = "Enabled"
 
+    # To separately preserve assessment + auto-assessment state information
+    ASSESSMENT_STATE_FILE = "AssessmentState.json"
+    AUTO_ASSESSMENT_MAXIMUM_DURATION = "PT1H"
+    MIN_AUTO_ASSESSMENT_INTERVAL = "PT6H"   # do not perform auto-assessment if the last assessment happened less than this time interval ago
+
     # wait time after status updates
     WAIT_TIME_AFTER_HEALTHSTORE_STATUS_UPDATE_IN_SECS = 20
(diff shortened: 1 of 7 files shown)


src/extension/src/TelemetryWriter.py
```diff
@@ -96,6 +98,7 @@ def write_event(self, message, event_level=Constants.TelemetryEventLevel.Informa
 
             self.__delete_older_events()
 
+            task_name = self.__task_name if task_name == Constants.TELEMETRY_TASK_NAME else task_name
```
rane-rajasi, line 101: Shouldn't this be:
task_name = self.__task_name if task_name == Constants.TELEMETRY_TASK_NAME else task_name **+ self.__task_name_watermark**

src/core/src/bootstrap/Constants.py
```diff
@@ -126,6 +126,11 @@ class AutoAssessmentStates(EnumBackport):
         DISABLED = "Disabled"
         ENABLED = "Enabled"
 
+    # To separately preserve assessment + auto-assessment state information
+    ASSESSMENT_STATE_FILE = "AssessmentState.json"
+    AUTO_ASSESSMENT_MAXIMUM_DURATION = "PT1H"
+    MIN_AUTO_ASSESSMENT_INTERVAL = "PT6H"   # do not perform auto-assessment if the last assessment happened less than this time interval ago
```
rane-rajasi, line 132: Are these used in the code anywhere?

benank: Not yet. Will be used in a future PR. Removed for now.

rane-rajasi (approved):

PR #152, opened 2022-06-24
Bugfix for error while updating top level message field in status file

rane-rajasi:
Existing code was failing to write top level message in status file with the error reported as
image

This was not a blocker or terminal error but affected our reporting status accurately in status blob

### DIFF
--- a/src/extension/src/file_handlers/ExtOutputStatusHandler.py
+++ b/src/extension/src/file_handlers/ExtOutputStatusHandler.py
@@ -141,7 +141,7 @@ class ExtOutputStatusHandler(object):
             self.update_key_value_safely(status_json[0], self.file_keys.status_status, status, self.file_keys.status_status)
             self.update_key_value_safely(status_json[0], self.file_keys.status_code, code, self.file_keys.status_status)
             self.update_key_value_safely(status_json[0], self.file_keys.timestamp_utc, str(datetime.datetime.utcnow().strftime(Constants.UTC_DATETIME_FORMAT)))
-            self.update_key_value_safely(status_json[0][self.file_keys.status_status], self.file_keys.status_formatted_message_message, str(message), self.file_keys.status_formatted_message)
+            self.update_key_value_safely(status_json[0][self.file_keys.status], self.file_keys.status_formatted_message_message, str(message), self.file_keys.status_formatted_message)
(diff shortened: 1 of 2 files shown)


src/extension/src/file_handlers/ExtOutputStatusHandler.py
```diff
@@ -138,10 +138,10 @@ def update_file(self, seq_no, status=Constants.Status.Transitioning.lower(), cod
...
                 return
-            self.update_key_value_safely(status_json, self.file_keys.status_status, status, self.file_keys.status_status)
-            self.update_key_value_safely(status_json, self.file_keys.status_code, code, self.file_keys.status_status)
-            self.update_key_value_safely(status_json, self.file_keys.timestamp_utc, str(datetime.datetime.utcnow().strftime(Constants.UTC_DATETIME_FORMAT)))
-            self.update_key_value_safely(status_json, self.file_keys.status_formatted_message_message, str(message), self.file_keys.status_formatted_message)
+            self.update_key_value_safely(status_json[0], self.file_keys.status_status, status, self.file_keys.status_status)
+            self.update_key_value_safely(status_json[0], self.file_keys.status_code, code, self.file_keys.status_status)
+            self.update_key_value_safely(status_json[0], self.file_keys.timestamp_utc, str(datetime.datetime.utcnow().strftime(Constants.UTC_DATETIME_FORMAT)))
+            self.update_key_value_safely(status_json[0][self.file_keys.status_status], self.file_keys.status_formatted_message_message, str(message), self.file_keys.status_formatted_message)
```
benank, line 144: Why is the `status_status` key used here?

mirichmo, line 144: should there be a comma between status_json[0] and self.file_keys?

mirichmo (changes requested):

src/extension/src/file_handlers/ExtOutputStatusHandler.py
rane-rajasi: Resolved.

rane-rajasi: No because it's to access the key inside status_json[0]

benank (approved):

PR #154, opened 2022-07-14
Fix for ConfigurePatching getting stuck in transitioning

benank:
Fix for ConfigurePatching getting stuck in transitioning. ConfigurePatching would run, enable auto-assessment, and auto assessment would start but load the status file from disk. After ConfigurePatching completes, auto assessment would run but would not have updated status file information since it did not read it from disk again after ConfigurePatching updated if when it finished its operation.

### DIFF
--- a/src/core/src/service_interfaces/LifecycleManagerAzure.py
+++ b/src/core/src/service_interfaces/LifecycleManagerAzure.py
@@ -96,6 +96,8 @@ class LifecycleManagerAzure(LifecycleManager):
 
             # Signalling take-over of core state by auto-assessment after safety checks for any competing process
             self.update_core_sequence(completed=False)
+            # Refresh status file in memory to be up-to-date
+            self.status_handler.load_status_file_components()
         else:
             # Logic for all non-Auto-assessment operations
             extension_sequence = self.read_extension_sequence()
--- a/src/core/src/service_interfaces/LifecycleManagerArc.py
+++ b/src/core/src/service_interfaces/LifecycleManagerArc.py
@@ -108,6 +108,8 @@ class LifecycleManagerArc(LifecycleManager):
 
             # Signalling take-over of core state by auto-assessment after safety checks for any competing process
             self.update_core_sequence(completed=False)
+            # Refresh status file in memory to be up-to-date
+            self.status_handler.load_status_file_components()
         else:
             # Logic for all non-Auto-assessment operations
             extension_sequence = self.read_extension_sequence()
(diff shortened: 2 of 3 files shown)


src/core/src/service_interfaces/LifecycleManagerAzure.py
```diff
@@ -96,6 +96,8 @@ def execution_start_check(self):
 
             # Signalling take-over of core state by auto-assessment after safety checks for any competing process
             self.update_core_sequence(completed=False)
+            # Refresh status file in memory to be up-to-date
+            self.status_handler.load_status_file_components()
```
mirichmo, line 100: 1. Should this be done on line 120 so it is done for both cases?
2. Should it be done in all code paths within update_core_sequence? Update_core_sequence has an early return.
3. What about all invocations of update_core_sequence? Should it be done there as well?

benank: 1. The second case should already have up-to-date status file information because there isn't any delay there. It isn't needed on line 120 but can be moved there.
2. Core sequence is read at the beginning of each loop within the `execution_start_check` method, so it should always have up-to-date information.
3. I don't think this is needed. It's only needed in areas of contention when there are multiple core processes running at the same time, and this is already taken care of (as mentioned above).

mirichmo (approved):

src/core/src/service_interfaces/LifecycleManagerAzure.py
rane-rajasi, line 100: 1. Test this change on a machine with Auto Patching and Auto Assessment
2. We don't have any tests mocking Auto assessment behavior, to test this scenario. See if it's possible to add that

benank: 1. Tested on a machine with those attributes, works as intended.
2. Looking into it.

rane-rajasi: When testing on a machine, check/follow these:
1. Create a VM with Auto Patching and Auto Assessment i.e. patchmode = AutomaticByPlatform and assessmentmode = AutomaticByPlatform
2. See if a ConfigurePatching call is triggered. If not, manually invoke ConfigurePatching API
3. Keep on monitoring the API result. Query ContextActivity and check ConfigurePatchingSummary, PatchAssessmentSummary substatus.
4. If both have succeeded, then the changes are valid. If not, the issue still exists

rane-rajasi: Add a task in our backlog for AutoAssessment tests

rane-rajasi (approved):

PR #155, opened 2022-07-18
Fix logging for dpkg was interrupted (Exit code 100)

benank:
This PR is a fix to improve the logging for the most common type of exit code 100 (there are multiple types of errors for this exit code) - `dpkg was interrupted, you must manually run 'sudo dpkg --configure -a' to correct the problem.`. 

While there is mitigation code for this error in `invoke_package_manager` already, it was not hit by this error's code path because the command was invoked directly from `install_update_and_dependencies`. Now it uses `invoke_package_manager` and the errors + mitigation steps will be available to the customer.


src/core/tests/Test_YumPackageManager.py
```diff
@@ -418,7 +418,7 @@ def test_ssl_certificate_issue_type1_fix_success(self):
         package_manager = self.container.get('package_manager')
         self.assertTrue(package_manager)
 
-        output = package_manager.invoke_package_manager(package_manager.yum_check)
+        output, code = package_manager.invoke_package_manager(package_manager.yum_check)
```
rane-rajasi, line 421: Do we have/need to make the same change, in Test_ZypperPackageManager.py?

benank: Yes, but Test_ZypperPackageManager.py doesn't use the output of `invoke_package_manager` anywhere so no changes are needed.

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -66,7 +66,7 @@ def refresh_repo(self):
         self.invoke_package_manager(self.repo_refresh)
 
     # region Get Available Updates
-    def invoke_package_manager(self, command):
+    def invoke_package_manager(self, command, raise_on_exception=True):
```
kjohn-msft, line 69: def invoke_package_manager_advanced(self, command, raise_on_exception=True):
     # full function body with additional behavior
.. return out, code

def invoke_package_manager(self, command):      # backwards compatible for 99% of cases
   out, code = self.invoke_package_manager_advanced(command, raise_on_exception=True)
   return out

kjohn-msft (changes requested):
Comment inline, as discussed.

kjohn-msft (approved):

PR #156, opened 2022-07-21
Fix CoreState.json file being empty sometimes

benank:
Sometimes the CoreState.json file can be empty, which causes a JSON decode error when reading and decoding the file. This change checks if the file is empty, and if so, writes a vanilla CoreState file to fix the problem.

[UTs in progress]

### DIFF
--- a/src/core/src/service_interfaces/LifecycleManager.py
+++ b/src/core/src/service_interfaces/LifecycleManager.py
@@ -75,6 +75,11 @@ class LifecycleManager(object):
             # Writes a vanilla core sequence file
             self.read_only_mode = False
             self.update_core_sequence()
+        elif os.path.exists(self.core_state_file_path) and os.path.isfile(self.core_state_file_path) and os.stat(self.core_state_file_path).st_size == 0:
+            # Core sequence file exists but is empty (unexpected state that will result in a JSON decode error)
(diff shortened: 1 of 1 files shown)


src/core/src/service_interfaces/LifecycleManager.py
```diff
@@ -64,18 +64,29 @@ def read_extension_sequence(self):
...
+        """ Checks for issues with the core sequence file (file not exists, is dir, etc) and attempts to mitigate them. """
         if not os.path.exists(self.core_state_file_path) or not os.path.isfile(self.core_state_file_path):
             # Neutralizes directories
             if os.path.isdir(self.core_state_file_path):
                 self.composite_logger.log_error("Core state file path returned a directory. Attempting to reset.")
                 shutil.rmtree(self.core_state_file_path)
             # Writes a vanilla core sequence file
+            self.composite_logger.log_warning("Core state file did not exist. Attempting to reset.")
+            self.read_only_mode = False
+            self.update_core_sequence()
```
kjohn-msft, line 77: Abstract into private method def write_default_core_sequence_file()
- repeated here and 79 to 81.

kjohn-msft (changes requested):
Minor comment inline.

kjohn-msft (approved):

PR #161, opened 2022-12-15
Add logs to get time taken by assessment job and install updates job

GAURAVRAMRAKHYANI:
Background: Current methods to query core execution time don’t scale well in Kusto. A single-line log can be queried better at scale across machines.

Added logs to get time taken by assessment and install updates jobs. Added relevant details in logs like patch service used (apt, yum or zypper), machine ram, machine cpu, OS details, etc.

Following are the examples of the added log for assessment and install updates job:

Assessment:

{'job': 'Assessment', 'startTime': '2022-12-15 13:40:34.314455', 'endTime': '2022-12-15 13:40:41.887585', 'patchService': 'AptitudePackageManager', 'machineDetails': {'platform_name': 'Ubuntu', 'platform_version': '20.04', 'machine_cpu': 'Intel(R) Xeon(R) Platinum 8272CL CPU @ 2.60GHz', 'machine_arch': 'x86_64', 'disk_type': 'Hard drive'}} [TC=105]

Install Updates:

{'job': 'Install Updates', 'startTime': '2022-12-15 12:36:18.153067', 'endTime': '2022-12-15 12:56:34.743747', 'installed_update_count': '13', 'update_run_successful': 'True', 'maintenance_window': '3:55:00', 'maintenance_window_exceeded': 'False', 'patchService': 'AptitudePackageManager', 'machineDetails': {'platform_name': 'Ubuntu', 'platform_version': '20.04', 'machine_cpu': 'Intel(R) Xeon(R) Platinum 8272CL CPU @ 2.60GHz', 'machine_arch': 'x86_64', 'disk_type': 'Hard drive'}} [TC=1216]

Manual testings performed:
Performed Check for updates and One-time update and checked that the new logs added are present in the logs file.


src/core/src/core_logic/PatchAssessor.py
```diff
@@ -87,6 +90,14 @@ def start_assessment(self):
                     self.status_handler.set_assessment_substatus_json(status=Constants.STATUS_ERROR)
                     raise
 
+        endTime = self.env_layer.datetime.datetime_utcnow()
+        patchService = self.package_manager.__class__.__name__
+        machine_details = TelemetryWriter.machine_info
```
rane-rajasi, line 95: Why do we need to print machine info again, since we do log those during TelemetryWriter init. (line 48 in TelemetryWriter: self.write_machine_config_info() logs these)

rane-rajasi:
A few questions and my thoughts here:

Adding new log statements to determine assessment and install jobs duration, seems a straightforward method. However, we already have the issue of the overall operation being slow because of the agent's restrictions in consuming telemetry events. Even though these are just 2 logs statements, it still contributes to the reduced pace. Most of the information added in these new logs is redundant (machine info is already logged, start and endtime are added in the status file)  

Question:

- Are we only looking at extension telemetry to determine assessment or installation duration? Since this info is already available in status file, can that be used instead? We won't need any code changes in that case

GAURAVRAMRAKHYANI:
> A few questions and my thoughts here:
> 
> Adding new log statements to determine assessment and install jobs duration, seems a straightforward method. However, we already have the issue of the overall operation being slow because of the agent's restrictions in consuming telemetry events. Even though these are just 2 logs statements, it still contributes to the reduced pace. Most of the information added in these new logs is redundant (machine info is already logged, start and endtime are added in the status file)
> 
> Question:
> 
> * Are we only looking at extension telemetry to determine assessment or installation duration? Since this info is already available in status file, can that be used instead? We won't need any code changes in that case

Yes, the information which we are logging here are already there in different places but to get all the related information for a job, currently we need join in kusto queries which is not scalable. We are logging all the information in one log so that we don't need join in kusto queries. We need efficient kusto queries because we are going to use kusto queries in future to calculate improvements that we are going to do in the script.

src/core/src/core_logic/PatchAssessor.py
GAURAVRAMRAKHYANI: The reason for this is same as I have mentioned in another comment:

Yes, the information which we are logging here are already there in different places but to get all the related information for a job, currently we need join in kusto queries which is not scalable. We are logging all the information in one log so that we don't need join in kusto queries. We need efficient kusto queries because we are going to use kusto queries in future to calculate improvements that we are going to do in the script.

src/core/src/core_logic/PatchAssessor.py
```diff
@@ -21,6 +21,7 @@
 import shutil
 import time
 from core.src.bootstrap.Constants import Constants
+from core.src.service_interfaces.TelemetryWriter import TelemetryWriter
```
kjohn-msft, line 24: TelemetryWriter already exists in this context and does not need to be imported.

Please use self.telemetryWriter that is instantiated first on line 34 and is available for the lifetime of this class instance.

src/core/src/core_logic/PatchInstaller.py
```diff
@@ -19,6 +19,7 @@
 import os
 import time
 from core.src.bootstrap.Constants import Constants
+from core.src.service_interfaces.TelemetryWriter import TelemetryWriter
```
kjohn-msft, line 22: TelemetryWriter already exists in this context and does not need to be imported.

Please use self.telemetryWriter that is instantiated first on line 32 and is available for the lifetime of this class instance.

src/core/src/core_logic/PatchAssessor.py
kjohn-msft, line 95: Please be aware that you're not logging this telemetry if there are terminal failures during assessment.

How I would suggest doing this differently overall is having a separate 'Stopwatch' class in core_logic that abstracts the stopwatch functionality out for simplified used everywhere. Consider the interface of that class and internals in terms of its consumers - I'll leave it to you but share thoughts before translating to code.

The Stopwatch class will have a dependency on telemetry writer (singleton). The stopwatch class will not be a singleton because it can have multiple instances with separate internal state. So you'd instantiate, start and stop (the stop may happen in one or more places in assessor given terminal states). Similarly for installer.

This is only a broad stroke comment - please flesh out the details and review with all reviewers. The motivation is to maximize code reusability and maintainability.

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -28,6 +28,8 @@
 class TelemetryWriter(object):
     """Class for writing telemetry data to data transports"""
 
+    machine_info = ""
+
```
kjohn-msft, line 32: Instantiate this under __init__ with self.machine_info

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -75,14 +77,14 @@ def write_package_info(self, package_name, package_ver, package_size, install_du
     # Composed payload
     def write_machine_config_info(self):
         # Machine info - sent only once at the start of the run
-        machine_info = {
+        TelemetryWriter.machine_info = {
```
kjohn-msft, line 80: self.machine_info - here and further below

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -75,14 +77,14 @@ def write_package_info(self, package_name, package_ver, package_size, install_du
...
             'platform_name': str(self.env_layer.platform.linux_distribution()[0]),
             'platform_version': str(self.env_layer.platform.linux_distribution()[1]),
             'machine_cpu': self.get_machine_processor(),
             'machine_arch': str(self.env_layer.platform.machine()),
             'disk_type': self.get_disk_type()
         }
```
kjohn-msft, line 86: rename this function to accurately represent what it is now going to be doing - **set_and_log_machine_config_info**

src/core/src/core_logic/PatchInstaller.py
```diff
@@ -88,6 +90,15 @@ def start_installation(self, simulate=False):
 
         self.composite_logger.log("\nInstalled update count: " + str(installed_update_count) + " (including dependencies)")
 
+        endTime = self.env_layer.datetime.datetime_utcnow()
+        patchService = self.package_manager.__class__.__name__
+        machine_details = TelemetryWriter.machine_info
```
kjohn-msft, line 95: Always use consistent nomenclature & naming conventions to promote clarity across the code and across code bases, and outcomes of code execution.

So end_time, package_manager_name (only for the variable to disambiguate from the class; but report as package_manager in line 97). machine_info both in name and variable in line 97.

Extrapolate this comment to all code changes - not repeating this everywhere.

src/core/src/core_logic/PatchInstaller.py
```diff
@@ -70,6 +71,7 @@ def start_installation(self, simulate=False):
                 self.composite_logger.log_debug("Attempting to reboot the machine prior to patch installation as there is a reboot pending...")
                 reboot_manager.start_reboot_if_required_and_time_available(maintenance_window.get_remaining_time_in_minutes(None, False))
 
+        startTime = self.env_layer.datetime.datetime_utcnow()
         # Install Updates
         installed_update_count, update_run_successful, maintenance_window_exceeded = self.install_updates(maintenance_window, package_manager, simulate)
```
kjohn-msft, line 76: nit note: this is a remanent of v1 -- this variable should have ideally been patch_installation_operation_successful

The nomenclature changed from "update" to "patch" from V1 to V2. "Run" became "operation". We aim to have consistency from the top of the stack to the bottom in naming for code and service quality.

You don't have to address this, but you can if you want to.

src/core/src/core_logic/PatchInstaller.py
```diff
@@ -88,6 +90,15 @@ def start_installation(self, simulate=False):
 
         self.composite_logger.log("\nInstalled update count: " + str(installed_update_count) + " (including dependencies)")
 
+        endTime = self.env_layer.datetime.datetime_utcnow()
+        patchService = self.package_manager.__class__.__name__
+        machine_details = TelemetryWriter.machine_info
+
+        InstallUpdatesDetails = {'job': "Install Updates", 'startTime': str(startTime), 'endTime': str(endTime), 'installed_update_count': str(installed_update_count), 'update_run_successful': str(update_run_successful),
+                             'maintenance_window': str(maintenance_window.duration), 'maintenance_window_exceeded': str(maintenance_window_exceeded), 'patchService': patchService, 'machineDetails': machine_details}
```
kjohn-msft, line 98: High level, this is about what is expected (naming consistency) --

        patch_installation_perf_log = {'task': "PatchInstallation", 'start_time': str(start_time), 'end_time': str(end_time), 'installed_update_count': str(installed_update_count), 'update_run_successful': str(update_run_successful),
                             'maintenance_window': str(maintenance_window.duration), 'maintenance_window_exceeded': str(maintenance_window_exceeded), 'package_manager': package_manager_name, 'machine_info': machine_info}

src/core/src/core_logic/PatchInstaller.py
```diff
@@ -88,6 +90,15 @@ def start_installation(self, simulate=False):
 
         self.composite_logger.log("\nInstalled update count: " + str(installed_update_count) + " (including dependencies)")
 
+        endTime = self.env_layer.datetime.datetime_utcnow()
+        patchService = self.package_manager.__class__.__name__
+        machine_details = TelemetryWriter.machine_info
+
+        InstallUpdatesDetails = {'job': "Install Updates", 'startTime': str(startTime), 'endTime': str(endTime), 'installed_update_count': str(installed_update_count), 'update_run_successful': str(update_run_successful),
```
kjohn-msft, line 97: I'd suggest doing the time diff right here (instead of taking that to kusto), and also measuring what percentage of the maintenance window was consumed.

kjohn-msft:
Please include test code to mock and validate changed functionality - you can do this after nailing down comments in the prod code.

kjohn-msft:
> A few questions and my thoughts here:
> 
> Adding new log statements to determine assessment and install jobs duration, seems a straightforward method. However, we already have the issue of the overall operation being slow because of the agent's restrictions in consuming telemetry events. Even though these are just 2 logs statements, it still contributes to the reduced pace. Most of the information added in these new logs is redundant (machine info is already logged, start and endtime are added in the status file)
> 
> Question:
> 
> * Are we only looking at extension telemetry to determine assessment or installation duration? Since this info is already available in status file, can that be used instead? We won't need any code changes in that case

To @rane-rajasi's note here (which I only saw after I wrote other comments), there's no need to write the existing telemetry event for machine_info if what's being written here happens reliably. Rajasi please review if that mitigates some of the valid concern raised.

kjohn-msft (changes requested):
Comments are inline in PR.

src/core/src/core_logic/PatchAssessor.py
kjohn-msft: Good question - the first one is better. But you can have two stop functions in stopwatch too - stop() - which just stops, and stop_and_write_telemetry(message) - which calls stop() internally, and then writes telemetry that concats internal attributes (start_time, end_time, time_taken) with additional fields that will go in the message. The message can be composed in the caller outside stopwatch for properties that are not internal to stopwatch.

SOLID principles in programming are a good read if you've not seen them before:
https://www.tutorialsteacher.com/csharp/single-responsibility-principle

Single responsibility and separation of concerns are good principles to lean into for design.

GAURAVRAMRAKHYANI: Removed original print machine info. Printing machine info as part of stop in Stopwatch class.

src/core/src/core_logic/PatchAssessor.py
GAURAVRAMRAKHYANI: Fixed this.

src/core/src/core_logic/PatchInstaller.py
GAURAVRAMRAKHYANI: Fixed.

src/core/src/core_logic/PatchAssessor.py
GAURAVRAMRAKHYANI: Implemented Stopwatch class with both stop() and stop_and_write_telemetry(message) methods

src/core/src/service_interfaces/TelemetryWriter.py
GAURAVRAMRAKHYANI: We don't need to initialize it. We can directly set it in set_machine_config_info() function which is called by init function

src/core/src/service_interfaces/TelemetryWriter.py
GAURAVRAMRAKHYANI: fixed

src/core/src/service_interfaces/TelemetryWriter.py
GAURAVRAMRAKHYANI: fixed

src/core/src/core_logic/PatchInstaller.py
GAURAVRAMRAKHYANI: fixed

src/core/src/core_logic/PatchInstaller.py
GAURAVRAMRAKHYANI: fixed

src/core/src/core_logic/PatchInstaller.py
GAURAVRAMRAKHYANI: done

src/core/src/core_logic/PatchAssessor.py
```diff
@@ -81,12 +88,20 @@ def start_assessment(self):
...
+                    stopwatch.stop_and_write_telemetry(str(self.assessment_perf_log))
                     self.status_handler.add_error_to_status(error_msg, Constants.PatchOperationErrorCodes.DEFAULT_ERROR)
                     if Constants.ERROR_ADDED_TO_STATUS not in repr(error):
                         error.args = (error.args, "[{0}]".format(Constants.ERROR_ADDED_TO_STATUS))
                     self.status_handler.set_assessment_substatus_json(status=Constants.STATUS_ERROR)
                     raise
 
+        assessment_perf_log = {Constants.LogStrings.TASK: Constants.LogStrings.ASSESSMENT, Constants.LogStrings.PACKAGE_MANAGER: self.package_manager_name,
+                               Constants.LogStrings.NUMBER_OF_TRIALS: str(number_of_tries), Constants.LogStrings.TASK_STATUS: Constants.LogStrings.SUCCEEDED,
+                               Constants.LogStrings.ERROR_MSG: ""}
```
rane-rajasi, line 103: This can be extracted out to a function set_assessment_perf_logs(), called on line 101 and 91 with the relevant values.

src/core/src/core_logic/PatchInstaller.py
```diff
@@ -72,22 +76,38 @@ def start_installation(self, simulate=False):
...
 
         self.composite_logger.log("\nInstalled update count: " + str(installed_update_count) + " (including dependencies)")
 
+        perc_maintenance_window_used = maintenance_window.get_percentage_maintenance_window_used()
+        patch_installation_perf_log  = {Constants.LogStrings.TASK: Constants.LogStrings.PATCH_INSTALLATION, Constants.LogStrings.PACKAGE_MANAGER: self.package_manager_name,
+                                        Constants.LogStrings.UPDATE_RUN_SUCCESSFUL: str(update_run_successful), Constants.LogStrings.INSTALLED_UPDATE_COUNT: str(installed_update_count),
+                                        Constants.LogStrings.NUMBER_OF_ROUNDS: str(number_of_rounds), Constants.LogStrings.MAINTENANCE_WINDOW: str(maintenance_window.duration),
+                                        Constants.LogStrings.PERC_MAINTENANCE_WINDOW_USED: str(perc_maintenance_window_used), Constants.LogStrings.MAINTENANCE_WINDOW_EXCEEDED: str(maintenance_window_exceeded),
+                                        Constants.LogStrings.TASK_STATUS: Constants.LogStrings.SUCCEEDED, Constants.LogStrings.ERROR_MSG: ""}
```
rane-rajasi, line 108: Similar to the comment in PatchAssessor, this can be extracted out to a function set_installer_perf_logs(), called on line 101 and 91 with the relevant values.

src/core/src/bootstrap/Constants.py
```diff
@@ -98,6 +98,33 @@ class ConfigSettings(EnumBackport):
...
 
+    # Strings used in perf logs
+    class LogStrings:
+        TASK = "task"
+        TASK_STATUS = "task_status"
+        ASSESSMENT = "Assessment"
+        PATCH_INSTALLATION = "PatchInstallation"
+        PACKAGE_MANAGER = "package_manager"
+        NUMBER_OF_TRIALS = "number_of_trials"
+        NUMBER_OF_ROUNDS = "number_of_rounds"
```
rane-rajasi, line 109: What is the difference between NUMBER_OF_TRIALS and NUMBER_OF_ROUNDS?

src/core/src/bootstrap/Constants.py
```diff
@@ -98,6 +98,33 @@ class ConfigSettings(EnumBackport):
...
+        FAILED = "failed"
+        SUCCEEDED = "succeeded"
+        ERROR_MSG = "error_msg"
+        INSTALLED_UPDATE_COUNT = "installed_update_count"
+        UPDATE_RUN_SUCCESSFUL = "update_run_successful"
```
rane-rajasi, line 114: I'm not fully convinced on adding all of these field names in here. Will think a bit more on this but in the meantime, a point to note, as Koshy mentioned in his comment, there have been some naming changes between v1 and v2. "Update" -> "patch" and "run" -> "operation"

src/core/src/bootstrap/Constants.py
```diff
@@ -98,6 +98,33 @@ class ConfigSettings(EnumBackport):
     # patch versions for healthstore when there is no maintenance run id
     PATCH_VERSION_UNKNOWN = "UNKNOWN"
 
+    # Strings used in perf logs
+    class LogStrings:
+        TASK = "task"
+        TASK_STATUS = "task_status"
+        ASSESSMENT = "Assessment"
+        PATCH_INSTALLATION = "PatchInstallation"
```
rane-rajasi, line 106: We do have operation constants already defined here (see lines 87-92, under #Operations), should re-use those

src/core/src/bootstrap/Constants.py
```diff
@@ -98,6 +98,33 @@ class ConfigSettings(EnumBackport):
...
+        MAINTENANCE_WINDOW_EXCEEDED = "maintenance_window_exceeded"
+        START_TIME = "start_time"
+        END_TIME = "end_time"
+        TIME_TAKEN = "time_taken"
+        MACHINE_INFO = "machine_info"
+        MESSAGE = "message"
+
+    # Stopwatch exception strings
+    STARTED_ALREADY = "Stopwatch is already started" 
+    STOPPED_ALREADY = "Stopwatch is already stoppped"
```
rane-rajasi, line 126: Since these are specific to Stopwatch and won't be re-used in other places, add these strings within Stopwatch

GAURAVRAMRAKHYANI: Done.

src/core/src/bootstrap/Constants.py
GAURAVRAMRAKHYANI: Done. Reusing the constant strings for Assessment and Installation

src/core/src/bootstrap/Constants.py
GAURAVRAMRAKHYANI: Done.

src/core/src/bootstrap/Constants.py
GAURAVRAMRAKHYANI: NUMBER_OF_TRIALS: Used in assessment. If there are any failures during assessment, the assessment is tried again. So it is number of tried to do asessment.

NUMBER_OF_ROUNDS: Used in install patches. Based on the flag PACKAGE_MGR_SETTING_REPEAT_PATCH_OPERATION (which is used in Zypper only), we do more than one round of install patching in some scenarios.

src/core/src/core_logic/PatchAssessor.py
GAURAVRAMRAKHYANI: Done.

src/core/src/bootstrap/Constants.py
```diff
@@ -98,6 +98,27 @@ class ConfigSettings(EnumBackport):
     # patch versions for healthstore when there is no maintenance run id
     PATCH_VERSION_UNKNOWN = "UNKNOWN"
 
+    # Strings used in perf logs
+    class LogStrings:
```
rane-rajasi, line 102: LogStrings seems generic and doesn't convey the exact use for this. How about PerfLogTrackerParams? Or open to any suggestions you have. The name should identify what it is used for

src/core/src/core_logic/MaintenanceWindow.py
```diff
@@ -71,3 +71,23 @@ def is_package_install_time_available(self, remaining_time_in_minutes=None):
...
+            elapsed_time_in_minutes = self.env_layer.datetime.total_minutes_from_time_delta(current_time - start_time)
+            percent_maintenance_window_used = (elapsed_time_in_minutes / total_time_in_minutes) * 100
+        except ValueError as error:
+            error_msg = "Error calculating percentage of maintenance window used."
+            self.composite_logger.log_error("\n" + error_msg)
+            self.status_handler.add_error_to_status(error_msg, Constants.PatchOperationErrorCodes.DEFAULT_ERROR)
+            if Constants.ERROR_ADDED_TO_STATUS not in repr(error):
+                error.args = (error.args, "[{0}]".format(Constants.ERROR_ADDED_TO_STATUS))
+            raise
+
```
rane-rajasi, line 92: Need to add UTs, that is the Codecov warning here

src/core/src/core_logic/PatchInstaller.py
```diff
@@ -72,22 +76,26 @@ def start_installation(self, simulate=False):
...
-
+        number_of_rounds = 1
         # Repeat patch installation if flagged as required and time is available
         if not maintenance_window_exceeded and package_manager.get_package_manager_setting(Constants.PACKAGE_MGR_SETTING_REPEAT_PATCH_OPERATION, False):
             self.composite_logger.log("\nInstalled update count (first round): " + str(installed_update_count))
             self.composite_logger.log("\nPatch installation run will be repeated as the package manager recommended it --------------------------------------------->")
             package_manager.set_package_manager_setting(Constants.PACKAGE_MGR_SETTING_REPEAT_PATCH_OPERATION, False)  # Resetting
             new_installed_update_count, update_run_successful, maintenance_window_exceeded = self.install_updates(maintenance_window, package_manager, simulate)
             installed_update_count += new_installed_update_count
+            number_of_rounds = number_of_rounds + 1
```
rane-rajasi, line 88: Add tests

src/core/src/core_logic/PatchInstaller.py
```diff
@@ -72,22 +76,26 @@ def start_installation(self, simulate=False):
...
 
             if package_manager.get_package_manager_setting(Constants.PACKAGE_MGR_SETTING_REPEAT_PATCH_OPERATION, False):  # We should not see this again
                 error_msg = "Unexpected repeated package manager update occurred. Please re-run the update deployment."
                 self.status_handler.add_error_to_status(error_msg, Constants.PatchOperationErrorCodes.PACKAGE_MANAGER_FAILURE)
+                self.write_installer_perf_logs(update_run_successful, installed_update_count, number_of_rounds, maintenance_window, maintenance_window_exceeded, Constants.LogStrings.FAILED, error_msg)
                 raise Exception(error_msg, "[{0}]".format(Constants.ERROR_ADDED_TO_STATUS))
```
rane-rajasi, line 93: Add tests

src/core/src/core_logic/Stopwatch.py
```diff
@@ -0,0 +1,58 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Requires Python 2.7+
+
+from core.src.bootstrap.Constants import Constants
+
+class Stopwatch(object):
```
rane-rajasi, line 19: Add UTs for all functions in here

src/core/src/bootstrap/Constants.py
```diff
@@ -98,6 +98,27 @@ class ConfigSettings(EnumBackport):
...
+        TASK = "task"
+        TASK_STATUS = "task_status"
```
rane-rajasi, line 104: nit: Can just refer to this as STATUS

src/core/src/bootstrap/Constants.py
```diff
@@ -98,6 +98,27 @@ class ConfigSettings(EnumBackport):
...
+        PACKAGE_MANAGER = "package_manager"
+        NUMBER_OF_TRIALS = "number_of_trials"
+        NUMBER_OF_ROUNDS = "number_of_rounds"
+        FAILED = "failed"
+        SUCCEEDED = "succeeded"
```
rane-rajasi, line 109: Since Failed and Succeeded are used to set the value for STATUS or (TASK_STATUS in the PR currently), better to have an enum for these two values.

src/core/src/core_logic/PatchInstaller.py
```diff
@@ -98,6 +106,16 @@ def start_installation(self, simulate=False):
...
 
+    def write_installer_perf_logs(self, patch_operation_successful, installed_patch_count, number_of_rounds, maintenance_window, maintenance_window_exceeded, task_status, error_msg):
+        perc_maintenance_window_used = maintenance_window.get_percentage_maintenance_window_used()
+
+        patch_installation_perf_log  = {Constants.LogStrings.TASK: Constants.INSTALLATION, Constants.LogStrings.PACKAGE_MANAGER: self.package_manager_name,
+                                        Constants.LogStrings.PATCH_OPERATION_SUCCESSFUL: str(patch_operation_successful), Constants.LogStrings.INSTALLED_PATCH_COUNT: str(installed_patch_count),
+                                        Constants.LogStrings.NUMBER_OF_ROUNDS: str(number_of_rounds), Constants.LogStrings.MAINTENANCE_WINDOW: str(maintenance_window.duration),
+                                        Constants.LogStrings.PERC_MAINTENANCE_WINDOW_USED: str(perc_maintenance_window_used), Constants.LogStrings.MAINTENANCE_WINDOW_EXCEEDED: str(maintenance_window_exceeded),
+                                        Constants.LogStrings.TASK_STATUS: task_status, Constants.LogStrings.ERROR_MSG: error_msg}
```
rane-rajasi, line 116: very minor: add TASK_STATUS right after, TASK. Improved readability and avoids confusions sine we also have PATCH_OPERATION_SUCCESSFUL

src/core/src/core_logic/PatchAssessor.py
```diff
@@ -81,15 +88,23 @@ def start_assessment(self):
...
                     self.status_handler.set_assessment_substatus_json(status=Constants.STATUS_ERROR)
                     raise
 
+        self.write_assessment_perf_logs(number_of_tries, Constants.LogStrings.SUCCEEDED, "")
         self.composite_logger.log("\nPatch assessment completed.\n")
         return True
 
+    def write_assessment_perf_logs(self, number_of_tries, task_status, error_msg):
+        assessment_perf_log = {Constants.LogStrings.TASK: Constants.ASSESSMENT, Constants.LogStrings.PACKAGE_MANAGER: self.package_manager_name,
+                               Constants.LogStrings.NUMBER_OF_TRIALS: str(number_of_tries), Constants.LogStrings.TASK_STATUS: task_status,
```
rane-rajasi, line 104: same comments as in Patchinstaller about keeping Task and Task_Status together

src/core/src/core_logic/Stopwatch.py
```diff
@@ -0,0 +1,58 @@
...
+        self.env_layer = env_layer
+        self.telemetry_writer = telemetry_writer
+        self.composite_logger = composite_logger
+        self.start_time = None
+        self.end_time = None
+
+        # Stopwatch exception strings
+        self.STARTED_ALREADY = "Stopwatch is already started"
+        self.STOPPED_ALREADY = "Stopwatch is already stoppped"
+
```
rane-rajasi, line 32: move these above __init__

src/core/src/core_logic/Stopwatch.py
```diff
@@ -0,0 +1,58 @@
...
+    def __del__(self):
+        # call stop only if end_time is None otherwise stop() is already called.
+        if (self.end_time == None):
```
rane-rajasi, line 35: nit: remove brackets i.e. if self.end_time == None:

src/core/src/core_logic/Stopwatch.py
```diff
@@ -0,0 +1,58 @@
...
+            self.stop()
+            self.task_details = {Constants.LogStrings.START_TIME: str(self.start_time), Constants.LogStrings.END_TIME: str(self.end_time), Constants.LogStrings.TIME_TAKEN: str(self.time_taken),
+                                 Constants.LogStrings.MACHINE_INFO: self.telemetry_writer.machine_info, Constants.LogStrings.MESSAGE: ""}
+            self.composite_logger.log(str(self.task_details))
```
rane-rajasi, line 39: Expand on the log description, something on the lines of 
self.composite_logger.log_debug(" Stopwatch details before instance is destroyed: " + str(task_details))

src/core/src/core_logic/Stopwatch.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def start(self):
+        if (self.start_time != None):
```
rane-rajasi, line 42: Remove brackets from if

rane-rajasi: Applies to all other places, not leaving comments on each of those

src/core/src/core_logic/Stopwatch.py
```diff
@@ -0,0 +1,58 @@
...
+        # Stopwatch exception strings
+        self.STARTED_ALREADY = "Stopwatch is already started"
+        self.STOPPED_ALREADY = "Stopwatch is already stoppped"
+
+    def __del__(self):
+        # call stop only if end_time is None otherwise stop() is already called.
+        if (self.end_time == None):
+            self.stop()
+            self.task_details = {Constants.LogStrings.START_TIME: str(self.start_time), Constants.LogStrings.END_TIME: str(self.end_time), Constants.LogStrings.TIME_TAKEN: str(self.time_taken),
+                                 Constants.LogStrings.MACHINE_INFO: self.telemetry_writer.machine_info, Constants.LogStrings.MESSAGE: ""}
```
rane-rajasi, line 38: This seems common between line 56 and here. Have an opportunity to extract out to a function set_task_details()

self.task_details will be set to None during init and set within __del__ and stop_and_write_telemetry()

src/core/src/core_logic/Stopwatch.py
```diff
@@ -0,0 +1,58 @@
...
+        self.end_time = self.env_layer.datetime.datetime_utcnow()
+        self.time_taken = self.env_layer.datetime.total_minutes_from_time_delta(self.end_time - self.start_time)
+
+    def stop_and_write_telemetry(self, message):
+        if (self.end_time != None):
+            raise Exception(self.STOPPED_ALREADY)
+        self.stop()
+        self.task_details = {Constants.LogStrings.START_TIME: str(self.start_time), Constants.LogStrings.END_TIME: str(self.end_time), Constants.LogStrings.TIME_TAKEN: str(self.time_taken),
+                             Constants.LogStrings.MACHINE_INFO: self.telemetry_writer.machine_info, Constants.LogStrings.MESSAGE: str(message)}
+        self.composite_logger.log(str(self.task_details))
```
rane-rajasi, line 58: Some test examples to get you started for Stopwatch:

1. To test start() function:
    Create an instance of Stopwatch(), by default self.start_time = None, verify this. Then call start(), now self.start_time() should not be None
2. Similar case for stop()
3. To test stop_and_write_telemetry(), all use cases from stop() plus verify if self.task_details and populated with expected values. If we are able to test this, don't need to verify what goes in telemetry event (the json file in events folder)

src/core/src/core_logic/PatchAssessor.py
GAURAVRAMRAKHYANI: Agree. I think error_msg should be together with task_status so should I move error_msg also with task_status?

src/core/src/core_logic/Stopwatch.py
GAURAVRAMRAKHYANI: Thanks! Added UTs

src/core/src/core_logic/Stopwatch.py
GAURAVRAMRAKHYANI: Thanks! Created new method set_task_details()

src/core/src/core_logic/Stopwatch.py
GAURAVRAMRAKHYANI: Done

src/core/src/core_logic/Stopwatch.py
GAURAVRAMRAKHYANI: Done

src/core/src/core_logic/Stopwatch.py
GAURAVRAMRAKHYANI: Done

src/core/src/core_logic/Stopwatch.py
GAURAVRAMRAKHYANI: Done

src/core/src/core_logic/PatchInstaller.py
GAURAVRAMRAKHYANI: Done

src/core/src/bootstrap/Constants.py
GAURAVRAMRAKHYANI: Done

src/core/src/core_logic/Stopwatch.py
GAURAVRAMRAKHYANI: Added UTs for all the methods.

src/core/src/core_logic/PatchInstaller.py
GAURAVRAMRAKHYANI: Added test

src/core/src/core_logic/PatchInstaller.py
GAURAVRAMRAKHYANI: Added test

src/core/src/core_logic/MaintenanceWindow.py
GAURAVRAMRAKHYANI: Added UTs.

src/core/src/bootstrap/Constants.py
GAURAVRAMRAKHYANI: Renamed to PerfLogTrackerParams

src/core/src/core_logic/Stopwatch.py
```diff
@@ -0,0 +1,69 @@
...
+        # call stop only if end_time is None otherwise stop() is already called.
+        if self.start_time is not None and self.end_time is None:
+            self.stop()
+            self.set_task_details("")
+            self.composite_logger.log("Stopwatch details before instance is destroyed: " + str(self.task_details))
+
+    def start(self):
+        if self.start_time is not None:
+            raise Exception(Stopwatch.STARTED_ALREADY)
+        self.start_time = self.env_layer.datetime.datetime_utcnow()
```
SathishMSFT, line 48: [Optional] If we are going to re-use an instance, it will be better if we reset other properties like end_time, when we are starting it.

src/core/src/core_logic/Stopwatch.py
```diff
@@ -0,0 +1,69 @@
...
+        self.composite_logger = composite_logger
+        self.start_time = None
+        self.end_time = None
+        self.time_taken = None
+        self.task_details = None
+
+    def __del__(self):
+        # if start_time is None that means Stopwatch is not started and hence no need to log
+        # call stop only if end_time is None otherwise stop() is already called.
+        if self.start_time is not None and self.end_time is None:
```
SathishMSFT, line 40: Do we need to call stop() here? even if don`t stop, the object will be deleted.

src/core/tests/Test_MaintenanceWindow.py
```diff
@@ -79,5 +79,17 @@ def test_check_available_time_after_duration_complete(self):
         self.assertEqual(runtime.maintenance_window.is_package_install_time_available(), False)
         runtime.stop()
 
+    def test_get_percentage_maintenance_window_used(self):
```
SathishMSFT, line 82: Add test for Exception ValueError if possible

src/core/src/core_logic/Stopwatch.py
GAURAVRAMRAKHYANI: Calling stop() here because in case there is some exception then the stopwatch object is disposed and this destructor is called and we want to log end_time and time_taken in this case also.

src/core/src/core_logic/MaintenanceWindow.py
rane-rajasi: It still shows L85-91 is not covered. Check if you are testing the exception scenarios

src/core/src/core_logic/PatchInstaller.py
rane-rajasi: Still shows L87 not covered

src/core/src/core_logic/PatchInstaller.py
rane-rajasi: L92 not covered

src/core/src/core_logic/Stopwatch.py
```diff
@@ -0,0 +1,69 @@
...
+            raise Exception(Stopwatch.NOT_STARTED)
+        if self.end_time is not None:
+            raise Exception(Stopwatch.STOPPED_ALREADY)
+        self.end_time = self.env_layer.datetime.datetime_utcnow()
+        self.time_taken = self.env_layer.datetime.total_minutes_from_time_delta(self.end_time - self.start_time)
+
+    def stop_and_write_telemetry(self, message):
+        if self.start_time is None:
+            raise Exception(Stopwatch.NOT_STARTED)
+        if self.end_time is not None:
```
rane-rajasi, line 61: L60 not covered in tests

src/core/src/core_logic/Stopwatch.py
```diff
@@ -0,0 +1,69 @@
...
+            raise Exception(Stopwatch.STOPPED_ALREADY)
+        self.stop()
```
rane-rajasi, line 63: L62 not covered

src/core/tests/Test_PatchAssessor.py
```diff
@@ -147,6 +147,19 @@ def test_convert_iso8601_duration_to_total_seconds(self):
         self.assertRaises(Exception, lambda: self.runtime.patch_assessor.convert_iso8601_duration_to_total_seconds('6H5M14S'))
         self.assertRaises(Exception, lambda: self.runtime.patch_assessor.convert_iso8601_duration_to_total_seconds(''))
 
+    def test_write_assessment_perf_logs(self):
+        self.runtime.patch_assessor.stopwatch = Stopwatch(self.runtime.env_layer, self.runtime.telemetry_writer, self.runtime.composite_logger)
```
rane-rajasi, line 151: Why do we need to initialize Stopwatch here? Call self.runtime.patch_assessor.start_assessment() that should auto initialize Stopwatch which also validates the initialization part.

Also, you seems like you are validating just the Stopwatch.start() function here. All of Stopwatch's functions should be validated in it's test. Test_PatchAssessor needs to validate the code flow within PatchAssessor.py

For eg: Since we added Stopwatch in start_assesment() in PatchAssessor.py, the test here would be: 
1. Call start_assessment with a Success scenario and a Failure scenario.
2. And in both cases, validate the start, end time and other params within Stopwatch and set as expected

src/core/tests/Test_PatchInstaller.py
```diff
@@ -240,6 +240,21 @@ def test_raise_if_telemetry_unsupported(self):
...
+        self.assertTrue(runtime.patch_installer.stopwatch.end_time is None)
+        self.assertTrue(runtime.patch_installer.stopwatch.time_taken is None)
+        self.assertTrue(runtime.patch_installer.stopwatch.task_details is None)
+        runtime.patch_installer.stopwatch.start()
+        self.assertTrue(runtime.patch_installer.stopwatch.start_time is not None)
+        runtime.patch_installer.write_installer_perf_logs(True, 10, 1, runtime.maintenance_window, False, Constants.TaskStatus.SUCCEEDED, "")
+        self.assertTrue(runtime.patch_installer.stopwatch.end_time is not None)
+        self.assertTrue(runtime.patch_installer.stopwatch.time_taken is not None)
+        self.assertTrue(runtime.patch_installer.stopwatch.task_details is not None)
+        runtime.stop()
```
rane-rajasi, line 256: Same comment as in Test_PatchAssessor.py

src/core/src/core_logic/Stopwatch.py
```diff
@@ -0,0 +1,69 @@
...
+from core.src.bootstrap.Constants import Constants
+
+
+class Stopwatch(object):
+    """Implements the stopwatch logic"""
+
+    # Stopwatch exception strings
+    STARTED_ALREADY = "Stopwatch is already started"
+    NOT_STARTED = "Stopwatch is not started"
+    STOPPED_ALREADY = "Stopwatch is already stoppped"
```
rane-rajasi, line 26: With the values growing, this could be an enum.

src/core/src/core_logic/PatchAssessor.py
rane-rajasi: Agreed and I see you have done this already in both PatchAssessor and PatchInstaller. Resolving this comment

src/core/src/core_logic/Stopwatch.py
kjohn-msft: Look for enumbackport in code to be python 2 compatible also

kjohn-msft (commented):
I've reviewed all the iterations so far, and it looks like the PR's getting there with mostly trivial to address comments from Rajasi and Sathish remaining. Aside from comments already there (to be addressed), I don't have any blocking comments.

src/core/src/core_logic/MaintenanceWindow.py
```diff
@@ -71,3 +71,23 @@ def is_package_install_time_available(self, remaining_time_in_minutes=None):
...
+    def get_percentage_maintenance_window_used(self):
+        """Calculate percentage of maintenance window used"""
+        try:
+            current_time = self.env_layer.datetime.datetime_utcnow()
+            start_time = self.env_layer.datetime.utc_to_standard_datetime(self.start_time)
+            dur = datetime.datetime.strptime(self.duration, "%H:%M:%S")
+            dura = timedelta(hours=dur.hour, minutes=dur.minute, seconds=dur.second)
+            total_time_in_minutes = self.env_layer.datetime.total_minutes_from_time_delta(dura)
+            elapsed_time_in_minutes = self.env_layer.datetime.total_minutes_from_time_delta(current_time - start_time)
```
rane-rajasi, line 83: We need to ensure start_time < current_time

src/core/src/core_logic/MaintenanceWindow.py
```diff
@@ -71,3 +71,23 @@ def is_package_install_time_available(self, remaining_time_in_minutes=None):
...
+            percent_maintenance_window_used = (elapsed_time_in_minutes / total_time_in_minutes) * 100
```
rane-rajasi, line 84: total_time_in_minutes != 0  or throw an error if that happens

src/core/tests/Test_PatchAssessor.py
GAURAVRAMRAKHYANI: Done

src/core/tests/Test_PatchInstaller.py
GAURAVRAMRAKHYANI: Done

src/core/src/core_logic/Stopwatch.py
GAURAVRAMRAKHYANI: Added test

src/core/src/core_logic/Stopwatch.py
GAURAVRAMRAKHYANI: Added test

src/core/tests/Test_MaintenanceWindow.py
GAURAVRAMRAKHYANI: Added test

src/core/src/core_logic/MaintenanceWindow.py
GAURAVRAMRAKHYANI: Done

src/core/src/core_logic/MaintenanceWindow.py
GAURAVRAMRAKHYANI: Throwing error if start_time > current_time

rane-rajasi:
Codecov is failing on a test failure,

image

Use the highilighted link to reach this page
image

Please ensure all tests work in a version of Python 2.7 or above and in Python 3.x

rane-rajasi (changes requested):
Comment inline

src/core/src/core_logic/MaintenanceWindow.py
```diff
@@ -71,3 +71,27 @@ def is_package_install_time_available(self, remaining_time_in_minutes=None):
...
+            if current_time < start_time:
+                raise Exception("Start time {} is greater than current time {}".format(str(start_time), str(current_time)))
+            dur = datetime.datetime.strptime(self.duration, "%H:%M:%S")
+            dura = timedelta(hours=dur.hour, minutes=dur.minute, seconds=dur.second)
+            total_time_in_minutes = self.env_layer.datetime.total_minutes_from_time_delta(dura)
+            if total_time_in_minutes == 0:
+                raise Exception("Total time in minutes is 0")
+            elapsed_time_in_minutes = self.env_layer.datetime.total_minutes_from_time_delta(current_time - start_time)
+            percent_maintenance_window_used = (elapsed_time_in_minutes / total_time_in_minutes) * 100
+        except Exception as error:
```
rane-rajasi, line 89: Since you are now catching "Exception" here than a "ValueError" that was used before, we don't need to raise Exceptions on line 81 and 86. If those scenarios occur, execution will come to L89 and throw an exception here.

Also, check L89-95, it has a bunch of things we do while raising an exception, there's no action on you for that, just as a good to know thing.

src/core/src/core_logic/MaintenanceWindow.py
```diff
@@ -71,3 +71,27 @@ def is_package_install_time_available(self, remaining_time_in_minutes=None):
...
             self.composite_logger.log_warning("Time Remaining: " + str(timedelta(seconds=int(remaining_time_in_minutes * 60))) + ", Cutoff time: " + str(timedelta(minutes=cutoff_time_in_minutes)) + " [Out of time!]")
             return False
+
+    def get_percentage_maintenance_window_used(self):
+        """Calculate percentage of maintenance window used"""
+        try:
+            current_time = self.env_layer.datetime.datetime_utcnow()
+            start_time = self.env_layer.datetime.utc_to_standard_datetime(self.start_time)
+            if current_time < start_time:
+                raise Exception("Start time {} is greater than current time {}".format(str(start_time), str(current_time)))
```
rane-rajasi, line 81: This exception will go away, check the other comment on L89, but for future reference and if it's used elsewhere, you need to use a placeholder in the string while using format() 
eg:  "Start time {**0**} is greater than current time {**1**}".format(str(start_time), str(current_time))

src/core/src/core_logic/PatchInstaller.py
```diff
@@ -72,22 +76,26 @@ def start_installation(self, simulate=False):
...
-
+        retry_count = 1
         # Repeat patch installation if flagged as required and time is available
         if not maintenance_window_exceeded and package_manager.get_package_manager_setting(Constants.PACKAGE_MGR_SETTING_REPEAT_PATCH_OPERATION, False):
             self.composite_logger.log("\nInstalled update count (first round): " + str(installed_update_count))
             self.composite_logger.log("\nPatch installation run will be repeated as the package manager recommended it --------------------------------------------->")
             package_manager.set_package_manager_setting(Constants.PACKAGE_MGR_SETTING_REPEAT_PATCH_OPERATION, False)  # Resetting
             new_installed_update_count, update_run_successful, maintenance_window_exceeded = self.install_updates(maintenance_window, package_manager, simulate)
             installed_update_count += new_installed_update_count
+            retry_count = retry_count + 1
```
rane-rajasi, line 88: L87 shows not covered. 
Suggestions to ensure coverage:

- Use ZypperPackageManager for your test. See Test_ZypperPackageManager.py for reference
- If that doesn't work, set Constants.PACKAGE_MGR_SETTING_REPEAT_PATCH_OPERATION only for this one test and make sure to reset it to it's original value, else it might break other tests

src/core/src/core_logic/PatchInstaller.py
```diff
@@ -72,22 +76,26 @@ def start_installation(self, simulate=False):
...
 
             if package_manager.get_package_manager_setting(Constants.PACKAGE_MGR_SETTING_REPEAT_PATCH_OPERATION, False):  # We should not see this again
                 error_msg = "Unexpected repeated package manager update occurred. Please re-run the update deployment."
                 self.status_handler.add_error_to_status(error_msg, Constants.PatchOperationErrorCodes.PACKAGE_MANAGER_FAILURE)
+                self.write_installer_perf_logs(update_run_successful, installed_update_count, retry_count, maintenance_window, maintenance_window_exceeded, Constants.TaskStatus.FAILED, error_msg)
                 raise Exception(error_msg, "[{0}]".format(Constants.ERROR_ADDED_TO_STATUS))
```
rane-rajasi, line 93: L92 not covered. See the other comment for inputs

src/core/src/core_logic/PatchInstaller.py
```diff
@@ -98,6 +106,16 @@ def start_installation(self, simulate=False):
 
         return overall_patch_installation_successful
 
+    def write_installer_perf_logs(self, patch_operation_successful, installed_patch_count, number_of_rounds, maintenance_window, maintenance_window_exceeded, task_status, error_msg):
+        perc_maintenance_window_used = maintenance_window.get_percentage_maintenance_window_used()
```
rane-rajasi, line 110: IF get_percentage_maintenance_window_used() throws an exception, it goes all the way up and will stop the extension execution. i.e. a failure in computing an writing perf logs stops execution. 

Checking if that is okay? @kjohn-msft @GAURAVRAMRAKHYANI @SathishMSFT

rane-rajasi: Question applies to all Stopwatch functions (i.e. start, stop, stop_and_write_telemetry) where an Exception is raised

src/core/tests/Test_PatchAssessor.py
```diff
@@ -147,6 +147,30 @@ def test_convert_iso8601_duration_to_total_seconds(self):
...
 
+    def test_write_assessment_perf_logs(self):
+        self.runtime.patch_assessor.start_assessment()
+        self.assertTrue(self.runtime.patch_assessor.stopwatch.start_time is not None)
+        self.assertTrue(self.runtime.patch_assessor.stopwatch.end_time is not None)
+        self.assertTrue(self.runtime.patch_assessor.stopwatch.time_taken is not None)
+        self.assertTrue(self.runtime.patch_assessor.stopwatch.task_details is not None)
+        self.assertTrue(self.runtime.patch_assessor.stopwatch.start_time <= self.runtime.patch_assessor.stopwatch.end_time)
+        self.assertTrue(self.runtime.patch_assessor.stopwatch.time_taken >= 0)
+        task_info = "'{}': '{}'".format(str(Constants.PerfLogTrackerParams.TASK), str(Constants.ASSESSMENT))
```
rane-rajasi, line 158: nit: use numbered index placeholder
task_info = "'{0}' : '{1}'".format(str(Constants.PerfLogTrackerParams.TASK), str(Constants.ASSESSMENT))

 in other places also

rane-rajasi: Same comment for Test_PatchInstaller.py

rane-rajasi (changes requested):
Comments inline

src/core/src/core_logic/MaintenanceWindow.py
GAURAVRAMRAKHYANI: We do need to throw exception if current_time < start_time because it will not be thrown otherwise. This function will return  negative value but not throw exception if we don't throw exception.

src/core/src/core_logic/MaintenanceWindow.py
GAURAVRAMRAKHYANI: Done for line at 86.
For line at 81, we do need to throw that exception as per reason mentioned on reply on comment on line 81.

src/core/src/core_logic/PatchInstaller.py
GAURAVRAMRAKHYANI: The code inside if block is not covered. As discussed, we can do it as a separate task.

src/core/src/core_logic/PatchInstaller.py
GAURAVRAMRAKHYANI: The code inside if block is not covered. As discussed, we can do it as a separate task.

src/core/tests/Test_PatchAssessor.py
GAURAVRAMRAKHYANI: Done

src/core/src/core_logic/PatchInstaller.py
GAURAVRAMRAKHYANI: Raising exceptions in Stopwatch functions and get_percentage_maintenance_window_used on unexpected scenarios as they are common code that are going to be used in different places. 

Swallowing exception in write_installer_perf_logs and write_assessment_perf_logs as they are specific to writing logs.

@rane-rajasi @kjohn-msft @SathishMSFT

kjohn-msft:
@GAURAVRAMRAKHYANI - can you please merge the latest from master please? There shouldn't be any conflict but it helps with the PR. @rane-rajasi please consider you're weekly planning and provide an ETA for final review. The ETA from me will be EOD business hours PST. @SathishMSFT, please review and sign off if you're satisfied with the changes.

GAURAVRAMRAKHYANI:
> @GAURAVRAMRAKHYANI - can you please merge the latest from master please? There shouldn't be any conflict but it helps with the PR. @rane-rajasi please consider you're weekly planning and provide an ETA for final review. The ETA from me will be EOD business hours PST. @SathishMSFT, please review and sign off if you're satisfied with the changes.

Merged latest master branch

src/core/src/core_logic/PatchInstaller.py
rane-rajasi: Make sure you have the follow up task in place @GAURAVRAMRAKHYANI

src/core/src/core_logic/PatchAssessor.py
```diff
@@ -81,15 +88,26 @@ def start_assessment(self):
...
         return True
 
+    def write_assessment_perf_logs(self, number_of_tries, task_status, error_msg):
+        try:
+            assessment_perf_log = {Constants.PerfLogTrackerParams.TASK: Constants.ASSESSMENT, Constants.PerfLogTrackerParams.TASK_STATUS: str(task_status),
+                                   Constants.PerfLogTrackerParams.ERROR_MSG: error_msg, Constants.PerfLogTrackerParams.PACKAGE_MANAGER: self.package_manager_name,
+                                   Constants.PerfLogTrackerParams.NUMBER_OF_TRIALS: str(number_of_tries)}
+            self.stopwatch.stop_and_write_telemetry(str(assessment_perf_log))
+        except Exception as error:
+            self.composite_logger.log_debug("Error in writing assessment performance logs. Error is: " + repr(error))
```
rane-rajasi, line 109: Everything else looks good, one minor comment

We aren't throwing Exceptions from Stopwatch class anymore, so this scenario should not occur. The only place throwing an Exception is get_percentage_maintenance_window_used() which is only called within write_installer_perf_logs()

rane-rajasi (changes requested):
Left one comment on write_assessment_perf_logs()

src/core/src/core_logic/PatchInstaller.py
```diff
@@ -98,6 +106,19 @@ def start_installation(self, simulate=False):
...
+
+            patch_installation_perf_log  = {Constants.PerfLogTrackerParams.TASK: Constants.INSTALLATION, Constants.PerfLogTrackerParams.TASK_STATUS: str(task_status), Constants.PerfLogTrackerParams.ERROR_MSG: error_msg, 
+                                            Constants.PerfLogTrackerParams.PACKAGE_MANAGER: self.package_manager_name, Constants.PerfLogTrackerParams.PATCH_OPERATION_SUCCESSFUL: str(patch_operation_successful), 
+                                            Constants.PerfLogTrackerParams.INSTALLED_PATCH_COUNT: str(installed_patch_count), Constants.PerfLogTrackerParams.NUMBER_OF_TRIALS: str(number_of_rounds), 
+                                            Constants.PerfLogTrackerParams.MAINTENANCE_WINDOW: str(maintenance_window.duration), Constants.PerfLogTrackerParams.PERC_MAINTENANCE_WINDOW_USED: str(perc_maintenance_window_used), 
+                                            Constants.PerfLogTrackerParams.MAINTENANCE_WINDOW_EXCEEDED: str(maintenance_window_exceeded)}
+            self.stopwatch.stop_and_write_telemetry(str(patch_installation_perf_log))
+        except Exception as error:
```
SathishMSFT, line 119: [optional] Move this exception to handle only L111. So, you don`t have to handle stopwatch.stop() in case of exception.

GAURAVRAMRAKHYANI: Done

src/core/src/core_logic/PatchAssessor.py
GAURAVRAMRAKHYANI: Removed the catch exception and removed corresponding unit test

SathishMSFT (approved):
Looks good to me.

src/core/src/bootstrap/Constants.py
```diff
@@ -102,6 +102,28 @@ class ConfigSettings(EnumBackport):
     # patch versions for healthstore when there is no maintenance run id
     PATCH_VERSION_UNKNOWN = "UNKNOWN"
 
+    # Strings used in perf logs
+    class PerfLogTrackerParams:
+        TASK = "task"
+        TASK_STATUS = "task_status"
+        PACKAGE_MANAGER = "package_manager"
+        NUMBER_OF_TRIALS = "number_of_trials"
```
kjohn-msft, line 110: Non-blocker to this PR, can be addressed trivially in a second PR: retry_count is a standard phrase over number_of_trials

src/core/src/core_logic/PatchAssessor.py
```diff
@@ -57,11 +62,13 @@ def start_assessment(self):
         self.composite_logger.log("\n\nGetting available patches...")
         self.package_manager.refresh_repo()
         self.status_handler.reset_assessment_data()
+        number_of_tries = 0
```
kjohn-msft, line 65: non-blocker: retry_count

src/core/src/core_logic/PatchInstaller.py
```diff
@@ -33,6 +33,7 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
         self.lifecycle_manager = lifecycle_manager
 
         self.package_manager = package_manager
+        self.package_manager_name = self.package_manager.__class__.__name__
```
kjohn-msft, line 36: Non-blocker: You're getting the class name here. The actual package manager name is obtained by: self.package_manager.get_package_manager_setting(Constants.PKG_MGR_SETTING_IDENTITY)

which will return things like Constants.APT, etc.

kjohn-msft (approved):

rane-rajasi (approved):

src/core/src/bootstrap/Constants.py
GAURAVRAMRAKHYANI: Sure, I will take care of this in a separate PR. Completing this PR.

src/core/src/core_logic/PatchAssessor.py
GAURAVRAMRAKHYANI: Sure, I will take care of this in a separate PR. Completing this PR.

src/core/src/core_logic/PatchInstaller.py
GAURAVRAMRAKHYANI: Thanks, I was not aware of this. I will take care of this in a separate PR. Completing this PR.

PR #162, opened 2022-12-17
Quick fix for the extension security issue in Ubuntu

rane-rajasi:


jemex (approved):

najams (approved):

swjai:
No, this is not the complete fix. It only handles error. I'll send a mail to you, where the error is not even thrown.

swjai:
From the script,
1. generate a guid. create filename, /tmp/{guid}/aumc-security.list, work.
2. Delete the folder after work.

Example,
---> create /tmp/123-312-3123/aumc-security.list
---> work on this.
---> delete this folder-file /tmp/123-312-3123/aumc-security.list.

rane-rajasi:
> No, this is not the complete fix. It only handles error. I'll send a mail to you, where we can do away with the error.

-- Yes this is not the complete fix, as mentioned in the title and also in Koshy's email, this is a quick fix to mitigate the issue by forcing a failure at our end. A proper fix will come in later.

PR #169, opened 2023-02-08
Use Ubuntu Pro Client to get reboot required status for ubuntu

SathishMSFT:
I have added the methods to give you an idea of the changes. Added some pre-requisites too along with this.

UT will be done when doing the feature implementation.

### DIFF
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -107,7 +107,7 @@ class Constants(object):
         TASK = "task"
         TASK_STATUS = "task_status"
         PACKAGE_MANAGER = "package_manager"
-        NUMBER_OF_TRIALS = "number_of_trials"
+        RETRY_COUNT = "retry_count"
         ERROR_MSG = "error_msg"
         INSTALLED_PATCH_COUNT = "installed_patch_count"
         PATCH_OPERATION_SUCCESSFUL = "patch_operation_successful"
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -30,7 +30,7 @@ class Constants(object):
     UNKNOWN = "Unknown"
 
     # Extension version (todo: move to a different file)
-    EXT_VERSION = "1.6.44"
+    EXT_VERSION = "1.6.45"
 
     # Runtime environments
     TEST = 'Test'
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -297,3 +297,5 @@ class Constants(object):
         NOT_SELECTED: 6
     }
 
+    # Feature flag for pro.
+    PRO_CLIENT_ENABLED = True
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -326,5 +326,6 @@ class Constants(object):
     # Ubuntu Pro Client constants.
     class UbuntuProClientSettings(EnumBackport):
         FEATURE_ENABLED = True,
+        MINIMUM_PYTHON_VERSION_REQUIRED = (3, 5)
         MAX_OS_MAJOR_VERSION_SUPPORTED = 18
 
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -326,7 +326,5 @@ class Constants(object):
     # Ubuntu Pro Client constants.
     class UbuntuProClientSettings(EnumBackport):
         FEATURE_ENABLED = True,
-        # MINIMUM_PYTHON_MAJOR_VERSION_REQUIRED = 3,
-        # MINIMUM_PYTHON_MINOR_VERSION_REQUIRED = 5,
         MAX_OS_MAJOR_VERSION_SUPPORTED = 18
 
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -326,6 +326,6 @@ class Constants(object):
     # Ubuntu Pro Client constants.
     class UbuntuProClientSettings(EnumBackport):
         FEATURE_ENABLED = True
-        MINIMUM_PYTHON_VERSION_REQUIRED = (3, 5)
+        MINIMUM_PYTHON_VERSION_REQUIRED = (3, 5)  # using tuple as we can compare this with sys.version_info. The comparison will happen in the same order. Major version checked first. Followed by Minor version.
         MAX_OS_MAJOR_VERSION_SUPPORTED = 18
 
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -325,7 +325,7 @@ class Constants(object):
 
     # Ubuntu Pro Client constants.
     class UbuntuProClientSettings(EnumBackport):
-        FEATURE_ENABLED = True,
+        FEATURE_ENABLED = True
         MINIMUM_PYTHON_VERSION_REQUIRED = (3, 5)
         MAX_OS_MAJOR_VERSION_SUPPORTED = 18
 
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -30,7 +30,7 @@ class Constants(object):
     UNKNOWN = "Unknown"
 
     # Extension version (todo: move to a different file)
-    EXT_VERSION = "1.6.43"
+    EXT_VERSION = "1.6.44"
 
     # Runtime environments
     TEST = 'Test'
@@ -101,6 +101,28 @@ class Constants(object):
 
     # patch versions for healthstore when there is no maintenance run id
     PATCH_VERSION_UNKNOWN = "UNKNOWN"
+
+    # Strings used in perf logs
+    class PerfLogTrackerParams:
+        TASK = "task"
+        TASK_STATUS = "task_status"
+        PACKAGE_MANAGER = "package_manager"
+        NUMBER_OF_TRIALS = "number_of_trials"
+        ERROR_MSG = "error_msg"
+        INSTALLED_PATCH_COUNT = "installed_patch_count"
+        PATCH_OPERATION_SUCCESSFUL = "patch_operation_successful"
+        MAINTENANCE_WINDOW = "maintenance_window"
+        PERC_MAINTENANCE_WINDOW_USED = "perc_maintenance_window_used"
+        MAINTENANCE_WINDOW_EXCEEDED = "maintenance_window_exceeded"
+        START_TIME = "start_time"
+        END_TIME = "end_time"
+        TIME_TAKEN = "time_taken"
+        MACHINE_INFO = "machine_info"
+        MESSAGE = "message"
+
+    class TaskStatus(EnumBackport):
+        SUCCEEDED = "succeeded"
+        FAILED = "failed"
 
     # Patch Modes for Configure Patching
     class PatchModes(EnumBackport):
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -30,7 +30,7 @@ class Constants(object):
     UNKNOWN = "Unknown"
 
     # Extension version (todo: move to a different file)
-    EXT_VERSION = "1.6.41"
+    EXT_VERSION = "1.6.43"
 
     # Runtime environments
     TEST = 'Test'
@@ -59,6 +59,7 @@ class Constants(object):
         CONFIG_FOLDER = "configFolder"
         STATUS_FOLDER = "statusFolder"
         EVENTS_FOLDER = "eventsFolder"
+        TEMP_FOLDER = "tempFolder"
         TELEMETRY_SUPPORTED = "telemetrySupported"
 
     class ConfigSettings(EnumBackport):
@@ -75,6 +76,9 @@ class Constants(object):
         PATCH_MODE = 'patchMode'
         ASSESSMENT_MODE = 'assessmentMode'
         MAXIMUM_ASSESSMENT_INTERVAL = 'maximumAssessmentInterval'
+
+    TEMP_FOLDER_DIR_NAME = "tmp"
+    TEMP_FOLDER_CLEANUP_ARTIFACT_LIST = ["*.list"]
 
     # File to save default settings for auto OS updates
     IMAGE_DEFAULT_PATCH_CONFIGURATION_BACKUP_PATH = "ImageDefaultPatchConfiguration.bak"
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -326,4 +326,7 @@ class Constants(object):
     # Ubuntu Pro Client constants.
     class UbuntuProClientSettings(EnumBackport):
         FEATURE_ENABLED = True,
+        # MINIMUM_PYTHON_MAJOR_VERSION_REQUIRED = 3,
+        # MINIMUM_PYTHON_MINOR_VERSION_REQUIRED = 5,
         MAX_OS_MAJOR_VERSION_SUPPORTED = 18
+
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -127,9 +127,13 @@ class Constants(object):
         DISABLED = "Disabled"
         ENABLED = "Enabled"
 
+    # To separately preserve assessment + auto-assessment state information
+    ASSESSMENT_STATE_FILE = "AssessmentState.json"
+    AUTO_ASSESSMENT_MAXIMUM_DURATION = "PT1H"
+    MIN_AUTO_ASSESSMENT_INTERVAL = "PT6H"   # do not perform auto-assessment if the last assessment happened less than this time interval ago
+
     # wait time after status updates
     WAIT_TIME_AFTER_HEALTHSTORE_STATUS_UPDATE_IN_SECS = 20
-    AUTO_ASSESSMENT_MAXIMUM_DURATION = "PT1H"
 
     # Status file states
     STATUS_TRANSITIONING = "Transitioning"
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -301,5 +301,7 @@ class Constants(object):
         NOT_SELECTED: 6
     }
 
-    # Feature flag for pro.
-    PRO_CLIENT_ENABLED = True
+    # Ubuntu Pro Client constants.
+    class UbuntuProClientSettings(EnumBackport):
+        FEATURE_ENABLED = True,
+        MAX_OS_MAJOR_VERSION_SUPPORTED = 18
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -155,8 +155,9 @@ class Constants(object):
 
     # To separately preserve assessment + auto-assessment state information
     ASSESSMENT_STATE_FILE = "AssessmentState.json"
-    AUTO_ASSESSMENT_MAXIMUM_DURATION = "PT1H"
-    MIN_AUTO_ASSESSMENT_INTERVAL = "PT6H"   # do not perform auto-assessment if the last assessment happened less than this time interval ago
+    AUTO_ASSESSMENT_MAXIMUM_DURATION = "PT1H"           # maximum time assessment is expected to take
+    AUTO_ASSESSMENT_CRON_INTERVAL = "PT1H"              # wake up to check for persistent assessment information this frequently
+    AUTO_ASSESSMENT_INTERVAL_BUFFER = "PT1H"            # allow for an hour's buffer from max interval passed down (PT6H) to keep within "max" SLA
 
     # wait time after status updates
     WAIT_TIME_AFTER_HEALTHSTORE_STATUS_UPDATE_IN_SECS = 20
--- a/src/core/src/core_logic/PatchInstaller.py
+++ b/src/core/src/core_logic/PatchInstaller.py
@@ -42,9 +42,6 @@ class PatchInstaller(object):
         self.last_still_needed_package_versions = None
         self.progress_template = "[Time available: {0} | A: {1}, S: {2}, F: {3} | D: {4}]\t {5}"
 
-        # Constants
-        self.REBOOT_PENDING_FILE_PATH = '/var/run/reboot-required'
-
     def start_installation(self, simulate=False):
         """ Kick off a patch installation run """
         self.status_handler.set_current_operation(Constants.INSTALLATION)
--- a/src/core/src/core_logic/PatchInstaller.py
+++ b/src/core/src/core_logic/PatchInstaller.py
@@ -33,7 +33,7 @@ class PatchInstaller(object):
         self.lifecycle_manager = lifecycle_manager
 
         self.package_manager = package_manager
-        self.package_manager_name = self.package_manager.__class__.__name__
+        self.package_manager_name = self.package_manager.get_package_manager_setting(Constants.PKG_MGR_SETTING_IDENTITY)
         self.package_filter = package_filter
         self.maintenance_window = maintenance_window
         self.reboot_manager = reboot_manager
@@ -106,7 +106,7 @@ class PatchInstaller(object):
 
         return overall_patch_installation_successful
 
-    def write_installer_perf_logs(self, patch_operation_successful, installed_patch_count, number_of_rounds, maintenance_window, maintenance_window_exceeded, task_status, error_msg):
+    def write_installer_perf_logs(self, patch_operation_successful, installed_patch_count, retry_count, maintenance_window, maintenance_window_exceeded, task_status, error_msg):
         perc_maintenance_window_used = -1
 
         try:
@@ -116,7 +116,7 @@ class PatchInstaller(object):
 
         patch_installation_perf_log = {Constants.PerfLogTrackerParams.TASK: Constants.INSTALLATION, Constants.PerfLogTrackerParams.TASK_STATUS: str(task_status), Constants.PerfLogTrackerParams.ERROR_MSG: error_msg,
                                        Constants.PerfLogTrackerParams.PACKAGE_MANAGER: self.package_manager_name, Constants.PerfLogTrackerParams.PATCH_OPERATION_SUCCESSFUL: str(patch_operation_successful),
-                                       Constants.PerfLogTrackerParams.INSTALLED_PATCH_COUNT: str(installed_patch_count), Constants.PerfLogTrackerParams.NUMBER_OF_TRIALS: str(number_of_rounds),
+                                       Constants.PerfLogTrackerParams.INSTALLED_PATCH_COUNT: str(installed_patch_count), Constants.PerfLogTrackerParams.RETRY_COUNT: str(retry_count),
                                        Constants.PerfLogTrackerParams.MAINTENANCE_WINDOW: str(maintenance_window.duration), Constants.PerfLogTrackerParams.PERC_MAINTENANCE_WINDOW_USED: str(perc_maintenance_window_used),
                                        Constants.PerfLogTrackerParams.MAINTENANCE_WINDOW_EXCEEDED: str(maintenance_window_exceeded)}
         self.stopwatch.stop_and_write_telemetry(str(patch_installation_perf_log))
--- a/src/core/src/core_logic/PatchInstaller.py
+++ b/src/core/src/core_logic/PatchInstaller.py
@@ -285,36 +285,6 @@ class PatchInstaller(object):
 
         return installed_update_count, patch_installation_successful, maintenance_window_exceeded
 
-    # def is_reboot_pending(self):
-    #     """ Checks if there is a pending reboot on the machine. """
-    #     try:
-    #         pending_file_exists = os.path.isfile(self.REBOOT_PENDING_FILE_PATH)
-    #         pending_processes_exists = self.package_manager.do_processes_require_restart()
-    #         self.composite_logger.log_debug(" - Reboot required debug flags: " + str(pending_file_exists) + ", " + str(pending_processes_exists) + ".")
-    #         is_reboot_required = pending_file_exists or pending_processes_exists
-    #
-    #         # Use Ubuntu Pro Client api in case of Ubuntu VMs.
-    #         if self.env_layer.get_package_manager() != Constants.APT:
-    #             return is_reboot_required
-    #         else:
-    #             self.composite_logger.log_debug("package_manager is AptitudePackageManager.")
-    #             is_pro_api_success, pro_reboot_status = self.package_manager.is_reboot_pending()
-    #             self.composite_logger.log_debug("is_pro_api_success: {0}, reboot_status {1}".format(is_pro_api_success, pro_reboot_status))
-    #
-    #             # Ubuntu Pro Client api failed. return default status.
-    #             if not is_pro_api_success:
-    #                 return is_reboot_required
-    #
-    #             # Compare Ubuntu Pro Client result and default result. Log if there is mismatch.
-    #             if is_reboot_required != pro_reboot_status:
-    #                 self.composite_logger.log_debug("Reboot status mismatch. pro = {0}, default = {1}".format(pro_reboot_status, is_reboot_required))
-    #
-    #             # Ubuntu Pro Client result is considered truth even if there is mismatch.
-    #             return pro_reboot_status
-    #     except Exception as error:
-    #         self.composite_logger.log_error('Error while checking for reboot pending: ' + repr(error))
-    #         return True     # defaults for safety
-
     def mark_installation_completed(self):
         """ Marks Installation operation as completed by updating the status of PatchInstallationSummary as success and patch metadata to be sent to healthstore.
         This is set outside of start_installation function to a restriction in CRP, where installation substatus should be marked as completed only after the implicit (2nd) assessment operation """
--- a/src/core/src/core_logic/PatchInstaller.py
+++ b/src/core/src/core_logic/PatchInstaller.py
@@ -64,7 +64,7 @@ class PatchInstaller(object):
         reboot_manager = self.reboot_manager
 
         # Early reboot if reboot is allowed by settings and required by the machine
-        reboot_pending = self.is_reboot_pending()
+        reboot_pending = self.package_manager.is_reboot_pending()
         self.status_handler.set_reboot_pending(reboot_pending)
         if reboot_pending:
             if reboot_manager.is_setting(Constants.REBOOT_NEVER):
@@ -233,7 +233,7 @@ class PatchInstaller(object):
                         self.composite_logger.log_warning("Retrying installation of package. [Package={0}]".format(package_manager.get_product_name(package_and_dependencies[0])))
 
             # Update reboot pending status in status_handler
-            self.status_handler.set_reboot_pending(self.is_reboot_pending())
+            self.status_handler.set_reboot_pending(self.package_manager.is_reboot_pending())
 
             if install_result == Constants.FAILED:
                 self.status_handler.set_package_install_status(package_manager.get_product_name(str(package_and_dependencies[0])), str(package_and_dependency_versions[0]), Constants.FAILED)
@@ -285,35 +285,35 @@ class PatchInstaller(object):
 
         return installed_update_count, patch_installation_successful, maintenance_window_exceeded
 
-    def is_reboot_pending(self):
-        """ Checks if there is a pending reboot on the machine. """
-        try:
-            pending_file_exists = os.path.isfile(self.REBOOT_PENDING_FILE_PATH)
-            pending_processes_exists = self.package_manager.do_processes_require_restart()
-            self.composite_logger.log_debug(" - Reboot required debug flags: " + str(pending_file_exists) + ", " + str(pending_processes_exists) + ".")
-            is_reboot_required = pending_file_exists or pending_processes_exists
-
-            # Use Ubuntu Pro Client api in case of Ubuntu VMs.
-            if self.env_layer.get_package_manager() != Constants.APT:
-                return is_reboot_required
-            else:
-                self.composite_logger.log_debug("package_manager is AptitudePackageManager.")
-                is_pro_api_success, pro_reboot_status = self.package_manager.is_reboot_pending()
-                self.composite_logger.log_debug("is_pro_api_success: {0}, reboot_status {1}".format(is_pro_api_success, pro_reboot_status))
-
-                # Ubuntu Pro Client api failed. return default status.
-                if not is_pro_api_success:
-                    return is_reboot_required
-
-                # Compare Ubuntu Pro Client result and default result. Log if there is mismatch.
-                if is_reboot_required != pro_reboot_status:
-                    self.composite_logger.log_debug("Reboot status mismatch. pro = {0}, default = {1}".format(pro_reboot_status, is_reboot_required))
-
-                # Ubuntu Pro Client result is considered truth even if there is mismatch.
-                return pro_reboot_status
-        except Exception as error:
-            self.composite_logger.log_error('Error while checking for reboot pending: ' + repr(error))
-            return True     # defaults for safety
+    # def is_reboot_pending(self):
+    #     """ Checks if there is a pending reboot on the machine. """
+    #     try:
+    #         pending_file_exists = os.path.isfile(self.REBOOT_PENDING_FILE_PATH)
+    #         pending_processes_exists = self.package_manager.do_processes_require_restart()
+    #         self.composite_logger.log_debug(" - Reboot required debug flags: " + str(pending_file_exists) + ", " + str(pending_processes_exists) + ".")
+    #         is_reboot_required = pending_file_exists or pending_processes_exists
+    #
+    #         # Use Ubuntu Pro Client api in case of Ubuntu VMs.
+    #         if self.env_layer.get_package_manager() != Constants.APT:
+    #             return is_reboot_required
+    #         else:
+    #             self.composite_logger.log_debug("package_manager is AptitudePackageManager.")
+    #             is_pro_api_success, pro_reboot_status = self.package_manager.is_reboot_pending()
+    #             self.composite_logger.log_debug("is_pro_api_success: {0}, reboot_status {1}".format(is_pro_api_success, pro_reboot_status))
+    #
+    #             # Ubuntu Pro Client api failed. return default status.
+    #             if not is_pro_api_success:
+    #                 return is_reboot_required
+    #
+    #             # Compare Ubuntu Pro Client result and default result. Log if there is mismatch.
+    #             if is_reboot_required != pro_reboot_status:
+    #                 self.composite_logger.log_debug("Reboot status mismatch. pro = {0}, default = {1}".format(pro_reboot_status, is_reboot_required))
+    #
+    #             # Ubuntu Pro Client result is considered truth even if there is mismatch.
+    #             return pro_reboot_status
+    #     except Exception as error:
+    #         self.composite_logger.log_error('Error while checking for reboot pending: ' + repr(error))
+    #         return True     # defaults for safety
 
     def mark_installation_completed(self):
         """ Marks Installation operation as completed by updating the status of PatchInstallationSummary as success and patch metadata to be sent to healthstore.
--- a/src/core/src/core_logic/PatchInstaller.py
+++ b/src/core/src/core_logic/PatchInstaller.py
@@ -20,7 +20,6 @@ import os
 import time
 from core.src.bootstrap.Constants import Constants
 from core.src.core_logic.Stopwatch import Stopwatch
-from core.src.package_managers import AptitudePackageManager
 
 
 class PatchInstaller(object):
--- a/src/core/src/core_logic/PatchInstaller.py
+++ b/src/core/src/core_logic/PatchInstaller.py
@@ -19,7 +19,7 @@ import datetime
 import os
 import time
 from core.src.bootstrap.Constants import Constants
-
+from core.src.core_logic.Stopwatch import Stopwatch
 
 class PatchInstaller(object):
     """" Wrapper class for a single patch installation operation """
@@ -33,6 +33,7 @@ class PatchInstaller(object):
         self.lifecycle_manager = lifecycle_manager
 
         self.package_manager = package_manager
+        self.package_manager_name = self.package_manager.__class__.__name__
         self.package_filter = package_filter
         self.maintenance_window = maintenance_window
         self.reboot_manager = reboot_manager
@@ -50,6 +51,9 @@ class PatchInstaller(object):
         self.raise_if_telemetry_unsupported()
 
         self.composite_logger.log('\nStarting patch installation...')
+
+        self.stopwatch = Stopwatch(self.env_layer, self.telemetry_writer, self.composite_logger)
+        self.stopwatch.start()
 
         self.composite_logger.log("\nMachine Id: " + self.env_layer.platform.node())
         self.composite_logger.log("Activity Id: " + self.execution_config.activity_id)
@@ -72,7 +76,7 @@ class PatchInstaller(object):
 
         # Install Updates
         installed_update_count, update_run_successful, maintenance_window_exceeded = self.install_updates(maintenance_window, package_manager, simulate)
-
+        retry_count = 1
         # Repeat patch installation if flagged as required and time is available
         if not maintenance_window_exceeded and package_manager.get_package_manager_setting(Constants.PACKAGE_MGR_SETTING_REPEAT_PATCH_OPERATION, False):
             self.composite_logger.log("\nInstalled update count (first round): " + str(installed_update_count))
@@ -80,13 +84,17 @@ class PatchInstaller(object):
             package_manager.set_package_manager_setting(Constants.PACKAGE_MGR_SETTING_REPEAT_PATCH_OPERATION, False)  # Resetting
             new_installed_update_count, update_run_successful, maintenance_window_exceeded = self.install_updates(maintenance_window, package_manager, simulate)
             installed_update_count += new_installed_update_count
+            retry_count = retry_count + 1
 
             if package_manager.get_package_manager_setting(Constants.PACKAGE_MGR_SETTING_REPEAT_PATCH_OPERATION, False):  # We should not see this again
                 error_msg = "Unexpected repeated package manager update occurred. Please re-run the update deployment."
                 self.status_handler.add_error_to_status(error_msg, Constants.PatchOperationErrorCodes.PACKAGE_MANAGER_FAILURE)
+                self.write_installer_perf_logs(update_run_successful, installed_update_count, retry_count, maintenance_window, maintenance_window_exceeded, Constants.TaskStatus.FAILED, error_msg)
                 raise Exception(error_msg, "[{0}]".format(Constants.ERROR_ADDED_TO_STATUS))
 
         self.composite_logger.log("\nInstalled update count: " + str(installed_update_count) + " (including dependencies)")
+
+        self.write_installer_perf_logs(update_run_successful, installed_update_count, retry_count, maintenance_window, maintenance_window_exceeded, Constants.TaskStatus.SUCCEEDED, "")
 
         # Reboot as per setting and environment state
         reboot_manager.start_reboot_if_required_and_time_available(maintenance_window.get_remaining_time_in_minutes(None, False))
@@ -97,6 +105,21 @@ class PatchInstaller(object):
         # NOTE: Not updating installation substatus at this point because we need to wait for the implicit/second assessment to complete first, as per CRP's instructions
 
         return overall_patch_installation_successful
+
+    def write_installer_perf_logs(self, patch_operation_successful, installed_patch_count, number_of_rounds, maintenance_window, maintenance_window_exceeded, task_status, error_msg):
+        perc_maintenance_window_used = -1
+
+        try:
+            perc_maintenance_window_used = maintenance_window.get_percentage_maintenance_window_used()
+        except Exception as error:
+            self.composite_logger.log_debug("Error in writing patch installation performance logs. Error is: " + repr(error))
+
+        patch_installation_perf_log = {Constants.PerfLogTrackerParams.TASK: Constants.INSTALLATION, Constants.PerfLogTrackerParams.TASK_STATUS: str(task_status), Constants.PerfLogTrackerParams.ERROR_MSG: error_msg,
+                                       Constants.PerfLogTrackerParams.PACKAGE_MANAGER: self.package_manager_name, Constants.PerfLogTrackerParams.PATCH_OPERATION_SUCCESSFUL: str(patch_operation_successful),
+                                       Constants.PerfLogTrackerParams.INSTALLED_PATCH_COUNT: str(installed_patch_count), Constants.PerfLogTrackerParams.NUMBER_OF_TRIALS: str(number_of_rounds),
+                                       Constants.PerfLogTrackerParams.MAINTENANCE_WINDOW: str(maintenance_window.duration), Constants.PerfLogTrackerParams.PERC_MAINTENANCE_WINDOW_USED: str(perc_maintenance_window_used),
+                                       Constants.PerfLogTrackerParams.MAINTENANCE_WINDOW_EXCEEDED: str(maintenance_window_exceeded)}
+        self.stopwatch.stop_and_write_telemetry(str(patch_installation_perf_log))
 
     def raise_if_telemetry_unsupported(self):
         if self.lifecycle_manager.get_vm_cloud_type() == Constants.VMCloudType.ARC and self.execution_config.operation not in [Constants.ASSESSMENT, Constants.INSTALLATION]:
--- a/src/core/src/core_logic/PatchInstaller.py
+++ b/src/core/src/core_logic/PatchInstaller.py
@@ -21,7 +21,6 @@ import time
 from core.src.bootstrap.Constants import Constants
 from core.src.core_logic.Stopwatch import Stopwatch
 
-
 class PatchInstaller(object):
     """" Wrapper class for a single patch installation operation """
     def __init__(self, env_layer, execution_config, composite_logger, telemetry_writer, status_handler, lifecycle_manager, package_manager, package_filter, maintenance_window, reboot_manager):
--- a/src/core/src/core_logic/PatchInstaller.py
+++ b/src/core/src/core_logic/PatchInstaller.py
@@ -20,6 +20,8 @@ import os
 import time
 from core.src.bootstrap.Constants import Constants
 from core.src.core_logic.Stopwatch import Stopwatch
+from core.src.package_managers import AptitudePackageManager
+
 
 class PatchInstaller(object):
     """" Wrapper class for a single patch installation operation """
@@ -291,7 +293,26 @@ class PatchInstaller(object):
             pending_file_exists = os.path.isfile(self.REBOOT_PENDING_FILE_PATH)
             pending_processes_exists = self.package_manager.do_processes_require_restart()
             self.composite_logger.log_debug(" - Reboot required debug flags: " + str(pending_file_exists) + ", " + str(pending_processes_exists) + ".")
-            return pending_file_exists or pending_processes_exists
+            is_reboot_required = pending_file_exists or pending_processes_exists
+
+            # Use Ubuntu Pro Client api in case of Ubuntu VMs.
+            if self.env_layer.get_package_manager() != Constants.APT:
+                return is_reboot_required
+            else:
+                self.composite_logger.log_debug("package_manager is AptitudePackageManager.")
+                is_pro_api_success, pro_reboot_status = self.package_manager.is_reboot_pending()
+                self.composite_logger.log_debug("is_pro_api_success: {0}, reboot_status {1}".format(is_pro_api_success, pro_reboot_status))
+
+                # Ubuntu Pro Client api failed. return default status.
+                if not is_pro_api_success:
+                    return is_reboot_required
+
+                # Compare Ubuntu Pro Client result and default result. Log if there is mismatch.
+                if is_reboot_required != pro_reboot_status:
+                    self.composite_logger.log_debug("Reboot status mismatch. pro = {0}, default = {1}".format(pro_reboot_status, is_reboot_required))
+
+                # Ubuntu Pro Client result is considered truth even if there is mismatch.
+                return pro_reboot_status
         except Exception as error:
             self.composite_logger.log_error('Error while checking for reboot pending: ' + repr(error))
             return True     # defaults for safety
--- a/src/core/src/package_managers/AptitudePackageManager.py
+++ b/src/core/src/package_managers/AptitudePackageManager.py
@@ -18,6 +18,7 @@
 import json
 import os
 import re
+import sys
 import uuid
 
 from core.src.package_managers.PackageManager import PackageManager
@@ -71,8 +72,7 @@ class AptitudePackageManager(PackageManager):
             self.__pro_client_prereq_met = False
             self.ubuntu_pro_client = UbuntuProClient(env_layer, composite_logger)
 
-            # Add condition to check if python3 is also installed.
-            if Constants.UbuntuProClientSettings.FEATURE_ENABLED and self.__get_os_major_version() <= Constants.UbuntuProClientSettings.MAX_OS_MAJOR_VERSION_SUPPORTED:
+            if Constants.UbuntuProClientSettings.FEATURE_ENABLED and self.__get_os_major_version() <= Constants.UbuntuProClientSettings.MAX_OS_MAJOR_VERSION_SUPPORTED and self.__is_minimum_required_python_installed():
                 self.ubuntu_pro_client.install_or_update_pro()
 
             # This flag will be used to determine if Ubuntu Pro Client can be used for querying reboot status or packages list.
@@ -553,11 +553,12 @@ class AptitudePackageManager(PackageManager):
         """check required conditions to use Ubuntu Pro Client"""
         if Constants.UbuntuProClientSettings.FEATURE_ENABLED \
                 and self.__get_os_major_version() <= Constants.UbuntuProClientSettings.MAX_OS_MAJOR_VERSION_SUPPORTED \
-                and self.ubuntu_pro_client.is_pro_working():
+                and self.ubuntu_pro_client.is_pro_working()\
+                and self.__is_minimum_required_python_installed():
             return True
         else:
-            # Add other conditions as well for debug, if needed.
-            self.composite_logger.log_debug("Feature_Enabled = {0}, OS Version = {1}".format(Constants.UbuntuProClientSettings.FEATURE_ENABLED, Constants.UbuntuProClientSettings.FEATURE_ENABLED, self.__get_os_major_version() <= Constants.UbuntuProClientSettings.MAX_OS_MAJOR_VERSION_SUPPORTED))
+            # Log each condition for debug.
+            self.composite_logger.log_debug("Feature_Enabled = {0}, OS Version = {1}, python {2}".format(Constants.UbuntuProClientSettings.FEATURE_ENABLED, Constants.UbuntuProClientSettings.FEATURE_ENABLED, self.__get_os_major_version() <= Constants.UbuntuProClientSettings.MAX_OS_MAJOR_VERSION_SUPPORTED, self.__is_minimum_required_python_installed()))
             return False
 
     def __get_os_major_version(self):
@@ -567,3 +568,7 @@ class AptitudePackageManager(PackageManager):
         os_version = self.env_layer.platform.linux_distribution()[1]
         os_major_version = int(os_version.split('.')[0])
         return os_major_version
+
+    def __is_minimum_required_python_installed(self):
+        """check if python version is at least 3.5"""
+        return sys.version_info >= (3, 5)
--- a/src/core/src/package_managers/AptitudePackageManager.py
+++ b/src/core/src/package_managers/AptitudePackageManager.py
@@ -571,7 +571,7 @@ class AptitudePackageManager(PackageManager):
                 return True
             else:
                 # Log each condition for debug.
-                self.composite_logger.log_debug("Feature_Enabled = {0}, OS Version = {1}, python {2}".format(Constants.UbuntuProClientSettings.FEATURE_ENABLED, self.__get_os_major_version() <= Constants.UbuntuProClientSettings.MAX_OS_MAJOR_VERSION_SUPPORTED, self.__is_minimum_required_python_installed()))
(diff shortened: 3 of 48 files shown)


rane-rajasi:
will review this iteration, tomorrow i.e. 2/14 PST

src/core/src/bootstrap/Constants.py
```diff
@@ -301,3 +301,5 @@ class EnvLayer(EnumBackport):
         NOT_SELECTED: 6
     }
 
+    # Feature flag for pro.
+    PRO_CLIENT_ENABLED = True
```
kjohn-msft, line 305: nit: UBUNTU_PRO_CLIENT_ENABLED

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -65,6 +66,21 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
         self.STR_DPKG_WAS_INTERRUPTED = "E: dpkg was interrupted, you must manually run 'sudo dpkg --configure -a' to correct the problem."
         self.ESM_MARKER = "The following packages could receive security updates with UA Infra: ESM service enabled:"
 
+        # pro client pre-requisite checks.
+        try:
+            self.maximum_os_version_supported = 18  # Currently we support only ubuntu 18 or lesser versions.
```
kjohn-msft, line 71: Have an enum in Constants like this:

class UbuntuProClientSettings(EnumBackport):
     FEATURE_ENABLED = True
... anything else that qualifies ...

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -65,6 +66,21 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
...
+            self.__pro_client_prereq_met = False
+            self.ubuntu_pro_client = UbuntuProClient(env_layer, composite_logger)
+
+            # Add condition to check if python3 is also installed.
+            if Constants.PRO_CLIENT_ENABLED and self.__get_os_major_version() < self.maximum_os_version_supported:
+                self.ubuntu_pro_client.install_or_update_pro()
+
+            # This flag will be used to determine if pro client can be used for querying reboot status or packages list.
```
kjohn-msft, line 79: Please use the fully qualified name in comments, code and any strings that go up the stack -- so Ubuntu Pro Client vs. just pro client.

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -65,6 +66,21 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
...
+            self.__pro_client_prereq_met = self.__is_pro_client_prereq_met()
+        except Exception as error:
+            self.composite_logger.log_debug("Pro client pre-requisite check failed. + " + repr(error))
```
kjohn-msft, line 82: This coverage warning can be met by replacing any of the functions called with a mock that throws an exception.

kjohn-msft:
Looks good, some minor comments. To get it to a mergeable state have test coverage pass even if the underlying code doesn't do anything. Or continue in this PR to the next milestone including tests.

kjohn-msft (commented):
Some minor comments - skeleton test coverage is preferred to pass coverage checks. The tests can be filled in further matching with product code changes in subsequent PRs.

kjohn-msft (changes requested):
Some minor comments - skeleton test coverage is preferred to pass coverage checks. The tests can be filled in further matching with product code changes in subsequent PRs.

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -65,6 +66,21 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
...
         self.ESM_MARKER = "The following packages could receive security updates with UA Infra: ESM service enabled:"
 
+        # pro client pre-requisite checks.
+        try:
+            self.maximum_os_version_supported = 18  # Currently we support only ubuntu 18 or lesser versions.
+            self.__pro_client_prereq_met = False
+            self.ubuntu_pro_client = UbuntuProClient(env_layer, composite_logger)
+
+            # Add condition to check if python3 is also installed.
+            if Constants.PRO_CLIENT_ENABLED and self.__get_os_major_version() < self.maximum_os_version_supported:
```
rane-rajasi, line 76: There's a difference in the major version check here and on line 556. 556 checks for <=

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -524,3 +540,27 @@ def update_os_patch_configuration_sub_setting(self, patch_configuration_sub_sett
...
+        """check required conditions to use pro client"""
+        if Constants.PRO_CLIENT_ENABLED \
+                and self.__get_os_major_version() <= self.maximum_os_version_supported \
+                and self.ubuntu_pro_client.is_pro_working():
+            return True
+        else:
+            return False
+
+    def __get_os_major_version(self):
+        """ get the os major version"""
```
rane-rajasi, line 563: Good to add a sample of what linux_distribution() would return

Also nit: """get the OS major version"""

rane-rajasi (changes requested):
Some comments inline

src/core/src/package_managers/AptitudePackageManager.py
SathishMSFT: it should be <= . Changes already made locally

src/core/src/package_managers/AptitudePackageManager.py
SathishMSFT: This code is in constructor and will be  called in the test setup phase itself. Will explore how this can be done.

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -65,6 +66,20 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
         self.STR_DPKG_WAS_INTERRUPTED = "E: dpkg was interrupted, you must manually run 'sudo dpkg --configure -a' to correct the problem."
         self.ESM_MARKER = "The following packages could receive security updates with UA Infra: ESM service enabled:"
 
+        # Ubuntu Pro Client pre-requisite checks.
+        try:
+            self.__pro_client_prereq_met = False
+            self.ubuntu_pro_client = UbuntuProClient(env_layer, composite_logger)
+
+            # Add condition to check if python3 is also installed.
```
kjohn-msft, line 74: Is this a TODO? Prefix with TODO: so both reviewers and you know it's a todo vs. a descriptive comment.

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -524,3 +539,31 @@ def update_os_patch_configuration_sub_setting(self, patch_configuration_sub_sett
     def do_processes_require_restart(self):
         """Defaulting this for Apt"""
         return False
+
+    def is_reboot_pending(self):
```
kjohn-msft, line 543: Just a heads up: because of an unrelated bug in assessment, I've delegated this function down to each package manager from the base class. That does not conflict with the direction you are taking but you'll need to reconcile and combine the function bodies. It shouldn't affect anything you were going to do - just the merge.

src/core/src/package_managers/UbuntuProClient.py
```diff
@@ -0,0 +1,45 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Requires Python 3.5+
+
+"""The is Ubuntu Pro Client implementation"""
```
kjohn-msft, line 17: _This is_ the Ubuntu Pro Client implementation ?

SathishMSFT: Will fix this. Thanks

src/core/src/package_managers/AptitudePackageManager.py
SathishMSFT: This is a TODO. Will update this in next iteration

src/core/src/package_managers/AptitudePackageManager.py
SathishMSFT: Got it. Thanks for the update.

src/core/src/core_logic/PatchInstaller.py
```diff
@@ -291,7 +293,26 @@ def is_reboot_pending(self):
...
+                # Ubuntu Pro Client api failed. return default status.
+                if not is_pro_api_success:
+                    return is_reboot_required
+
+                # Compare Ubuntu Pro Client result and default result. Log if there is mismatch.
+                if is_reboot_required != pro_reboot_status:
+                    self.composite_logger.log_debug("Reboot status mismatch. pro = {0}, default = {1}".format(pro_reboot_status, is_reboot_required))
+
+                # Ubuntu Pro Client result is considered truth even if there is mismatch.
+                return pro_reboot_status
```
kjohn-msft, line 315: In general, we'd want to avoid PackageManager specific behaviors inside generalized functions like PatchInstaller or PatchAssessor where possible.

Also this comment applies in this instance: https://github.com/Azure/LinuxPatchExtension/pull/169/files#r1107385713

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -65,6 +67,19 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
...
+            self.__pro_client_prereq_met = False
+            self.ubuntu_pro_client = UbuntuProClient(env_layer, composite_logger)
+
+            if Constants.UbuntuProClientSettings.FEATURE_ENABLED and self.__get_os_major_version() <= Constants.UbuntuProClientSettings.MAX_OS_MAJOR_VERSION_SUPPORTED and self.__is_minimum_required_python_installed():
+                self.ubuntu_pro_client.install_or_update_pro()
+
+            # This flag will be used to determine if Ubuntu Pro Client can be used for querying reboot status or packages list.
+            self.__pro_client_prereq_met = self.__is_pro_client_prereq_met()
+        except Exception as error:
+            self.composite_logger.log_debug("Ubuntu Pro Client pre-requisite check failed. + " + repr(error))
```
kjohn-msft, line 81: Minor: for better code organization, take this into a function like 'check_pro_client_prereqs', but initialize the class-level variables with defaults right in the constructor (so all of them are visible and trackable easily).

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -524,3 +539,36 @@ def update_os_patch_configuration_sub_setting(self, patch_configuration_sub_sett
...
+        """get the OS major version"""
+        # Sample output for linux_distribution():
+        # ['Ubuntu', '20.04', 'focal']
+        os_version = self.env_layer.platform.linux_distribution()[1]
+        os_major_version = int(os_version.split('.')[0])
+        return os_major_version
+
+    def __is_minimum_required_python_installed(self):
+        """check if python version is at least 3.5"""
+        return sys.version_info >= (3, 5)
```
kjohn-msft, line 574: belongs in that enum in Constants (the min version)

src/core/src/package_managers/UbuntuProClient.py
```diff
@@ -0,0 +1,74 @@
...
+    def get_all_updates(self):
+        pass
+
+    def get_other_updates(self):
+        pass
+
+    def is_reboot_pending(self):
+        """query pro api to get the reboot status"""
+        try:
+            from uaclient.api.u.pro.security.status.reboot_required.v1 import reboot_required
```
kjohn-msft, line 63: Has this code been run in the compiled py file (especially on an Ubuntu machine without this import available)?

src/core/tests/Test_UbuntuProClient.py
```diff
@@ -0,0 +1,195 @@
...
+        else:
+            version_module = imp.new_module('version_module')
+            mock_method = getattr(self, method_name)
+            setattr(version_module, mock_name, mock_method)
+            sys.modules['uaclient'] = version_module
+            sys.modules['uaclient.api'] = version_module
+            sys.modules['uaclient.api.u'] = version_module
+            sys.modules['uaclient.api.u.pro'] = version_module
+            sys.modules['uaclient.api.u.pro.version'] = version_module
+            sys.modules['uaclient.api.u.pro.version.v1'] = version_module
```
kjohn-msft, line 49: Can't this pattern be achieved with a string split and a loop that concats in each iteration? And potentially have that behavior generalized for use in other cases of this pattern?

src/core/tests/Test_UbuntuProClient.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def test_install_or_update_pro_exception(self):
+        package_manager = self.container.get('package_manager')
+        package_manager.env_layer.run_command_output = self.mock_run_command_output_raise_exception
+        self.assertIsNone(package_manager.ubuntu_pro_client.install_or_update_pro())
+
+    def test_is_pro_working_success(self):
+        obj = MockVersionResult()
+        obj.mock_import_uaclient_version_module('version', 'mock_version')
+        #self.mock_import_uaclient_version_module('version', 'mock_version')
```
kjohn-msft, line 135: code commented out

kjohn-msft (changes requested):
Comments inline

src/core/src/package_managers/AptitudePackageManager.py
SathishMSFT: @kjohn-msft: Do you mean, i have something like 
`self.is_minimum_required_python_installed = self.__is_minimum_required_python_installed()`

and use `is_minimum_required_python_installed` here? Please confirm.

src/core/src/core_logic/PatchInstaller.py
SathishMSFT: @kjohn-msft : I see that your changes are still in review. Should i wait for that to be checked-in?

src/core/tests/Test_UbuntuProClient.py
SathishMSFT: I was not sure of the approach i took here. Hence left it here for review. If you are fine with this approach, i will go ahead and refactor.

src/core/src/package_managers/UbuntuProClient.py
SathishMSFT: By the time it reaches here, pro-client will be installed. If for some reason, we don`t have the imports at this point, exception ImportError will be thrown, which is handled in the same method. I have verified this with a script locally.

SathishMSFT: To test this scenario specifically, i installed pro client in a VM, but deleted the module files manually (so import will not be possible). When i ran patch operations, i got this below error 
`        DEBUG: Exception in Ubuntu Pro Client api reboot status: ModuleNotFoundError("No module named 'uaclient.api.u.pro.security.status.reboot_required'",)
        DEBUG: is_pro_api_success: False, reboot_status False`

src/core/src/core_logic/PatchInstaller.py
```diff
@@ -20,6 +20,8 @@
 import time
 from core.src.bootstrap.Constants import Constants
 from core.src.core_logic.Stopwatch import Stopwatch
+from core.src.package_managers import AptitudePackageManager
```
kjohn-msft, line 23: note that'll need to be resolved -- shouldn't be importing this here given class hierarchies and dependencies. This is an indication that something needs to change in code organization.

src/core/src/core_logic/PatchInstaller.py
kjohn-msft: It's sort of a predecessor to what you're doing here, so it'd help in more than one way if you looked at that PR - both in influencing and aligning what you do here, and getting that in. You can pre-emptively align before it merges once you get the picture.

src/core/src/package_managers/AptitudePackageManager.py
kjohn-msft: Minor: Original comment author needs to see the change and then resolve the conversation. Otherwise it's difficult to track if a change was actually made and in the way that was discussed or expected. Unresolving for now as I don't see the change in the latest iteration yet.

src/core/src/package_managers/AptitudePackageManager.py
kjohn-msft: self.__pro_client_prereq_met being set is essentially the entire core outcome of the code right. So you can have: 

self.__pro_client_prereq_met = self.__<your function generally encapsulating lines 71 through 81>()   

The generic comment about self.<variables> was to never initialize them for the first time in some function (first initialization in the constructor is preferred with a safe-default for long-term code maintenance.

src/core/src/package_managers/UbuntuProClient.py
kjohn-msft: Sound good. To sanity-check how the imports land for my peace of mind, do you mind e-mailing me the MsftLinuxPatchCore.py file compiled from the code in the current iteration. I just want to be sure the Packager is working correctly here.

src/core/src/package_managers/AptitudePackageManager.py
SathishMSFT: Got. it. Thanks.

src/core/tests/Test_UbuntuProClient.py
kjohn-msft: We should discuss this more - maybe the Monday sync. It's okay to refactor before then because it becomes overall more maintainable long-term and easier for your to make any further changes. Lmk if there are concerns before then.

src/core/tests/Test_UbuntuProClient.py
kjohn-msft: Minor: Same thing about commenter resolving after verifying change made its way into code. Unresolving as the iteration with the change is probably not pushed.

kjohn-msft (changes requested):
Reviewed as requested.

src/core/src/package_managers/AptitudePackageManager.py
SathishMSFT: Got it. Thanks

src/core/src/core_logic/PatchInstaller.py
SathishMSFT: This is stale entry now. Will remove this. I have changed way to check if package_manager is apt.

src/core/src/package_managers/UbuntuProClient.py
SathishMSFT: Sure. Have sent you email with the output.

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -65,6 +67,15 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
...
         self.ESM_MARKER = "The following packages could receive security updates with UA Infra: ESM service enabled:"
 
+        # Ubuntu Pro Client pre-requisite checks.
+        try:
+            # This flag will be used to determine if Ubuntu Pro Client can be used for querying reboot status or get packages list.
+            self.__pro_client_prereq_met = False
+            self.ubuntu_pro_client = UbuntuProClient(env_layer, composite_logger)
+            self.check_pro_client_prerequisites()
+        except Exception as error:
+            self.composite_logger.log_debug("Ubuntu Pro Client pre-requisite check failed. + " + repr(error))
```
kjohn-msft, line 77: Don't know what happened to the comment thread here so I'm not able to refer back to the comment, but very minor difference between what was intended and what was changed - please bring up for dicussion in sync today. 

The whole try-catch block was intended for the function, with what's in line 73 and line 74 init to False and None respectively before calling the function which would do everything else. The intent is to declutter the constructor.

src/core/src/package_managers/UbuntuProClient.py
kjohn-msft: Looks good. Closing comments.

src/core/src/package_managers/UbuntuProClient.py
```diff
@@ -0,0 +1,74 @@
...
+        """install/update pro(ubuntu-advantage-tools) to the latest version"""
+        try:
+            # Install Ubuntu Pro Client.
+            code, output = self.env_layer.run_command_output(self.ubuntu_pro_client_install_cmd, False, False)
+            if code == 0:
+                self.composite_logger.log_debug("Ubuntu Pro Client installation successful. " + output)
+            else:
+                self.composite_logger.log_debug("Ubuntu Pro Client installation failed. " + output)
+        except Exception as error:
+            self.composite_logger.log_debug("Exception in installing Ubuntu Pro Client installation " + repr(error))
```
kjohn-msft, line 36: This function can return a boolean True or False, which, at minimum, you can use in tests (True for success). This can also optionally be used in the product now or later if ever warranted. As it is right now, aside from the logs written, callers never really know what happen inside the function without other environment checks.

src/core/tests/Test_UbuntuProClient.py
```diff
@@ -0,0 +1,180 @@
...
+
+    def test_install_or_update_pro_failure(self):
+        self.runtime.set_legacy_test_type('SadPath')
+        package_manager = self.container.get('package_manager')
+        self.assertIsNone(package_manager.ubuntu_pro_client.install_or_update_pro())
+
+    def test_install_or_update_pro_exception(self):
+        package_manager = self.container.get('package_manager')
+        package_manager.env_layer.run_command_output = self.mock_run_command_output_raise_exception
+        self.assertIsNone(package_manager.ubuntu_pro_client.install_or_update_pro())
```
kjohn-msft, line 116: Something's off about this test - can you check if it's actually raising the execption? This is also suggested by code coverage suggesting the mock_run_command_output_raise_exception() is not actually executing. You can also use a return within the function in places like this to have better asserts,

self.assertIsNone does nothing here because it will always be None irrespective of the code path that runs internally.

src/core/src/package_managers/AptitudePackageManager.py
kjohn-msft: Ah this comment was what I was looking for (see this for more context on related comment)

src/core/tests/Test_UbuntuProClient.py
```diff
@@ -0,0 +1,180 @@
...
+    def mock_reboot_required_return_yes(self):
+        return MockRebootRequiredResult()
+
+    def mock_reboot_required_return_no(self):
+        return MockRebootRequiredResult("no")
+
+    def mock_reboot_required_raises_exception(self):
+        raise
+
+    def mock_import_uaclient_reboot_required_module(self, mock_name, method_name):
```
SathishMSFT, line 78: Add explanation.

src/core/src/package_managers/AptitudePackageManager.py
SathishMSFT: This was the comment: 
"Minor: for better code organization, take this into a function like 'check_pro_client_prereqs', but initialize the class-level variables with defaults right in the constructor (so all of them are visible and trackable easily)."

So, as per this comment, i have the pro_client and the boolean in constructor and moved the rest to a function.

src/core/tests/Test_UbuntuProClient.py
SathishMSFT: Done.

src/core/src/package_managers/AptitudePackageManager.py
kjohn-msft: I may not have shared what I was thinking clearly -- this is what I meant in mixed pseudo-code:

<constructor block>
...
self.__pro_client_prereq_met = False
self.ubuntu_pro_client = None
<function call that sets these - including removing try-catch from constructor>
...


<function definition in this file>
	try:
            # This flag will be used to determine if Ubuntu Pro Client can be used for querying reboot status or get packages list.
            self.__pro_client_prereq_met = False
            self.ubuntu_pro_client = UbuntuProClient(env_layer, composite_logger)
            self.check_pro_client_prerequisites()
        except Exception as error:
            self.composite_logger.log_debug("Ubuntu Pro Client pre-requisite check failed. + " + repr(error))

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -524,3 +533,77 @@ def update_os_patch_configuration_sub_setting(self, patch_configuration_sub_sett
...
+        try:
+            pending_file_exists = os.path.isfile(self.REBOOT_PENDING_FILE_PATH)
+            pending_processes_exists = self.do_processes_require_restart()
+            self.composite_logger.log_debug(" - Reboot required debug flags: " + str(pending_file_exists) + ", " + str(
+                pending_processes_exists) + ".")
+            is_reboot_required = pending_file_exists or pending_processes_exists
+
+            # Use Ubuntu Pro Client api in case of Ubuntu VMs.
+            if not self.__pro_client_prereq_met:
+                self.composite_logger.log_debug("Ubuntu Pro Client pre-requisite not met.")
```
kjohn-msft, line 548: [Draft Q] Does this need to be logged here if the logging happens when __pro_client_prereq_met variable is set.

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -524,3 +533,77 @@ def update_os_patch_configuration_sub_setting(self, patch_configuration_sub_sett
...
+                return is_reboot_required
+
+            is_pro_api_success, pro_reboot_status = self.ubuntu_pro_client.is_reboot_pending()
+            self.composite_logger.log_debug(
+                "is_pro_api_success: {0}, reboot_status {1}".format(is_pro_api_success, pro_reboot_status))
```
kjohn-msft, line 553: Two part comment:
1. Nit: Try to use a single line of code if feasible because we usually have more horizontal real-estate in our dev environments than vertical. This is just a long-term productivity ask. Multi-line calls make sense where it's a very critical function call that will usually receive a lot of scrutiny.

2. Standard logging format we use is: "Descriptive but unique statement. [VariableName1=VariableValue1][VariableName2=VariableValue2]". This helps with standardized querying patterns in the end. The variable names need to be concise and not necessarily what's in code. For e.g. here: "Ubuntu Pro Client reboot status check completed. [Successful={0}][RebootStatus={1}]"

kjohn-msft: Additional note after seeing further lines: This log can be written later. See how to reduce number of log lines written without losing any critical information.

For example: can all the information we need be written in a single line, and can we have just one return statement to achieve that (to avoid too many places requiring logging before returning)?

kjohn-msft: We should discuss this further if you have questions.

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -524,3 +533,77 @@ def update_os_patch_configuration_sub_setting(self, patch_configuration_sub_sett
...
+    def __is_pro_client_prereq_met(self):
+        """check required conditions to use Ubuntu Pro Client"""
+        if Constants.UbuntuProClientSettings.FEATURE_ENABLED \
+                and self.__get_os_major_version() <= Constants.UbuntuProClientSettings.MAX_OS_MAJOR_VERSION_SUPPORTED \
+                and self.__is_minimum_required_python_installed() \
+                and self.ubuntu_pro_client.is_pro_working():
+            return True
+        else:
+            # Log each condition for debug.
+            self.composite_logger.log_debug("Feature_Enabled = {0}, OS Version = {1}, python {2}".format(Constants.UbuntuProClientSettings.FEATURE_ENABLED, self.__get_os_major_version() <= Constants.UbuntuProClientSettings.MAX_OS_MAJOR_VERSION_SUPPORTED, self.__is_minimum_required_python_installed()))
```
rane-rajasi, line 580: self.__get_os_major_version() <= Constants.UbuntuProClientSettings.MAX_OS_MAJOR_VERSION_SUPPORTED

This will log a boolean instead of the OS Version

rane-rajasi: Noticed the same in python log too. If the intention is to log whether the conditions are met, the text part could be:
"ProClientCompatibility checks: IsFeatureEnabled = {0}, IsOSVersionCompatible = {1}, IsPythonCompatible {2}"

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -524,3 +533,77 @@ def update_os_patch_configuration_sub_setting(self, patch_configuration_sub_sett
...
+            return False
+
+    def check_pro_client_prerequisites(self):
```
rane-rajasi, line 583: This and __is_pro_client_prereq_met() are very similar, barring the Exception part. Do we need 2 functions?, If yes, can the common parts be extracted?

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -524,3 +533,77 @@ def update_os_patch_configuration_sub_setting(self, patch_configuration_sub_sett
...
+        """get the OS major version"""
+        # Sample output for linux_distribution():
+        # ['Ubuntu', '20.04', 'focal']
+        os_version = self.env_layer.platform.linux_distribution()[1]
+        os_major_version = int(os_version.split('.')[0])
+        return os_major_version
+
+    def __is_minimum_required_python_installed(self):
+        """check if python version is at least 3.5"""
+        return sys.version_info >= Constants.UbuntuProClientSettings.MINIMUM_PYTHON_VERSION_REQUIRED
```
rane-rajasi, line 609: Constants.UbuntuProClientSettings.MINIMUM_PYTHON_VERSION_REQUIRED is a tuple, will this condition check against all values in it?

src/core/src/bootstrap/Constants.py
```diff
@@ -323,3 +323,9 @@ class EnvLayer(EnumBackport):
         NOT_SELECTED: 6
     }
 
+    # Ubuntu Pro Client constants.
+    class UbuntuProClientSettings(EnumBackport):
+        FEATURE_ENABLED = True,
+        MINIMUM_PYTHON_VERSION_REQUIRED = (3, 5)
```
rane-rajasi, line 329: Is this 3, 5 or 3.5? If 3.5, why a tuple?

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi, line 609: Depends on the answer for q on Constants.UbuntuProClientSettings.MINIMUM_PYTHON_VERSION_REQUIRED.

In case it is a tuple, will this condition check against all values in it?

src/core/src/package_managers/UbuntuProClient.py
```diff
@@ -0,0 +1,77 @@
...
+    def install_or_update_pro(self):
+        """install/update pro(ubuntu-advantage-tools) to the latest version"""
+        try:
+            # Install Ubuntu Pro Client.
+            code, output = self.env_layer.run_command_output(self.ubuntu_pro_client_install_cmd, False, False)
+            if code == 0:
+                self.composite_logger.log_debug("Ubuntu Pro Client installation successful. " + output)
+                return True
+            else:
+                self.composite_logger.log_debug("Ubuntu Pro Client installation failed. " + output)
```
rane-rajasi, line 35: Do we need to add the output here to error details within status blob?

src/core/src/package_managers/UbuntuProClient.py
```diff
@@ -0,0 +1,77 @@
...
+                return False
+        except Exception as error:
+            self.composite_logger.log_debug("Exception in installing Ubuntu Pro Client installation " + repr(error))
```
rane-rajasi, line 38: Similar question on status blob, does this error need to be added in status?

src/core/src/package_managers/UbuntuProClient.py
```diff
@@ -0,0 +1,77 @@
...
+            return False
+
+    def is_pro_working(self):
+        """check if pro version api returns the version without any errors/warnings."""
+        try:
+            self.composite_logger.log_debug("Ubuntu Pro Client try importing.")
+            from uaclient.api.u.pro.version.v1 import version
```
rane-rajasi, line 45: Move import outside class

rane-rajasi, line 45: Why is this import within the function?

rane-rajasi (changes requested):
Comments inline

src/core/src/bootstrap/Constants.py
SathishMSFT: using this compare the python version. 
e.g., below:
`>>> t = (3,5)
>>> (2,5) < t
True
>>> (3,5) < t
False
>>> (3,4) < t
True
>>> (3,6) < t
False
`

SathishMSFT: This looked cleaner to compare both minor and major versions.

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -524,3 +533,77 @@ def update_os_patch_configuration_sub_setting(self, patch_configuration_sub_sett
...
+    def is_reboot_pending(self):
+        """ Checks if there is a pending reboot on the machine. """
+        try:
+            pending_file_exists = os.path.isfile(self.REBOOT_PENDING_FILE_PATH)
+            pending_processes_exists = self.do_processes_require_restart()
+            self.composite_logger.log_debug(" - Reboot required debug flags: " + str(pending_file_exists) + ", " + str(
+                pending_processes_exists) + ".")
+            is_reboot_required = pending_file_exists or pending_processes_exists
+
+            # Use Ubuntu Pro Client api in case of Ubuntu VMs.
```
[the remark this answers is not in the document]
SathishMSFT: Yes. Missed this . Will change it.

src/core/src/package_managers/AptitudePackageManager.py
SathishMSFT: yes, the log for os version and python version is already available elsewhere. The intention is to find which condition failed

src/core/src/package_managers/AptitudePackageManager.py
SathishMSFT: yes. Will make this change.

src/core/src/package_managers/AptitudePackageManager.py
SathishMSFT: pasting the cases i tried out in python interpretor here for reference:
`
Python 3.6.9 (default, Nov 25 2022, 14:10:45)
[GCC 8.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> t = (3,5)
>>> (2,5) < t
True
>>> (3,5) < t
False
>>> (3,4) < t
True
>>> (3,6) < t
False
>>> (2,7) < t
True

`

src/core/src/package_managers/UbuntuProClient.py
SathishMSFT: It is to log the import related errors. Also,if pro client is not installed ( which we do in install_or_update_pro), placing the import at top may fail this.hence wanted to keep it within the method.

src/core/src/package_managers/UbuntuProClient.py
SathishMSFT: yes. This need not be in status blob. I will change this

src/core/src/package_managers/AptitudePackageManager.py
SathishMSFT: Done.

src/core/src/package_managers/UbuntuProClient.py
SathishMSFT: Yes, this is for telemetry. wanted to see if this happens frequently.

src/core/src/package_managers/AptitudePackageManager.py
SathishMSFT: Thanks. made the changes.

src/core/src/package_managers/UbuntuProClient.py
SathishMSFT: Got it. Made the changes.

src/core/tests/Test_UbuntuProClient.py
SathishMSFT: Modified the tests to check for False.

src/core/src/package_managers/AptitudePackageManager.py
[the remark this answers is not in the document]
SathishMSFT: Changes done.

src/core/src/bootstrap/Constants.py
```diff
@@ -323,3 +323,9 @@ class EnvLayer(EnumBackport):
         NOT_SELECTED: 6
     }
 
+    # Ubuntu Pro Client constants.
+    class UbuntuProClientSettings(EnumBackport):
+        FEATURE_ENABLED = True,
```
GAURAVRAMRAKHYANI, line 328: This flag is always true. This flag is used only in one if condition and in one log. Why do we need this flag?

SathishMSFT: This is by design. In case if there is an issue that we discovered, we can quickly set this to FALSE and deploy, while we look into fixing the issue.

src/core/src/package_managers/UbuntuProClient.py
```diff
@@ -0,0 +1,75 @@
...
+    def install_or_update_pro(self):
+        """install/update pro(ubuntu-advantage-tools) to the latest version"""
+        try:
+            # Install Ubuntu Pro Client.
+            code, output = self.env_layer.run_command_output(self.ubuntu_pro_client_install_cmd, False, False)
+            if code == 0:
+                self.composite_logger.log_debug("Ubuntu Pro Client installation successful.")
+                return True
+            else:
+                self.composite_logger.log_debug("Ubuntu Pro Client installation failed.")
```
GAURAVRAMRAKHYANI, line 35: Should we also log output in case of failure? It might be helpful to know the error message.

GAURAVRAMRAKHYANI, line 35: In the function install_update_and_dependencies in the file src/package_managers/PackageManager.py, there is a check self.is_package_version_installed() to confirm if package is installed. As per line 264, if code!=0 then also there is assumption that package is installed because of self.is_package_version_installed returned true.
So, looks like as per existing code, there is possibility that code != 0 but package is installed.

So, is it better to use directly use the function install_update_and_dependencies instead of run_command_output so that above scenario is handled?
input could be:
package_and_dependencies: ["ubuntu-advantage-tools"]
package_and_dependency_versions: [Constants.DEFAULT_UNSPECIFIED_VALUE]

Although, we are not dependent on the output of this function. We are dependent on the output of is_pro_working to determine if the pro is installed. So, it should be fine if this function returns false even if the pro is installed.

SathishMSFT: I am running command output to keep it clean. AS you mentioned, we need not worry about the output here. The source of truth is only is_pro_working() method.
This is also similar to another pre-requisite we alreadt have in yum : https://github.com/Azure/LinuxPatchExtension/blob/53dcb8822f9ac2cc33558ce7f4f066acfe9b6d13/src/core/src/package_managers/YumPackageManager.py#L50

SathishMSFT: Just removed it as this will get added to status and we don`t need it as of now.

src/core/tests/Test_UbuntuProClient.py
```diff
@@ -0,0 +1,187 @@
...
+        self.assertTrue(package_manager.ubuntu_pro_client.install_or_update_pro())
+
+    def test_install_or_update_pro_failure(self):
+        self.runtime.set_legacy_test_type('SadPath')
+        package_manager = self.container.get('package_manager')
+        self.assertFalse(package_manager.ubuntu_pro_client.install_or_update_pro())
+
+    def test_install_or_update_pro_exception(self):
+        package_manager = self.container.get('package_manager')
+        package_manager.env_layer.run_command_output = self.mock_run_command_output_raise_exception
```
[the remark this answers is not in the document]
SathishMSFT: Yes. Will do.

src/core/src/package_managers/UbuntuProClient.py
GAURAVRAMRAKHYANI: OK, resolving this comment.

GAURAVRAMRAKHYANI: ok. Resolving this comment

src/core/src/package_managers/UbuntuProClient.py
GAURAVRAMRAKHYANI: Can we have a separate verbose log for output so that it does not get in status blob but is there in telemetry?

src/core/tests/Test_UbuntuProClient.py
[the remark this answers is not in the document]
SathishMSFT: Done

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -65,6 +67,13 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
         self.STR_DPKG_WAS_INTERRUPTED = "E: dpkg was interrupted, you must manually run 'sudo dpkg --configure -a' to correct the problem."
         self.ESM_MARKER = "The following packages could receive security updates with UA Infra: ESM service enabled:"
 
+        # Ubuntu Pro Client pre-requisite checks.
+
```
kjohn-msft, line 71: nit: remove line between comment and code. Empty space here suggests that there is code gone missing when that's not the case.

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -65,6 +67,13 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
...
+        # This flag will be used to determine if Ubuntu Pro Client can be used for querying reboot status or get packages list.
```
kjohn-msft, line 72: nit: this comment only applies to 73 and not the whole "paragraph", so it's better to append to the end of the line that it applies to (line 73)

src/core/src/package_managers/UbuntuProClient.py
kjohn-msft: nit: "Exception in installing Ubuntu Pro Client installation " => ""Exception in installing Ubuntu Pro Client: " 
(installing .. installation is redundant)

src/core/src/package_managers/UbuntuProClient.py
```diff
@@ -0,0 +1,75 @@
...
+            result = reboot_required()
+
+            # Check if the reboot_required is yes. the values "yes-kernel-livepatches-applied"/"no" are considered as reboot not required.
+            if result.reboot_required == "yes":
+                return True, True
+            else:
+                return True, False
+        except Exception as error:
+            self.composite_logger.log_debug("Exception in Ubuntu Pro Client api reboot status: " + repr(error))
+            return False, False
```
kjohn-msft, line 75: I'm inferring this from the order of the return code - the first bool is success of the operation, and the second is if reboot is required. To avoid ambiguity, declare two local variables prior to the try.. catch. Use the first one to capture the success of the check, and the second one for the outcome (default the local reboot_required to True until set by the line 66, 69 executions - it's better to say that its needed and not need it, than to say its not needed and need it - do it redundantly here even if it's done in the caller).

A singular return statement can happen

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -858,6 +859,18 @@ def fix_ssl_certificate_issue(self):
...
+    def is_reboot_pending(self):
+        """ Checks if there is a pending reboot on the machine. """
+        try:
+            pending_file_exists = os.path.isfile(self.REBOOT_PENDING_FILE_PATH)  # not intended for yum, but supporting as back-compat
+            pending_processes_exist = self.do_processes_require_restart()
+            self.composite_logger.log_debug(" - Reboot required debug flags (yum): " + str(pending_file_exists) + ", " + str(pending_processes_exist) + ".")
+            return pending_file_exists or pending_processes_exist
+        except Exception as error:
+            self.composite_logger.log_error('Error while checking for reboot pending (yum): ' + repr(error))
+            return True  # defaults for safety
```
kjohn-msft, line 872: See this example as a reference to safe defaults talked about in this comment: https://github.com/Azure/LinuxPatchExtension/pull/169/files#r1128737191

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -524,3 +533,58 @@ def update_os_patch_configuration_sub_setting(self, patch_configuration_sub_sett
...
+
+    def check_pro_client_prerequisites(self):
+        try:
+            if Constants.UbuntuProClientSettings.FEATURE_ENABLED and self.__get_os_major_version() <= Constants.UbuntuProClientSettings.MAX_OS_MAJOR_VERSION_SUPPORTED and self.__is_minimum_required_python_installed():
+                self.ubuntu_pro_client.install_or_update_pro()
+                self.__pro_client_prereq_met = self.ubuntu_pro_client.is_pro_working()
+                return True
+            else:
+                # Log each condition for debug.
+                self.composite_logger.log_debug("ProClientPreReq checks:[IsFeatureEnabled={0}][IsOSVersionCompatible={1}][IsPythonCompatible={2}]".format(Constants.UbuntuProClientSettings.FEATURE_ENABLED, self.__get_os_major_version() <= Constants.UbuntuProClientSettings.MAX_OS_MAJOR_VERSION_SUPPORTED, self.__is_minimum_required_python_installed()))
```
kjohn-msft, line 574: You can write this log right after the try statement (so it comes out irrespective of conditions).

Also nit: standard nomenclature for pro client for consistency: "ProClientPreReq checks:" => "Ubuntu Pro Client prerequisite checks:"

src/core/src/package_managers/UbuntuProClient.py
```diff
@@ -0,0 +1,75 @@
...
+            code, output = self.env_layer.run_command_output(self.ubuntu_pro_client_install_cmd, False, False)
+            if code == 0:
+                self.composite_logger.log_debug("Ubuntu Pro Client installation successful.")
+                return True
+            else:
+                self.composite_logger.log_debug("Ubuntu Pro Client installation failed.")
+                return False
+        except Exception as error:
+            self.composite_logger.log_debug("Exception in installing Ubuntu Pro Client installation " + repr(error))
+            return False
```
kjohn-msft, line 39: Use the single log, single return pattern. Success or failure is one of the parameters in the log. The error (or lack there of ) is the other.

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -524,3 +533,58 @@ def update_os_patch_configuration_sub_setting(self, patch_configuration_sub_sett
...
+
+            # Compare Ubuntu Pro Client result and default result. Log if there is mismatch.
+            if is_reboot_required != pro_reboot_status:
+                self.composite_logger.log_debug("Reboot status mismatch. [ProClient={0}][Default={1}]".format(pro_reboot_status, is_reboot_required))
+
+            # Ubuntu Pro Client result is considered truth even if there is mismatch.
+            return pro_reboot_status
+        except Exception as error:
+            self.composite_logger.log_error('Error while checking for reboot pending: ' + repr(error))
+            return True  # defaults for safety
```
kjohn-msft, line 564: Follow up on this: https://github.com/Azure/LinuxPatchExtension/pull/169#discussion_r1123494141

Can be collapsed into a single log, single return (starting with local "default" variables as necessary to achieve as in similar comments.

Parameters to have in the single log (log itself can be: "- Reboot required advanced debug flags: "):
- [DefaultPendingFileExists={0}]
- DefaultPendingProcessesExists ...
- UbuntuProClientCheckSuccessful
- UbuntuProClientRebootStatus
- ReportedRebootStatus

src/core/src/package_managers/UbuntuProClient.py
kjohn-msft: Also, Rajasi's question was about if the customer needs to know about this failure. Not initially, may be revisited later.

kjohn-msft (changes requested):
Comments inline.

src/core/src/package_managers/AptitudePackageManager.py
SathishMSFT: Got it. Will make changes.

src/core/src/package_managers/AptitudePackageManager.py
SathishMSFT: Sure. Will do.

src/core/src/package_managers/AptitudePackageManager.py
SathishMSFT: This means i will have to re-arrange the code. Will look into this today.

SathishMSFT: Made the changes.

src/core/src/package_managers/AptitudePackageManager.py
SathishMSFT: Done.

src/core/src/package_managers/UbuntuProClient.py
SathishMSFT: Got it. will make the changes.

src/core/src/package_managers/YumPackageManager.py
SathishMSFT: Making this change accordingly.

src/core/src/package_managers/UbuntuProClient.py
SathishMSFT: Thanks. made the changes to show single log and return once.

src/core/src/package_managers/UbuntuProClient.py
SathishMSFT: Done. Thanks

src/core/src/bootstrap/Constants.py
```diff
@@ -323,3 +323,9 @@ class EnvLayer(EnumBackport):
         NOT_SELECTED: 6
     }
 
+    # Ubuntu Pro Client constants.
+    class UbuntuProClientSettings(EnumBackport):
+        FEATURE_ENABLED = True
+        MINIMUM_PYTHON_VERSION_REQUIRED = (3, 5)
```
rane-rajasi, line 329: Do add the reasoning behind why (3,5) is a tuple in a comment.

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -524,3 +531,56 @@ def update_os_patch_configuration_sub_setting(self, patch_configuration_sub_sett
...
+        default_exception = None
+        default_pending_file_exists = False
+        default_pending_processes_exists = False
+
+        # Default reboot check.
+        try:
+            default_pending_file_exists = os.path.isfile(self.REBOOT_PENDING_FILE_PATH)
+            default_pending_processes_exists = self.do_processes_require_restart()
+        except Exception as error:
+            default_exception = repr(error)
```
rane-rajasi, line 549: Do we need to log?

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -524,3 +531,56 @@ def update_os_patch_configuration_sub_setting(self, patch_configuration_sub_sett
     def do_processes_require_restart(self):
         """Defaulting this for Apt"""
         return False
+
+    def is_reboot_pending(self):
+        """ Checks if there is a pending reboot on the machine. """
```
rane-rajasi, line 536: Q of logic for is_reboot_pending between here and UbuntuProClient:
Instead of having ubuntu_pro_client_check_success, ubuntu_pro_client_reboot_status, default_exception, what if we preset is_reboot_pending to the default value (i.e. default reboot check) and only overwrite it if ubuntu_pro_prereq_is_met and is installed successfully?

src/core/tests/Test_AptitudePackageManager.py
```diff
@@ -349,6 +367,123 @@ def test_update_image_default_patch_mode_raises_exception(self):
...
+        self.assertIsNotNone(obj)
+        self.assertIsNotNone(obj.ubuntu_pro_client)
+
+        UbuntuProClient.UbuntuProClient.install_or_update_pro = backup_package_manager_ubuntu_pro_client_install_or_update_pro
+
+    def mock_is_reboot_pending_returns_True(self):
+        return True, True
+
+    def mock_is_reboot_pending_returns_False(self):
+        return False, False
```
rane-rajasi, line 423: nit: keep all mocked functions together, in this case before all tests

src/core/tests/Test_AptitudePackageManager.py
```diff
@@ -349,6 +367,123 @@ def test_update_image_default_patch_mode_raises_exception(self):
...
+        runtime.package_manager.ubuntu_pro_client.is_reboot_pending = self.mock_is_reboot_pending_returns_False
+        runtime.package_manager._AptitudePackageManager__pro_client_prereq_met = True
+
+        # test returns true because, we return True if there is exception.
+        self.assertTrue(runtime.package_manager.is_reboot_pending())
+
+        runtime.package_manager.do_processes_require_restart = backup_package_manager_do_processes_require_restart
+        runtime.package_manager.ubuntu_pro_client.is_reboot_pending = backup_package_manager_is_reboot_pending
+        version_mock.mock_unimport_uaclient_version_module()
+        reboot_mock.mock_unimport_uaclient_reboot_required_module()
```
rane-rajasi, line 474: These and the mock_import calls are used in multiple functions, can be extracted to one mock_import<>() and one mock_unimport() function, which are then called in tests

rane-rajasi: To add to this, see if there can be a setup_is_reboot_pending_test() function that extracts out common setup portions between these tests (eg: mock_imports, mocking is_reboot_pending, setting _AptitudePackageManager__pro_client_prereq_met, etc)

SathishMSFT: The mock imports can have different parameters based on the tests which will make it difficult to to refactor into a common setup code . Whatever was common between these modules are already refactored into a common class (MockSystemModules class in Test_UbuntuProClient.py) which i am planning to use for updates module too.

src/core/src/package_managers/AptitudePackageManager.py
SathishMSFT: we log every thing at the end of the function as per our last discussion on single log/ single return per method.

src/core/src/package_managers/AptitudePackageManager.py
SathishMSFT: As per our discussion, I calculate the default reboot check first. The pro client reboot status is checked next. Everything is logged at the end and pro client status is preferred if available.

src/core/src/bootstrap/Constants.py
SathishMSFT: Added comment.

src/core/tests/Test_AptitudePackageManager.py
SathishMSFT: Done.Thanks

src/core/src/package_managers/AptitudePackageManager.py
kjohn-msft: Looks good. Closing.

src/core/tests/Test_AptitudePackageManager.py
kjohn-msft: This seems to have been missed. 419 through 423 is supposed to be moved to around line 55.

SathishMSFT: The checkin was not pushed. pushed it now.

kjohn-msft:
@SathishMSFT please ensure @nikhim-um and @GAURAVRAMRAKHYANI sign off if they are good with the changes today - it's important they catch things so more time-consuming iterations don't need to happen outside timezones.

@rane-rajasi to also review this and the pre-dependency in AUM GA bugfixes. Hopefully that merges today or tomorrow, and you can get latest from master.

Mapping it out, this PR could potentially merge Thursday PST / Friday morning IST. And we'd take this too in the end of week release.

src/core/tests/library/LegacyEnvLayerExtensions.py
```diff
@@ -538,7 +540,10 @@ def run_command_output(self, cmd, no_output=False, chk_err=True):
                              "model name	: Intel(R) Core(TM) i7-6700 CPU @ 3.40GHz\n" + \
                              "model name	: Intel(R) Core(TM) i7-6700 CPU @ 3.40GHz"
                 elif self.legacy_package_manager_name is Constants.APT:
-                    output = ''
+                    if cmd.find('sudo apt-get install ubuntu-advantage-tools -y') > -1:
+                        code = 1
+                    else:
```
GAURAVRAMRAKHYANI, line 545: nit: else is not required. The default value of output is '' so no need to set it.

src/core/tests/Test_ZypperPackageManager.py
```diff
@@ -635,6 +635,17 @@ def test_package_manager_exit_repeat_operation(self):
...
+    def test_is_reboot_pending_return_true_when_exception_raised(self):
+        package_manager = self.container.get('package_manager')
+        backup_do_process_require_restart = package_manager.do_processes_require_restart
+        package_manager.do_processes_require_restart = self.mock_do_processes_require_restart_raise_exception
+
+        self.assertTrue(package_manager.is_reboot_pending())
+
+        package_manager.do_processes_require_restart = backup_do_process_require_restart
+
+    def mock_do_processes_require_restart_raise_exception(self):
```
GAURAVRAMRAKHYANI, line 647: nit: mock_ functions can be placed together.

src/core/tests/Test_YumPackageManager.py
```diff
@@ -583,6 +583,18 @@ def test_update_image_default_patch_mode_raises_exception(self):
...
+    def test_is_reboot_pending_return_true_when_exception_raised(self):
+        package_manager = self.container.get('package_manager')
+        backup_do_process_require_restart = package_manager.do_processes_require_restart
+        package_manager.do_processes_require_restart = self.mock_do_processes_require_restart_raise_exception
+
+        self.assertTrue(package_manager.is_reboot_pending())
+
+        package_manager.do_processes_require_restart = backup_do_process_require_restart
+
+    def mock_do_processes_require_restart_raise_exception(self):
```
GAURAVRAMRAKHYANI, line 595: nit: mock_ functions can be placed together.

src/core/tests/library/LegacyEnvLayerExtensions.py
SathishMSFT: Done

src/core/tests/Test_ZypperPackageManager.py
SathishMSFT: Done

src/core/tests/Test_YumPackageManager.py
SathishMSFT: Done

src/core/tests/Test_AptitudePackageManager.py
```diff
@@ -349,6 +373,117 @@ def test_update_image_default_patch_mode_raises_exception(self):
...
+        package_manager = self.container.get('package_manager')
+        package_manager._AptitudePackageManager__pro_client_prereq_met = False
+
+        self.assertFalse(package_manager.is_reboot_pending())
+
+    def test_is_reboot_pending_prerequisite_met_should_return_true(self):
+        package_manager = self.container.get('package_manager')
+        package_manager._AptitudePackageManager__pro_client_prereq_met = True
+
+        self.assertTrue(True, package_manager.is_reboot_pending())
```
GAURAVRAMRAKHYANI, line 386: Here, the statement should be self.assertTrue(package_manager.is_reboot_pending()) or self.assertEqual(True, package_manager.is_reboot_pending()).

I did debug and found that here package_manager.is_reboot_pending() is returning False because of exception while finding reboot status using pro.

SathishMSFT: done. Thanks

src/core/tests/Test_AptitudePackageManager.py
```diff
@@ -349,6 +373,117 @@ def test_update_image_default_patch_mode_raises_exception(self):
...
+        runtime.package_manager._AptitudePackageManager__pro_client_prereq_met = True
+        self.assertFalse(runtime.package_manager.is_reboot_pending())
+
+        runtime.package_manager._AptitudePackageManager__pro_client_prereq_met = backup_AptitudePackageManager__pro_client_prereq_met
+        version_mock.mock_unimport_uaclient_version_module()
+        reboot_mock.mock_unimport_uaclient_reboot_required_module()
+
+    def test_is_reboot_pending_test_mismatch(self):
+        version_mock = MockVersionResult()
+        version_mock.mock_import_uaclient_version_module('version', 'mock_version')
```
GAURAVRAMRAKHYANI, line 445: There is no need of version_mock because setting  runtime.package_manager._AptitudePackageManager__pro_client_prereq_met = True

Also, there is no need of reboot_mock as setting runtime.package_manager.ubuntu_pro_client.is_reboot_pending = self.mock_is_reboot_pending_returns_True

src/core/tests/Test_AptitudePackageManager.py
```diff
@@ -349,6 +373,117 @@ def test_update_image_default_patch_mode_raises_exception(self):
...
+
+        obj = AptitudePackageManager.AptitudePackageManager(package_manager.env_layer, execution_config, package_manager.composite_logger, package_manager.telemetry_writer, package_manager.status_handler)
+
+        self.assertIsNotNone(obj)
+        self.assertIsNotNone(obj.ubuntu_pro_client)
+
+        UbuntuProClient.UbuntuProClient.install_or_update_pro = backup_package_manager_ubuntu_pro_client_install_or_update_pro
+
+    def test_is_reboot_pending_pro_client_success(self):
+        version_mock = MockVersionResult()
```
GAURAVRAMRAKHYANI, line 430: There is no need of version_mock because setting runtime.package_manager._AptitudePackageManager__pro_client_prereq_met = True

src/core/tests/Test_AptitudePackageManager.py
```diff
@@ -349,6 +373,117 @@ def test_update_image_default_patch_mode_raises_exception(self):
...
+
+    def test_is_reboot_pending_test_mismatch(self):
+        version_mock = MockVersionResult()
+        version_mock.mock_import_uaclient_version_module('version', 'mock_version')
+        reboot_mock = MockRebootRequiredResult()
+        reboot_mock.mock_import_uaclient_reboot_required_module('reboot_required', 'mock_reboot_required_return_yes')
+        runtime = RuntimeCompositor(ArgumentComposer().get_composed_arguments(), True, Constants.APT)
+        backup_package_manager_is_reboot_pending = runtime.package_manager.ubuntu_pro_client.is_reboot_pending
+        runtime.package_manager.ubuntu_pro_client.is_reboot_pending = self.mock_is_reboot_pending_returns_True
+        runtime.package_manager._AptitudePackageManager__pro_client_prereq_met = True
```
GAURAVRAMRAKHYANI, line 451: runtime.package_manager._AptitudePackageManager__pro_client_prereq_met is not reverted back in test case test_is_reboot_pending_test_mismatch but it is reverted back in the test case test_is_reboot_pending_pro_client_success. I think it should be reverted back in end of the test case test_is_reboot_pending_test_mismatch.

src/core/tests/Test_AptitudePackageManager.py
```diff
@@ -349,6 +373,117 @@ def test_update_image_default_patch_mode_raises_exception(self):
...
+
+        # test should return true as we fall back to Ubuntu Pro Client api`s result.
+        self.assertTrue(runtime.package_manager.is_reboot_pending())
+
+        runtime.package_manager.ubuntu_pro_client.is_reboot_pending = backup_package_manager_is_reboot_pending
+        version_mock.mock_unimport_uaclient_version_module()
+        reboot_mock.mock_unimport_uaclient_reboot_required_module()
+
+    def test_is_reboot_pending_test_raises_exception(self):
+        version_mock = MockVersionResult()
```
GAURAVRAMRAKHYANI, line 461: There is no need of version_mock because setting runtime.package_manager._AptitudePackageManager__pro_client_prereq_met = True

Also, there is no need of reboot_mock as setting runtime.package_manager.ubuntu_pro_client.is_reboot_pending = self.mock_is_reboot_pending_returns_False

src/core/tests/Test_AptitudePackageManager.py
```diff
@@ -349,6 +373,117 @@ def test_update_image_default_patch_mode_raises_exception(self):
...
+        version_mock.mock_import_uaclient_version_module('version', 'mock_version')
+        reboot_mock = MockRebootRequiredResult()
+        reboot_mock.mock_import_uaclient_reboot_required_module('reboot_required', 'mock_reboot_required_return_yes')
+        runtime = RuntimeCompositor(ArgumentComposer().get_composed_arguments(), True, Constants.APT)
+        backup_package_manager_do_processes_require_restart = runtime.package_manager.do_processes_require_restart
+        runtime.package_manager.do_processes_require_restart = self.mock_do_processes_require_restart_raises_exception
+        backup_package_manager_is_reboot_pending = runtime.package_manager.ubuntu_pro_client.is_reboot_pending
+        runtime.package_manager.ubuntu_pro_client.is_reboot_pending = self.mock_is_reboot_pending_returns_False
+        runtime.package_manager._AptitudePackageManager__pro_client_prereq_met = True
```
GAURAVRAMRAKHYANI, line 470: Need to revert back runtime.package_manager._AptitudePackageManager__pro_client_prereq_met

GAURAVRAMRAKHYANI:
Please share the list of manual scenarios tested.

src/core/tests/Test_AptitudePackageManager.py
SathishMSFT: Done.

src/core/tests/Test_AptitudePackageManager.py
SathishMSFT: Made the changes to remove the reboot_pending method mock. Done.,

src/core/tests/Test_AptitudePackageManager.py
SathishMSFT: Done

src/core/tests/Test_AptitudePackageManager.py
SathishMSFT: Done

src/core/tests/Test_AptitudePackageManager.py
```diff
@@ -35,6 +38,27 @@ def mock_read_with_retry_raise_exception(self):
...
+    def mock_minimum_required_python_installed_return_true(self):
+        return True
+
+    def mock_install_or_update_pro_raise_exception(self):
+        raise Exception
+
+    def mock_do_processes_require_restart_raises_exception(self):
+        raise Exception
+
+    def mock_is_reboot_pending_returns_True(self):
```
GAURAVRAMRAKHYANI, line 56: This method is not required now

GAURAVRAMRAKHYANI (dismissed):

src/core/src/core_logic/PatchInstaller.py
```diff
@@ -64,7 +64,7 @@ def start_installation(self, simulate=False):
         reboot_manager = self.reboot_manager
 
         # Early reboot if reboot is allowed by settings and required by the machine
-        reboot_pending = self.is_reboot_pending()
+        reboot_pending = self.package_manager.is_reboot_pending()
```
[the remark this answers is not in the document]
SathishMSFT: As per code, we don`t call this reboot required anywhere in assessment. So, its not needed.

nikhim-um (dismissed):

src/core/src/core_logic/PatchInstaller.py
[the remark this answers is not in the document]
kjohn-msft: One of the bugfixes in the code you'll be re-merging from master is to have it called in assessment (it should have always been there). Forwarding an e-mail that you'd have gotten last month. 

Will look in depth but calling out to save time: if there's a implicit dependency on it not being called in assessment, that needs to be addressed.

GAURAVRAMRAKHYANI (dismissed):
Looks good to me.

src/core/tests/Test_AptitudePackageManager.py
SathishMSFT: removed

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -521,17 +528,59 @@ def update_os_patch_configuration_sub_setting(self, patch_configuration_sub_sett
     # endregion
 
     # region Reboot Management
+    def do_processes_require_restart(self):
+        """ Fulfilling base class contract """
+        return False
+    # endregion Reboot Management
```
GAURAVRAMRAKHYANI, line 534: nit: small suggestion - "# endregion Reboot Management" should be after def is_reboot_pending

GAURAVRAMRAKHYANI (approved):
Looks good to me. Just a small suggestion added as a comment.

nikhim-um (approved):

src/core/tests/Test_YumPackageManager.py
```diff
@@ -597,6 +600,15 @@ def test_update_image_default_patch_mode_raises_exception(self):
         self.runtime.env_layer.file_system.write_with_retry = self.mock_write_with_retry_raise_exception
         self.assertRaises(Exception, package_manager.update_os_patch_configuration_sub_setting)
 
+    def test_is_reboot_pending_return_true_when_exception_raised(self):
+        package_manager = self.container.get('package_manager')
+        backup_do_process_require_restart = package_manager.do_processes_require_restart
+        package_manager.do_processes_require_restart = self.mock_do_processes_require_restart_raise_exception
+
```
rane-rajasi, line 607: very minor: It's okay if we don't have an extra line break for smaller test sections. Eg: there's just one assert here, so I wouldn't  add the empty line on 607 and 609

rane-rajasi (approved):
Left a minor comment

src/core/src/package_managers/ZypperPackageManager.py
```diff
@@ -758,13 +758,13 @@ def update_os_patch_configuration_sub_setting(self, patch_configuration_sub_sett
     def is_reboot_pending(self):
         """ Checks if there is a pending reboot on the machine. """
         try:
-            pending_file_exists = os.path.isfile(self.REBOOT_PENDING_FILE_PATH)     # not intended for zypper, but supporting as back-compat
```
GAURAVRAMRAKHYANI, line 761: The changes in this file is only removing of spaces. These changes can now be reverted to remove this file from PR.
Same for the YumPackageManager.py file.

kjohn-msft (approved):

src/core/tests/Test_YumPackageManager.py
SathishMSFT: Keeping this as is to make it easier to separate the setup/ test /clean up stages.

src/core/src/package_managers/ZypperPackageManager.py
SathishMSFT: This is merge code. for now  i will keep this.

src/core/src/package_managers/AptitudePackageManager.py
SathishMSFT: Will take this in next PR. as its minor.

src/core/src/package_managers/AptitudePackageManager.py
SathishMSFT: @kjohn-msft : Is this good to resolve?

GAURAVRAMRAKHYANI (approved):

PR #179, opened 2023-03-16
Additional python interpreter locations

kjohn-msft:
Including additional python interpreter locations for reliable execution.

### DIFF
--- a/src/extension/src/MsftLinuxPatchExtShim.sh
+++ b/src/extension/src/MsftLinuxPatchExtShim.sh
@@ -36,7 +36,7 @@ function find_python(){
     local python_exec_command=$1
 
     # Check if there is python defined.
-    for p in python3 python python2 /usr/libexec/platform-python; do
+    for p in python3 python python2 /usr/libexec/platform-python /usr/share/oem/python/bin/python3 /usr/share/oem/python/bin/python; do
         if command -v "${p}" ; then
             eval ${python_exec_command}=${p}
             return
--- a/src/extension/src/MsftLinuxPatchExtShim.sh
+++ b/src/extension/src/MsftLinuxPatchExtShim.sh
@@ -36,7 +36,7 @@ function find_python(){
     local python_exec_command=$1
 
     # Check if there is python defined.
-    for p in python3 python python2 /usr/libexec/platform-python /usr/share/oem/python/bin/python3 /usr/share/oem/python/bin/python; do
+    for p in python3 /usr/share/oem/python/bin/python3 python python2 /usr/libexec/platform-python /usr/share/oem/python/bin/python; do
         if command -v "${p}" ; then
             eval ${python_exec_command}=${p}
             return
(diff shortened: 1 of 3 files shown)


src/extension/src/MsftLinuxPatchExtShim.sh
```diff
@@ -36,7 +36,7 @@ function find_python(){
     local python_exec_command=$1
 
     # Check if there is python defined.
-    for p in python3 python python2 /usr/libexec/platform-python; do
+    for p in python3 python python2 /usr/libexec/platform-python /usr/share/oem/python/bin/python3 /usr/share/oem/python/bin/python; do
```
rane-rajasi, line 39: We'll need to add these and "/usr/libexec/platform-python" in /extension/src/ProcessHandler.py --> get_python_cmd() also, for Extension to invoke Core successfully

kjohn-msft: Are we really missing 2 existing entries (last 2) + the 2 new ones there??

kjohn-msft: Updated to auto-discover from manual discovery. Manual discovery is only a fallback for now. To be removed later.

feng-j678 (approved):

src/extension/src/ProcessHandler.py
```diff
@@ -131,6 +131,16 @@ def stage_auto_assess_sh_safely(self, core_process_command):
             self.logger.log_error("Unable to stage auto-assess shim. [Error={0}]".format(str(error)))
 
     def get_python_cmd(self):
+        # auto-discover what is already being used by extension
+        python_path = os.path.realpath(sys.executable)
+        code, out = self.env_layer.run_command_output("which " + python_path, False, False)
```
rane-rajasi, line 136: This will not discover python from /usr/share/oem/python/bin/python3 or /usr/libexec/platform-python or /usr/share/oem/python/bin/python paths. causing handler to eventually fail with python not found error, which was the case found in RHEL 8-gen2

kjohn-msft: Why is that? It's inheriting the path already discovered by the shell script that launched handler.

rane-rajasi: Did not see line 135 where it fetches from what is already used. In that case, we only need to verify, with line 135-136, that python paths are correctly discovered.

kjohn-msft: Good to go then? There is a second PR (left open in parallel) to remove the old code once this is proven in production with a release. If this code fails (0.01% chance), the old code picks up.

rane-rajasi (approved):

src/extension/src/ProcessHandler.py
rane-rajasi: Yes.

PR #180, opened 2023-03-21
Reporting-related mitigation on auto-assessment with configurepatching & cleaned up traces in StatusHandler

kjohn-msft:
Addresses an additional mitigation required by an internal higher-level component in status reporting.

Noticed traces in StatusHandler while reviewing the fix and opportunistically cleaned that up too.

### DIFF
--- a/src/core/src/service_interfaces/StatusHandler.py
+++ b/src/core/src/service_interfaces/StatusHandler.py
@@ -156,11 +156,10 @@ class StatusHandler(object):
     def set_package_install_status(self, package_names, package_versions, status="Pending", classification=None):
         """ Externally available method to set installation status for one or more packages of the **SAME classification and status** """
         self.composite_logger.log_debug("Setting package installation status in bulk. [Count={0}]".format(str(len(package_names))))
-
         package_names, package_versions = self.validate_packages_being_installed(package_names, package_versions)
+        package_install_status_summary = ""
 
         for package_name, package_version in zip(package_names, package_versions):
-            self.composite_logger.log_debug("Logging progress [Package: " + package_name + "; Status: " + status + "]")
             patch_already_saved = False
             patch_id = self.__get_patch_id(package_name, package_version)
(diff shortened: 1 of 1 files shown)


src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -323,8 +328,9 @@ def __new_assessment_summary_json(self, assessment_packages_json, status, code):
             else:
                 other_patch_count += 1
 
-        # discern started by
-        started_by = Constants.PatchAssessmentSummaryStartedBy.PLATFORM if self.execution_config.exec_auto_assess_only else Constants.PatchAssessmentSummaryStartedBy.USER
+        # discern started by - either pure auto-assessment or assessment data being included with configure patching with assessmentMode set to AutomaticByPlatform
+        include_assessment_with_configure_patching = (self.execution_config.operation == Constants.CONFIGURE_PATCHING and self.execution_config.assessment_mode == Constants.AssessmentModes.AUTOMATIC_BY_PLATFORM)
+        started_by = Constants.PatchAssessmentSummaryStartedBy.PLATFORM if (self.execution_config.exec_auto_assess_only or include_assessment_with_configure_patching) else Constants.PatchAssessmentSummaryStartedBy.USER
```
rane-rajasi, line 334: With these checks changing overtime, can we have a centralized place which defines what happens when? For eg: in this case, include_assessment_with_configure_patching can be computed in some common location and then be re-used in multiple places (status updates, started_by, etc)

kjohn-msft: Will refactor this in the context of the bigger PR (that started later): https://github.com/Azure/LinuxPatchExtension/pull/183 
Plan is to take it into a 'derived' field in execution config and fan out from there.

rane-rajasi (approved):

rane-rajasi (approved):

PR #183, opened 2023-03-23
Always run Assessment with ConfigurePatching - CRP/ARG issue mitigation

kjohn-msft:
**Reason for this change:**

Configurepatching results are wrapped inside assessment results at Azure Resource Graph. 

Which means that some assessment result needs to be written for the configurepatching results to be written. CRP can fake it – but none of the outcomes were deemed acceptable:
-	Writing success with defaults (0 security, 0 critical, etc) – this is misleading as it pretends the machine is secure. (Not good at scale either)
-	Writing warning with defaults (0 security, etc) – in current consumption experiences, this is also misleading. And the warning would be overwriting older but more accurate assessment results.
-	Writing failed with defaults – this will be considered a bug as assessment never failed. It was never attempted (or is still running). (At scale, some portion of a fleet will always have ‘failed’ assessment even though nothing failed)

If the Windows/Linux extensions attempt to write assessment (as ‘Platform’) with every configure patching result, this will always allow CRP to write accurate information to ARG (even if it fails at the extension).

**This PR mitigates the identified root cause in the Linux Patch Extension.**

Note: includes a change from another open PR that may merge before this one.

### DIFF
--- a/src/core/src/CoreMain.py
+++ b/src/core/src/CoreMain.py
@@ -79,7 +79,9 @@ class CoreMain(object):
             try:
                 patch_assessment_successful = patch_assessor.start_assessment()
             except Exception as error:
-                assessment_exception_error = error
+                assessment_exception_error = error  # hold this until configure patching is closed out
+
+            # close out configure patching if needed & then raise any 'uncontrolled' assessment exception if it occurred
             if not execution_config.exec_auto_assess_only and patch_operation_requested == Constants.CONFIGURE_PATCHING.lower():
                 configure_patching_processor.set_configure_patching_final_overall_status()  # guarantee configure patching status write prior to throwing on any catastrophic assessment error
             if assessment_exception_error is not None:
--- a/src/core/src/CoreMain.py
+++ b/src/core/src/CoreMain.py
@@ -74,10 +74,10 @@ class CoreMain(object):
             if not execution_config.exec_auto_assess_only:
                 configure_patching_successful = configure_patching_processor.start_configure_patching()
 
-            # Assessment happens for an Auto Assessment request or for all Non Auto Assessment operations, except for ConfigurePatching iff AssessmentMode is set to AutomaticByPlatform
-            include_assessment_with_configure_patching = (patch_operation_requested == Constants.CONFIGURE_PATCHING.lower() and execution_config.assessment_mode == Constants.AssessmentModes.AUTOMATIC_BY_PLATFORM)
-            if execution_config.exec_auto_assess_only or patch_operation_requested != Constants.CONFIGURE_PATCHING.lower() or include_assessment_with_configure_patching:
-                patch_assessment_successful = patch_assessor.start_assessment()
+            # Assessment happens for all operations. If the goal seeking tracked operation is CP, then its final status can only be written after assessment reaches a terminal state
+            patch_assessment_successful = patch_assessor.start_assessment()
+            if not execution_config.exec_auto_assess_only and patch_operation_requested == Constants.CONFIGURE_PATCHING.lower():
+                configure_patching_processor.set_configure_patching_final_overall_status()
 
             # Patching + additional assessment occurs if the operation is 'Installation' and not Auto Assessment. Need to check both since operation_requested from prev run is preserved in Auto Assessment
             if not execution_config.exec_auto_assess_only and patch_operation_requested == Constants.INSTALLATION.lower():
--- a/src/core/src/CoreMain.py
+++ b/src/core/src/CoreMain.py
@@ -75,9 +75,15 @@ class CoreMain(object):
                 configure_patching_successful = configure_patching_processor.start_configure_patching()
 
             # Assessment happens for all operations. If the goal seeking tracked operation is CP, then its final status can only be written after assessment reaches a terminal state
-            patch_assessment_successful = patch_assessor.start_assessment()
+            assessment_exception_error = None
+            try:
+                patch_assessment_successful = patch_assessor.start_assessment()
+            except Exception as error:
+                assessment_exception_error = error
             if not execution_config.exec_auto_assess_only and patch_operation_requested == Constants.CONFIGURE_PATCHING.lower():
-                configure_patching_processor.set_configure_patching_final_overall_status()
+                configure_patching_processor.set_configure_patching_final_overall_status()  # guarantee configure patching status write prior to throwing on any catastrophic assessment error
+            if assessment_exception_error is not None:
+                raise assessment_exception_error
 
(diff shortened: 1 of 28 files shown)


src/core/src/core_logic/ConfigurePatchingProcessor.py
```diff
@@ -47,16 +48,28 @@ def start_configure_patching(self):
...
         return self.configure_patching_successful
 
+    def set_configure_patching_final_overall_status(self):
+        """ Writes the final overall status after any pre-requisite operation is also in a terminal state - currently this is only assessment """
+        overall_status = Constants.STATUS_SUCCESS if self.configure_patching_successful else Constants.STATUS_ERROR
+        if self.configure_patching_exception_error is None:
+            self.__report_consolidated_configure_patch_status(status=overall_status)
+        else:
+            self.__report_consolidated_configure_patch_status(status=overall_status, error=self.configure_patching_exception_error)
+
```
feng-j678, line 72: doesn't this function's assertion of the status cover the above logic?

def test_patch_mode_set_failure_for_configure_patching(self):  
        self.assertTrue(substatus_file_data[0]["status"].lower() == Constants.STATUS_ERROR.lower())
        self.assertEqual(len(substatus_file_data), 2)
        self.assertTrue(substatus_file_data[0]["name"] == Constants.PATCH_ASSESSMENT_SUMMARY)   # assessment is 
        now part of the CP flow
        self.assertTrue(substatus_file_data[0]["status"].lower() == Constants.STATUS_SUCCESS.lower())
        self.assertTrue(substatus_file_data[1]["name"] == Constants.CONFIGURE_PATCHING_SUMMARY)
        self.assertTrue(substatus_file_data[1]["status"].lower() == Constants.STATUS_ERROR.lower())

src/core/src/CoreMain.py
```diff
@@ -74,10 +74,10 @@ def __init__(self, argv):
...
 
-            # Assessment happens for an Auto Assessment request or for all Non Auto Assessment operations, except for ConfigurePatching iff AssessmentMode is set to AutomaticByPlatform
-            include_assessment_with_configure_patching = (patch_operation_requested == Constants.CONFIGURE_PATCHING.lower() and execution_config.assessment_mode == Constants.AssessmentModes.AUTOMATIC_BY_PLATFORM)
-            if execution_config.exec_auto_assess_only or patch_operation_requested != Constants.CONFIGURE_PATCHING.lower() or include_assessment_with_configure_patching:
-                patch_assessment_successful = patch_assessor.start_assessment()
+            # Assessment happens for all operations. If the goal seeking tracked operation is CP, then its final status can only be written after assessment reaches a terminal state
+            patch_assessment_successful = patch_assessor.start_assessment()
+            if not execution_config.exec_auto_assess_only and patch_operation_requested == Constants.CONFIGURE_PATCHING.lower():
+                configure_patching_processor.set_configure_patching_final_overall_status()
```
mirichmo, line 80: If assessment throws (raises?) an exception, does the except ensure a final status is written?

kjohn-msft: Assessment has built in full-cover exception handling with retries, and there's status reconciliation in the highest-level exception handling, but there's still a slim possibility of a super-fundamental problem (like being unable to log something basic that is never expected to fail and would need our attention anyway) that could throw this off. Evaluating if something additional should be done for that.

src/core/src/core_logic/ConfigurePatchingProcessor.py
```diff
@@ -47,16 +48,28 @@ def start_configure_patching(self):
...
+            # If the tracked operation is Configure patching, we cannot write a final status until assessment has also written a final status (mitigation for a CRP bug)
+            if self.execution_config.operation.lower() != Constants.CONFIGURE_PATCHING.lower():
+                self.set_configure_patching_final_overall_status()
         except Exception as error:
             self.current_auto_assessment_state = Constants.AutoAssessmentStates.ERROR
-            self.__report_consolidated_configure_patch_status(status=Constants.STATUS_ERROR, error=error)
+            self.configure_patching_exception_error = error
+            # If the tracked operation is Configure patching, we cannot write a final status until assessment has also written a final status (mitigation for a CRP bug)
+            if self.execution_config.operation != Constants.CONFIGURE_PATCHING.lower():
+                self.__report_consolidated_configure_patch_status(status=Constants.STATUS_ERROR, error=self.configure_patching_exception_error)
```
rane-rajasi, line 59: error=repr(self.configure_patching_exception_error) or convert to string?

kjohn-msft:
@rane-rajasi  -- leaving as a note here to holistically take in this PR:
image

src/core/src/CoreMain.py
kjohn-msft: Addressed with more safety checks.

src/core/src/core_logic/ConfigurePatchingProcessor.py
kjohn-msft: Yes, this covered, it's a coverage check failure on tooling.

kjohn-msft:
> @rane-rajasi -- leaving as a note here to holistically take in this PR: image

This was addressed too.

feng-j678 (approved):

src/core/src/CoreMain.py
```diff
@@ -74,10 +74,18 @@ def __init__(self, argv):
...
+            # Assessment happens for all operations. If the goal seeking tracked operation is CP, then its final status can only be written after assessment reaches a terminal state
+            assessment_exception_error = None
+            try:
                 patch_assessment_successful = patch_assessor.start_assessment()
+            except Exception as error:
+                assessment_exception_error = error  # hold this until configure patching is closed out
+
+            # close out configure patching if needed & then raise any 'uncontrolled' assessment exception if it occurred
+            if not execution_config.exec_auto_assess_only and patch_operation_requested == Constants.CONFIGURE_PATCHING.lower():
+                configure_patching_processor.set_configure_patching_final_overall_status()  # guarantee configure patching status write prior to throwing on any catastrophic assessment error
```
rane-rajasi, line 86: Won't CRP stop reading, if it gets this status update before it gets assessment error?

rane-rajasi (changes requested):
Left one comment

src/core/src/CoreMain.py
rane-rajasi: Even with retries, if all the re-attempts fail, there could still be the issue, where line 86 updates CP status, which is read by agent and hence CRP before line 88 executes, causing CRP to to think that is the terminal status, where CPSummary is Success but PatchAssessmentSummary is Transitioning

kjohn-msft: As discussed, covered in PatchAssessor.

rane-rajasi (approved):

PR #186, opened 2023-04-07
Bugfix/never reboot is selected change patch status to completed with warning

feng-j678:
bug fix:

- set patch status to warning when never reboot status is selected and there's pending


feng-j678:
draft pr

PR #188, opened 2023-04-13
Bugfix/never reboot is selected change patch status to warning

feng-j678:
- Never reboot is selected and there's reboot pending in reboot manager, set the patch status to warning


kjohn-msft (approved):

src/core/tests/Test_CoreMain.py
```diff
@@ -468,6 +469,54 @@ def test_install_all_packages_for_centos_autopatching(self):
 
         LegacyEnvLayerExtensions.LegacyPlatform.linux_distribution = backup_envlayer_platform_linux_distribution
 
+    def test_install_all_packages_for_centos_autopatching_warning(self):
```
rane-rajasi, line 472: Update test name and description to reflect your test case, i.e. test_install_for_centos_autopatching_with_reboot_never.

Test names should reflect what we are mainly validating, my assumption is you are majorly verifying PATCH_INSTALLATION_SUMMARY should have warning state on NeverReboot, it's good to have other asserts but the test name/description should have the main assert

src/core/tests/Test_CoreMain.py
```diff
@@ -479,6 +528,7 @@ def test_install_only_critical_and_security_packages_for_redhat_autopatching(sel
         classifications_to_include = ["Security", "Critical"]
         argument_composer.maintenance_run_id = str(maintenance_run_id)
         argument_composer.classifications_to_include = classifications_to_include
+        argument_composer.reboot_setting = 'Reboot'
```
rane-rajasi, line 531: Same comment as the other about reboot_setting value, please see the referenced resources there for a valid value to be used

src/core/tests/Test_CoreMain.py
```diff
@@ -522,6 +572,60 @@ def test_install_only_critical_and_security_packages_for_redhat_autopatching(sel
 
         LegacyEnvLayerExtensions.LegacyPlatform.linux_distribution = backup_envlayer_platform_linux_distribution
 
+    def test_install_only_critical_and_security_packages_for_redhat_autopatching_warning(self):
```
rane-rajasi, line 575: Same comment as the other on the test name/description reflecting the main validation or assert

src/core/tests/Test_PatchInstaller.py
```diff
@@ -177,16 +178,24 @@ def healthstore_writes_helper(self, health_store_id, maintenance_run_id, expecte
...
+            substatus_file_data = json.load(file_handle)[0]["status"]["substatus"]
+
+        if is_force_reboot:
+            self.assertEqual('warning', substatus_file_data[0]['status'])
+        else:
+            self.assertEqual('success', substatus_file_data[0]['status'])
 
-        self.assertEqual(True, json.loads(substatus_file_data['formattedMessage']['message'])['shouldReportToHealthStore'])
-        self.assertEqual(expected_patch_version, json.loads(substatus_file_data['formattedMessage']['message'])['patchVersion'])
+        self.assertEqual(True, json.loads(substatus_file_data[1]['formattedMessage']['message'])['shouldReportToHealthStore'])
```
rane-rajasi, line 197: Why did the index change here? What was added at 0th index?

src/core/tests/Test_RebootManager.py
```diff
@@ -65,6 +65,15 @@ def test_reboot_never(self):
         self.assertEqual(reboot_manager.start_reboot_if_required_and_time_available(20), False)
         runtime.stop()
 
+    def test_reboot_never_pending(self):
```
rane-rajasi, line 68: edit test name to: test_reboot_pending_with_reboot_setting_as_never()

rane-rajasi (changes requested):

src/core/tests/Test_PatchInstaller.py
feng-j678: this is helper func that used for 5 healthstore utests, and i want to reuse the set up. the substatus is a list of object and at index 0 it has the status field which needs to be test either warning or success or other state. if substatus_file_data is at index 1, then i will not able to get the status value

image

rane-rajasi: Did not quite understand this. From what I get here is: substatuses are PatchAssessmentSummary, PatchInstallationSummary, PatchMetadatForHealthStore and ConfigurePatchingSummary. There aren't any changes to these overall substatuses', just "status" field within PatchInstallationSummary. So if these substatuses already exists in the result, why would their index change with a change explicitly to the "status" field within "PatchInstallationSummary" substatus?

rane-rajasi: Adding to the comment above, the position (index) of these substatuses within substatus_file_data should not change because the helper function is not changing that

feng-j678: sorry that image was a bad example, here's the actual substatus_file result from the test:
image
as you can at index 0 is patch installation, if I keep it at index 1 then i am looking at the patchmetahealth section

rane-rajasi: The question is why did PatchMetadataFromHealthStore go from index 0 to 1? 

Your change in healthstore_writes_helper() set runtime.package_manager.force_reboot to True based on the incoming is_force_reboot function parameter. After which, in line 198-190, it is reading the contents of from status_file_path.
So there is no explicit file write in this code change (the write part happens inherently in the background, like before).

So to answer the question on why did PatchMetadataFromHealthStore go from index 0 to 1, we need to find these things:
1. Did the previous test (i.e. before this change) not have PatchInstallationSummary substatus at all? 
2. Or it did and was at a different index? In that case, why did the indices change? (ideally should not have )
3. If PatchInstallationSummary wasn't there before, why so and why is it added now?

feng-j678: comments and questions are resolved in our codebase meeting, it's okay to set index at 1 in assert()

rane-rajasi (approved):

kjohn-msft (approved):

PR #190, opened 2023-05-10
add complete status

feng-j678:
- add status complete code
- modify the write_status_file in each operations and writes <seq.no>.complete.status to status directory
- add code for latest complete status file, get file kb size, and remove outdated complete status file

 
image

### DIFF
--- a/src/core/src/core_logic/ExecutionConfig.py
+++ b/src/core/src/core_logic/ExecutionConfig.py
@@ -69,15 +69,20 @@ class ExecutionConfig(object):
 
         # Accommodation for bugs in higher-level components where 'Security' is being selected without selecting 'Critical' - should be rolled back no later than Jan 2022
         if self.included_classifications_list is not None and ('Security' in self.included_classifications_list and 'Critical' not in self.included_classifications_list):
-            self.composite_logger.log_debug("The included_classifications_list was corrected to include 'Critical' when 'Security' was specified.")
+            self.composite_logger.log_debug("The included_classifications_list was corrected to include 'Critical' "
+                                            "when 'Security' was specified.")
             self.included_classifications_list = ['Critical'] + self.included_classifications_list
 
         # Derived Settings
         self.log_file_path = os.path.join(self.log_folder, str(self.sequence_number) + ".core.log")
+        self.status_complete_file_path = os.path.join(self.status_folder, str(self.sequence_number)+".complete" + ".status")
         self.status_file_path = os.path.join(self.status_folder, str(self.sequence_number) + ".status")
         self.include_assessment_with_configure_patching = (self.operation == Constants.CONFIGURE_PATCHING and self.assessment_mode == Constants.AssessmentModes.AUTOMATIC_BY_PLATFORM)
-        self.composite_logger.log_debug(" - Derived execution-config settings. [CoreLog={0}][StatusFile={1}][IncludeAssessmentWithConfigurePatching={2}]"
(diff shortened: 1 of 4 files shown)


src/core/src/core_logic/ExecutionConfig.py
```diff
@@ -69,15 +69,20 @@ def __init__(self, env_layer, composite_logger, execution_parameters):
 
         # Accommodation for bugs in higher-level components where 'Security' is being selected without selecting 'Critical' - should be rolled back no later than Jan 2022
         if self.included_classifications_list is not None and ('Security' in self.included_classifications_list and 'Critical' not in self.included_classifications_list):
-            self.composite_logger.log_debug("The included_classifications_list was corrected to include 'Critical' when 'Security' was specified.")
+            self.composite_logger.log_debug("The included_classifications_list was corrected to include 'Critical' "
+                                            "when 'Security' was specified.")
```
kjohn-msft, line 73: Let's not do this -- we have more horizontal real-estate than vertical on our screens. We have a principle to avoid this in this code base except in very critical function calls.

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -310,8 +312,16 @@ def set_assessment_substatus_json(self, status=Constants.STATUS_TRANSITIONING, c
...
         self.__assessment_substatus_json = self.__new_substatus_json_for_operation(Constants.PATCH_ASSESSMENT_SUMMARY, status, code, json.dumps(self.__assessment_summary_json))
 
-        # Update status on disk
-        self.__write_status_file()
+        # Update status complete on disk
+        self.__write_status_complete_file()
+
+        # Write truncated status on disk, apply truncation logic here
+        status_json_payload = self.__read_status_file_raw_data(self.status_complete_file_path)
+        # print('what is status json_payload', status_json_payload)
```
kjohn-msft, line 320: Assuming because this is a work in progress, but don't have debug comments going into the code base.

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -310,8 +312,16 @@ def set_assessment_substatus_json(self, status=Constants.STATUS_TRANSITIONING, c
...
+        self.env_layer.file_system.write_with_retry_using_temp_file(self.status_file_path,
+                                                                    '[{0}]'.format(json.dumps(status_json_payload)),
+                                                                    mode='w+')
```
kjohn-msft, line 323: Not leaving the same comment everywhere but avoid multi-lines where possible. You can stop the IDE from prompting you on this. The IDE prompts are based on being good to folk who are trying to edit code in the terminal later (this does not apply to us).

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -783,3 +807,37 @@ def __set_errors_json(self, error_count_by_operation, errors_by_operation):
...
+
+        return latest_file
+
+    def __get_latest_file_kb_size(self, latest_file):
+        """ Get the file kb size in two decimal places of the latest status complete file"""
+        if latest_file is None:
+            raise Exception("Error: File is not found.")
+        file_kb_size = round(os.path.getsize(latest_file) / 1024, 2)
+        return file_kb_size
+    # endregion
```
kjohn-msft, line 843: Files must always end with a new line. You'll get flagged in the IDE and in Github for this. (you can see it here).

kjohn-msft (changes requested):
Please have @rane-rajasi go through as primary. I've left some comments for general code expectations.

PR #192, opened 2023-06-06
add installation truncation

feng-j678:
What is in this PR:
[X] Compressed assessment and installation patches
[X] Add tombstone record for compressed assessment per classifications.
[X] Add tombstone record for compressed installation.
[X] Add error details for compressed assessment.
[X] Add error details for compressed installation.
[X] set status as warning when compression applied.
[X] Clean up multi-line formatting.
[X] Test assessment and installation compression.
[X] Unit test coverage...
[x] delay truncation logic - add condition to check current time vs truncation timestamp if 60 secs have passed. when timestamp is below 60 secs, truncation logic will not apply during application process unless the terminal state has reached (success or error) 

_____________________________
image
image
image
image


src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -692,12 +717,15 @@ def __write_complete_status_file(self):
         if os.path.isdir(self.complete_status_file_path):
             self.composite_logger.log_error("Core state file path returned a directory. Attempting to reset.")
             shutil.rmtree(self.complete_status_file_path)
+            shutil.rmtree(self.complete_status_file_path)
```
rane-rajasi, line 720: Duplicate line.

src/core/src/CoreMain.py
```diff
@@ -101,6 +101,7 @@ def __init__(self, argv):
                     patch_installer.mark_installation_completed()
                     overall_patch_installation_operation_successful = True
                 self.update_patch_substatus_if_pending(patch_operation_requested, overall_patch_installation_operation_successful, patch_assessment_successful, configure_patching_successful, status_handler, composite_logger)
+            status_handler.log_truncated_removed_packages_if_any()
```
rane-rajasi, line 105: Status truncation should happen at every status file write and not just at the end of a patch operation. Reason for this as we have discussed before is, the agent reads status file at regular intervals and reports it back to CRP. If the status file exceeds limit during any of these times, agent will truncate it and report the broken json

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -855,26 +857,27 @@ def log_truncated_removed_packages_if_any(self):
...
         """ Truncate the substatus summary patch list when complete status file size is more than 126kb """
         """
             __write_status_file(self, complete_status_file_payload):
-                status_file_size = __get_byte_size(complete_status_file_payload)
+                status_file_size = __calc_status_size_on_disk(complete_status_file_payload)
                 if status_file_size > 126kb
                     _index, _name = self.__get_index_name()
-                    empty_list_file_size = __get_twice_byte_size(__size_of_constant_status_data(complete_status_file_payload))
-                    packages_size_limit = 126kb - empty_list_file_size
+                    status_file_no_list_size = __calc_package_payload_size_on_disk(__get_size_of_constant_status_data(complete_status_file_payload))
```
rane-rajasi, line 868: rename status_file_no_list_size to status_file_without_package_list_size or status_file_size_minus_package_list_size

src/core/src/CoreMain.py
```diff
@@ -101,6 +101,7 @@ def __init__(self, argv):
...
```
rane-rajasi, line 104: rename to log_truncated_packages(). Truncated and removed indicate the same thing and the function should log "No packages truncated" if there are no packages truncated.

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -404,20 +425,29 @@ def __new_installation_summary_json(self, installation_packages_json):
         self.__refresh_installation_reboot_status()
 
         # Compose substatus message
+        maintenance_run_id = self.execution_config.maintenance_run_id if self.execution_config.maintenance_run_id is not None else ''
+        errors = self.__set_errors_json(self.__installation_total_error_count, self.__installation_errors)
+        substatus_message = self.__compose_installation_substatus_msg(self.execution_config.activity_id, self.__installation_reboot_status, self.__maintenance_window_exceeded, not_selected_patch_count,
```
rane-rajasi, line 430: Same comment as assessment about using named arguments

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -404,20 +425,29 @@ def __new_installation_summary_json(self, installation_packages_json):
...
+            "rebootStatus": str(reboot_status),
+            "maintenanceWindowExceeded": maintenance_window,
+            "notSelectedPatchCount": not_selected,
+            "excludedPatchCount": excluded,
+            "pendingPatchCount": pending,
+            "installedPatchCount": installed,
+            "failedPatchCount": failed,
+            "patches": installation_packages,
+            "startTime": str(start_time),
+            "lastModifiedTime": str(last_modified_time),
```
rane-rajasi, line 448: Check and confirm what happens if last_modified_time passed in to this function is already a str

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -693,11 +734,14 @@ def __write_complete_status_file(self):
             self.composite_logger.log_error("Core state file path returned a directory. Attempting to reset.")
             shutil.rmtree(self.complete_status_file_path)
 
-        # Write complete status file <seq.no>.complete.status
-        self.env_layer.file_system.write_with_retry_using_temp_file(self.complete_status_file_path, '[{0}]'.format(json.dumps(complete_status_payload)), mode='w+')
+        # Write complete status file <seq.no>.complete
```
rane-rajasi, line 737: <seq.no>.complete.status

to avoid any confusions later

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -693,11 +734,14 @@ def __write_complete_status_file(self):
...
+        complete_status_payload_json_dumps = json.dumps(complete_status_payload)
+        self.env_layer.file_system.write_with_retry_using_temp_file(self.complete_status_file_path, '[{0}]'.format(complete_status_payload_json_dumps), mode='w+')
+
+        # complete_status_file = self.__load_complete_status_file_data(self.complete_status_file_path)    # avoid reference modification
```
rane-rajasi, line 741: remove this comment

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -795,14 +835,402 @@ def __try_add_error(error_list, detail):
...
+        if truncated:
+            error_count += 1    # add 1 because of truncation
+            code = Constants.PatchOperationTopLevelErrorCode.WARNING if code != error_code else error_code
+
+        message = "{0} error/s reported.".format(error_count)
+        message += " The latest {0} error/s are shared in detail. To view all errors, review this log file on the machine: {1}".format(len(errors_by_operation), self.__log_file_path) if error_count > 0 else ""
         return {
-            "code": Constants.PatchOperationTopLevelErrorCode.SUCCESS if error_count_by_operation == 0 else Constants.PatchOperationTopLevelErrorCode.ERROR,
+            "code": code,
             "details": errors_by_operation,
```
rane-rajasi, line 854: Are we passing the truncated error message also in here?

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -795,14 +835,402 @@ def __try_add_error(error_list, detail):
...
+                                    
+                            current_package_list_size -= diff
+                            if current_package_list_size <= max_package_list_capacity:
+                                break
+                            __get_new_size_in_bytes_after_truncation() 
+                    
+                write_with_retry_using_temp_file()  
+        """
+        status_file_size_in_bytes = len(complete_status_file_payload.encode("utf-8"))   # calc complete_status_file_payload byte size on disk
+        truncated_status_file = json.loads(complete_status_file_payload)    # reload payload into python object
```
rane-rajasi, line 924: do not name this as truncated_status_file. At this moment, it is not truncated

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -795,14 +835,402 @@ def __try_add_error(error_list, detail):
...
+            packages_capacity_diff = 0
+            assessment_substatus_index = self.__get_substatus_index(Constants.PATCH_ASSESSMENT_SUMMARY, truncated_status_file['status']['substatus'])
+            installation_substatus_index = self.__get_substatus_index(Constants.PATCH_INSTALLATION_SUMMARY, truncated_status_file['status']['substatus'])
+
+            if assessment_substatus_index is not None:        # if assessment data exists
+                self.__assessment_substatus_msg_copy = self.__get_substatus_msg_at_index(truncated_status_file, assessment_substatus_index)
+                self.__assessment_packages_copy = self.__assessment_substatus_msg_copy['patches']
+
+            if installation_substatus_index is not None:      # if installation data exists
+                low_pri_index = self.__get_installation_low_pri_index(self.__installation_packages_copy, low_pri_index)
```
rane-rajasi, line 938: shouldn't this line come after line 940? i.e. when self.__installation_packages_copy is initialized

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -795,14 +835,402 @@ def __try_add_error(error_list, detail):
...
+
+        return truncated_assessment_list, packages_removed_assessment, truncated_installation_list, packages_removed_installation
+
+    def __get_installation_low_pri_index(self, priority_sorted_installation_packages, low_pri_index):
+        """" Get the first index of Pending, Excluded, or Not_Selected installation packages """
+        for index, package in enumerate(priority_sorted_installation_packages):
+            package_state = package['patchInstallationState']
+            if Constants.PENDING in package_state or Constants.EXCLUDED in package_state or Constants.NOT_SELECTED in package_state:
+                low_pri_index = index
+                return low_pri_index
```
rane-rajasi, line 1037: very minor: since we are returning low_pri_index at the end of the function, we can simply break out of the loop here and let line 1038 return low_pri_index

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -795,14 +835,402 @@ def __try_add_error(error_list, detail):
...
+            if assessment_substatus_index is not None:        # if assessment data exists
+                self.__assessment_substatus_msg_copy = self.__get_substatus_msg_at_index(truncated_status_file, assessment_substatus_index)
+                self.__assessment_packages_copy = self.__assessment_substatus_msg_copy['patches']
+
+            if installation_substatus_index is not None:      # if installation data exists
+                low_pri_index = self.__get_installation_low_pri_index(self.__installation_packages_copy, low_pri_index)
+                self.__installation_substatus_msg_copy = self.__get_substatus_msg_at_index(truncated_status_file, installation_substatus_index)
+                self.__installation_packages_copy = self.__installation_substatus_msg_copy['patches']
+
+            status_file_without_package_list_size = self.__get_size_of_constant_status_data(copy.deepcopy(truncated_status_file), assessment_substatus_index, installation_substatus_index)  # deepcopy fully copy the object avoid reference modification
```
rane-rajasi, line 942: rename to size_of_constant_status_data

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -795,14 +835,402 @@ def __try_add_error(error_list, detail):
...
+        return len(json.dumps(full_status).encode("utf-8"))
+
+    def __calc_package_payload_size_on_disk(self, package_list):
+        """ Calculate final package list size in bytes (because of escape chars) """
+        first_json_dump = json.dumps(package_list)
+        return len(json.dumps(first_json_dump).encode("utf-8"))
+
+    def __get_size_of_constant_status_data(self, complete_status_file_payload, assessment_status_index, installation_status_index):
+        """ Get the size in bytes of the complete_status_file without package data  """
+        status_file_no_list_data = complete_status_file_payload
```
rane-rajasi, line 1098: If complete_status_file_payload is a deep copy from the caller, what is the need to assign it to a new variable here?

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -795,14 +835,402 @@ def __try_add_error(error_list, detail):
...
+        error_msg = Constants.StatusTruncationConfig.TRUNCATION_WARNING_MESSAGE
+        truncated_error_detail = self.__set_error_detail(Constants.PatchOperationErrorCodes.TRUNCATION, error_msg)      # Reuse the errors object set up
+        self.__try_add_error(truncation_detail_list, truncated_error_detail)
+        truncated_errors_json = self.__set_errors_json(count_total_errors, truncation_detail_list, True)    # True for truncated
+
+        return truncated_errors_json
+
+    def __recompose_assessment_message(self, assessment_substatus_msg, assessment_packages, assessment_errors):
+        """ Recompose assessment substatus message """
+        activity_id = assessment_substatus_msg['assessmentActivityId']
```
rane-rajasi, line 1167: This is introducing a dependency on the substatus keys/field name (i.e. assessmentActivityId, rebootPending, startTime, etc). Any changes to these field names in future will need changes in this function and in __compose_assessment_substatus_msg()

Why not reuse __compose_assessment_substatus_msg() passing an empty package list to compute assessment substatus without packages instead?

rane-rajasi: Just saw you are using __compose_assessment_substatus_msg below. 

The concern on field name duplication still exists. Suggestion: we could add all of these field names in an enum in Constants and use the reference from those in here.

src/core/src/service_interfaces/StatusHandler.py
rane-rajasi: Use the name reformatted_status_file instead of complete_status_file_payload or status_file_no_list_data

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -795,14 +835,402 @@ def __try_add_error(error_list, detail):
...
+                self.__assessment_substatus_msg_copy = self.__get_substatus_msg_at_index(truncated_status_file, assessment_substatus_index)
+                self.__assessment_packages_copy = self.__assessment_substatus_msg_copy['patches']
+
+            if installation_substatus_index is not None:      # if installation data exists
+                low_pri_index = self.__get_installation_low_pri_index(self.__installation_packages_copy, low_pri_index)
+                self.__installation_substatus_msg_copy = self.__get_substatus_msg_at_index(truncated_status_file, installation_substatus_index)
+                self.__installation_packages_copy = self.__installation_substatus_msg_copy['patches']
+
+            status_file_without_package_list_size = self.__get_size_of_constant_status_data(copy.deepcopy(truncated_status_file), assessment_substatus_index, installation_substatus_index)  # deepcopy fully copy the object avoid reference modification
+            max_package_list_capacity = self.__internal_file_capacity - status_file_without_package_list_size
```
rane-rajasi, line 943: rename to size_of_max_allowed_packages

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -795,14 +835,402 @@ def __try_add_error(error_list, detail):
...
+            current_package_list_size = self.__get_current_packages_size_in_bytes(self.__assessment_packages_copy, self.__installation_packages_copy)
```
rane-rajasi, line 944: rename to size_of_dumped_package_lists

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -795,14 +835,402 @@ def __try_add_error(error_list, detail):
...
+
+            while status_file_size_in_bytes > self.__internal_file_capacity and current_package_list_size > max_package_list_capacity:
```
rane-rajasi, line 946: We don't need status_file_size_in_bytes > self.__internal_file_capacity, since:
1. current_package_list_size is always recomputed
2. We have a fixed value for max_package_list_capacity
So as long as current_package_list_size > max_package_list_capacity, this loop will continue

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -795,14 +835,402 @@ def __try_add_error(error_list, detail):
...
+                # Start truncation process
+                packages_truncated_in_assessment, packages_removed_from_assessment, packages_truncated_in_installation, packages_removed_from_installation = \
```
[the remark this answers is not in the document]
rane-rajasi: why are they initialized with '\'?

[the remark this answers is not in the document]
rane-rajasi: Q: Why are you not using self.__assessment_packages_removed directly instead of packages_removed_from_assessment?

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -795,14 +835,402 @@ def __try_add_error(error_list, detail):
...
+                    self.__apply_truncation_process(self.__assessment_packages_copy, self.__installation_packages_copy, max_package_list_capacity - packages_capacity_diff, low_pri_index)
```
rane-rajasi, line 949: What is packages_capacity_diff supposed to be? It is 0 at this point in the first run of while loop

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -795,14 +835,402 @@ def __try_add_error(error_list, detail):
...
+
+        # Only assessment data exists, get assessment list capacity
+        if not installation_package_list:
+            return self.__calc_package_payload_size_on_disk(assessment_package_list)
+
+        # Both assessment and installation data exist
+        return self.__calc_package_payload_size_on_disk(assessment_package_list) + self.__calc_package_payload_size_on_disk(installation_package_list)
+
+    def __cut_list_and_get_list_capacity(self, package_list, max_package_list_capacity):
+        """ Split package list, keep 5 minimum packages, and remaining packages for truncation """
```
rane-rajasi, line 996: Why are we calculating sizes here? IIUC, the main aim of this function is split assessment list into 2 parts, where the 1 part contains 5 packages and the 2nd one contains remaining.

This function is not performing any truncation related activity, simply dividing a list into 2. Let's keep it to that and keep all size computations restricted to the truncation function

rane-rajasi (changes requested):
Comments inline.

src/core/src/CoreMain.py
feng-j678: done

src/core/src/bootstrap/Constants.py
```diff
@@ -168,6 +168,13 @@ class AutoAssessmentStates(EnumBackport):
     STATUS_SUCCESS = "Success"
     STATUS_WARNING = "Warning"
 
+    # Status file size
+    class StatusTruncationConfig(EnumBackport):
+        INTERNAL_FILE_SIZE_LIMIT_IN_BYTES = 126 * 1024
+        AGENT_STATUS_FILE_SIZE_LIMIT_IN_BYTES = 128 * 1024
```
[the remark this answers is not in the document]
feng-j678: done

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -343,24 +359,29 @@ def __new_assessment_summary_json(self, assessment_packages_json, status, code):
...
-            "patches": assessment_packages_json,
-            "startTime": str(self.execution_config.start_time),
-            "lastModifiedTime": str(self.env_layer.datetime.timestamp()),
-            "startedBy": str(started_by),
-            "errors": self.__set_errors_json(self.__assessment_total_error_count, self.__assessment_errors)
-        }
+        # Compose substatus message
+        errors = self.__set_errors_json(self.__assessment_total_error_count, self.__assessment_errors)
+        substatus_message = self.__compose_assessment_substatus_msg(self.execution_config.activity_id, self.is_reboot_pending, critsec_patch_count, other_patch_count,
+        assessment_packages_json, self.execution_config.start_time, self.env_layer.datetime.timestamp(), started_by, errors)
```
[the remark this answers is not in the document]
feng-j678: done

src/core/src/service_interfaces/StatusHandler.py
feng-j678: done

src/core/src/service_interfaces/StatusHandler.py
feng-j678: both status files yields the same result:
image
image

src/core/src/service_interfaces/StatusHandler.py
feng-j678: done

src/core/src/service_interfaces/StatusHandler.py
feng-j678: we are passing the new truncated_detail object into the errors json if error code is not error, if it is error then take the existing details object and add code and msg
image

src/core/src/service_interfaces/StatusHandler.py
feng-j678: rewrote this logic to write_complete_status_file()

src/core/src/service_interfaces/StatusHandler.py
feng-j678: yes, you're right, good catch on this. this will perform the full installation truncation until a follow up operation (like status change) then will use the initialized __installation_packages_copy for high and low pri truncation

src/core/src/service_interfaces/StatusHandler.py
feng-j678: return index instead since we are assigning to low_pri_index line 953.

src/core/src/service_interfaces/StatusHandler.py
feng-j678: done

src/core/src/service_interfaces/StatusHandler.py
feng-j678: it was done per koshy's suggestion to help code readability, but

src/core/src/service_interfaces/StatusHandler.py
feng-j678: done

src/core/src/service_interfaces/StatusHandler.py
feng-j678: done

src/core/src/service_interfaces/StatusHandler.py
feng-j678: logic is removed

src/core/src/service_interfaces/StatusHandler.py
feng-j678: logic move to __apply_truncation_process()

src/core/src/service_interfaces/StatusHandler.py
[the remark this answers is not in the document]
feng-j678: > nit: use retained instead of truncated
> 
> rename the variables to assessment_packages_retained_after_truncation, assessment_packages_removed_after_truncation, and similarly for installation lists

done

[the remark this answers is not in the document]
feng-j678: > why are they initialized with '\'?

\ means newline, there are too many characters in that line

src/core/src/service_interfaces/StatusHandler.py
feng-j678: the packages_capacity_diff is the hidden byte of escape chars in the package list, for example if max package list 100byte data, after final json.dump that 100byte data will become 110 byte, so in reality our max package list should be 90byte to account for escape chars. those extra byes vary by # of elements in the list, length of elements, so i need a variable to keep track of those escape chars then do subtraction to get the real max package list byte size

src/core/src/service_interfaces/StatusHandler.py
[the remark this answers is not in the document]
feng-j678: > Q: Why are you not using self.__assessment_packages_removed directly instead of packages_removed_from_assessment?
it's a preference, if we have any values for packages_removed_from_assessment then i won't create a truncated status file and not initialize __assessment_packages_removed, then it will pick up by gc. using __assessment_packages_removed instead, then i need to initialize it to packages_removed_from_assessment, so it won't get pick up by gc

src/core/src/CoreMain.py
```diff
@@ -101,6 +101,7 @@ def __init__(self, argv):
                     patch_installer.mark_installation_completed()
                     overall_patch_installation_operation_successful = True
                 self.update_patch_substatus_if_pending(patch_operation_requested, overall_patch_installation_operation_successful, patch_assessment_successful, configure_patching_successful, status_handler, composite_logger)
+            status_handler.log_truncated_packages()
```
rane-rajasi, line 105: q: Wouldn't truncated packages be logged at the end of each status file write?

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -343,24 +358,34 @@ def __new_assessment_summary_json(self, assessment_packages_json, status, code):
...
-            "patches": assessment_packages_json,
-            "startTime": str(self.execution_config.start_time),
-            "lastModifiedTime": str(self.env_layer.datetime.timestamp()),
-            "startedBy": str(started_by),
-            "errors": self.__set_errors_json(self.__assessment_total_error_count, self.__assessment_errors)
-        }
+        # Compose substatus message
+        errors = self.__set_errors_json(self.__assessment_total_error_count, self.__assessment_errors)
+        # substatus_message = self.__compose_assessment_substatus_msg(self.execution_config.activity_id, self.is_reboot_pending, critsec_patch_count, other_patch_count,
+        # assessment_packages_json, self.execution_config.start_time, self.env_layer.datetime.timestamp(), started_by, errors)
```
rane-rajasi, line 364: remove commented lines

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -404,20 +429,34 @@ def __new_installation_summary_json(self, installation_packages_json):
         self.__refresh_installation_reboot_status()
 
         # Compose substatus message
+        maintenance_run_id = self.execution_config.maintenance_run_id if self.execution_config.maintenance_run_id is not None else ''
+        errors = self.__set_errors_json(self.__installation_total_error_count, self.__installation_errors)
+        # substatus_message = self.__compose_installation_substatus_msg(self.execution_config.activity_id, self.__installation_reboot_status, self.__maintenance_window_exceeded, not_selected_patch_count,
+        #     excluded_patch_count, pending_patch_count, installed_patch_count, failed_patch_count, installation_packages_json, self.execution_config.start_time,
+        #     self.env_layer.datetime.timestamp(), maintenance_run_id, errors)
```
rane-rajasi, line 436: remove comments

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -614,32 +662,34 @@ def load_status_file_components(self, initial_load=False):
...
+        """ Get the substatus payload message by index """
+        return json.loads(status_file_data['status']['substatus'][substatus_index]['formattedMessage']['message'])
+
     def __load_complete_status_file_data(self, file_path):
-        # Read the status file - raise exception on persistent failure
         for i in range(0, Constants.MAX_FILE_OPERATION_RETRY_COUNT):
             try:
                 with self.env_layer.file_system.open(file_path, 'r') as file_handle:
-                    complete_status_file_data = json.load(file_handle)[0]    # structure is array of 1
+                    complete_status_file_file_data = json.load(file_handle)[0]  # structure is array of 1
```
rane-rajasi, line 684: 'file' repeated twice in complete_status_file_file_data

src/core/src/CoreMain.py
rane-rajasi: @feng-j678 pending

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -795,14 +842,348 @@ def __try_add_error(error_list, detail):
...
+                        __recompose_truncated_status_file()                                    
+                            __get_current_complete_status_errors()
+                            __recompose_truncated_substatus_msg()
+                                __recompose_substatus_msg_errors()
+                                __create_installation_tombstone
+                                __recreate_installation_summary_json()
+                                
+                        status_file_size_in_bytes, status_file_agent_size_diff = __get_new_size_in_bytes_after_truncation(truncated_status_file)
+                        size_of_max_packages_allowed_in_status -= status_file_agent_size_diff  
+        """
```
rane-rajasi, line 925: nit: note to remove this commented section before PR merge

src/core/src/service_interfaces/StatusHandler.py
rane-rajasi: @feng-j678 this is still pending

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -795,14 +842,348 @@ def __try_add_error(error_list, detail):
...
+        """ Calculate final package list size in bytes (because of escape chars) """
+        first_json_dump = json.dumps(package_list)
+        return len(json.dumps(first_json_dump).encode("utf-8"))
+
+    def size_of_constant_status_data(self, complete_status_file_payload, assessment_status_index, installation_status_index):
+        """ Get the size in bytes of the complete_status_file without package data  """
+        status_file_no_list_data = complete_status_file_payload
+        if assessment_status_index is not None:
+            assessment_substatus_msg_copy = self.__get_substatus_msg_at_index(status_file_no_list_data, assessment_status_index)
+            assessment_msg_empty_list = self.__update_substatus_msg_packages_and_errors(assessment_substatus_msg_copy, [], assessment_substatus_msg_copy['errors'])
```
rane-rajasi, line 1078: What is the need to update errors in the substatus? Since the source (i.e. assessment_substatus_msg_copy['errors']) is the same (i.e. assessment_substatus_msg_copy), errors should already be a part of it

rane-rajasi, line 1078: rename assessment_msg_empty_list to assessment_msg_empty_patches_list or assessment_msg_without_packges. Please note: follow the same convention throughout. There was a variable name with 'without_packages' used somewhere earlier. If that's the case, then use 'without packages' for all other variables wherever relevant

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -795,14 +842,348 @@ def __try_add_error(error_list, detail):
...
+
+        if installation_substatus_index is not None:      # if installation data exists
+            self.__installation_substatus_msg_copy = self.__get_substatus_msg_at_index(truncated_status_file, installation_substatus_index)
+            self.__installation_packages_copy = self.__installation_substatus_msg_copy['patches']
+            low_pri_index = self.__get_installation_low_pri_index(self.__installation_packages_copy)
+
+        status_file_without_package_list_size = self.size_of_constant_status_data(copy.deepcopy(truncated_status_file), assessment_substatus_index, installation_substatus_index)  # deepcopy fully copy the object avoid reference modification
+        size_of_max_packages_allowed_in_status = self.__internal_file_capacity - status_file_without_package_list_size
+
+        while status_file_size_in_bytes > self.__internal_file_capacity:
```
rane-rajasi, line 944: status_file_size_in_bytes > self.__internal_file_capacity:
this condition is already checked in load_complete_status_file() which calls __create_truncated_status_file()

The part we need to check here is is we need to remove more patches from status. I remember an older comment in one of the previous iterations, on this PR, which had 2 conditions, wherein I had recommended to only keep the 2nd check and remove this one. Please re-surface that comment or make that change if it is agreeable

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -795,14 +842,348 @@ def __try_add_error(error_list, detail):
...
+            # Start truncation process
+            packages_retained_in_assessment, assessment_packages_removed_after_truncation, packages_retained_in_installation, installation_packages_removed_after_truncation = \
```
rane-rajasi, line 946: keep naming structure similar. For eg: if using 'packages_retained_in_assessment' the other variable should be 'packages_removed_from_assessment'.

src/core/src/service_interfaces/StatusHandler.py
rane-rajasi: @feng-j678 the current_package_list_size is computed from status_file_size_in_bytes. So ensuring current_package_list_size is within the expected limit, also ensures status_file_size is being honored.

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -795,14 +842,348 @@ def __try_add_error(error_list, detail):
...
+    def __get_new_size_in_bytes_after_truncation(self, truncated_status_file):
+        """ Get new size in bytes for status_file,and difference of status file - 126kb """
+        status_file_size_in_bytes = self.__calc_status_size_on_disk(truncated_status_file)
+        status_file_agent_size_diff = status_file_size_in_bytes - self.__internal_file_capacity
+
+        return status_file_size_in_bytes, status_file_agent_size_diff
+
+    def __split_assessment_list(self, assessment_packages):
+        """ Split package list, keep 5 minimum packages, and remaining packages for truncation """
+        min_packages_count = Constants.StatusTruncationConfig.MIN_TRUNCATED_PACKAGE_COUNT
```
rane-rajasi, line 976: The constant StatusTruncationConfig.MIN_TRUNCATED_PACKAGE_COUNT is used exclusively for assessment so the name should reflect that. Please rename the constant to StatusTruncationConfig.MIN_ASSESSMENT_TRUNCATED_PACKAGE_COUNT

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -795,14 +842,348 @@ def __try_add_error(error_list, detail):
...
+        assessment_keep_five, assessment_packages_for_truncate = (assessment_packages[:min_packages_count], assessment_packages[min_packages_count:]) \
```
rane-rajasi, line 977: rename assessment_keep_five to min_assessment_patches_to_retain and assessment_packages_for_truncate to assessment_patches_eligible_for_truncation

important note, never use a constant value in variable name, for eg, assessment_keep_five, the min assessment patch to retain count can change anytime

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -795,14 +842,348 @@ def __try_add_error(error_list, detail):
...
+            if len(assessment_packages) > min_packages_count else (assessment_packages, [])
+
+        return assessment_keep_five, assessment_packages_for_truncate
+
+    def __apply_truncation_process(self, assessment_packages, installation_packages, max_package_list_capacity, low_pri_index=None):
+        """ Truncation function call split assessment method and apply truncation on assessment and installation packages """
+        installation_low_pri = []
+        installation_high_pri = installation_packages
+        # cut assessment list into [:5], [5:]
+        assessment_keep_five, assessment_packages_for_truncate = self.__split_assessment_list(assessment_packages)
```
rane-rajasi, line 987: same comment on not using a constant value in variable names. Please rename both variables according to 
image

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -795,14 +842,348 @@ def __try_add_error(error_list, detail):
...
+
+        # Empty list after 2xjson.dumps have 4-5 bytes, might cause to lose 1 extra package in truncation process
```
rane-rajasi, line 989: need more clarity in this comment

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -795,14 +842,348 @@ def __try_add_error(error_list, detail):
...
+        if len(assessment_keep_five) > 0:
+            max_package_list_capacity = max_package_list_capacity - self.__calc_package_payload_size_on_disk(assessment_keep_five)
+
+        # Apply high priority (Failed, Installed) and low priority (Pending, Excluded, Not_Selected) installation logic, and keep min 5 assessment packages
+        if low_pri_index:
+            installation_high_pri = installation_packages[:low_pri_index]
+            installation_low_pri = installation_packages[low_pri_index:]
+
+        installation_high_pri_list, packages_removed_high_pri, remain_list_capacity = self.__apply_truncation(installation_high_pri, max_package_list_capacity)
+        remain_assessment, packages_removed_assessment, remain_list_capacity = self.__apply_truncation(assessment_packages_for_truncate, remain_list_capacity)
```
rane-rajasi, line 999: rename remain_assessment to assessment_packages_retained

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -795,14 +842,348 @@ def __try_add_error(error_list, detail):
...
+        # Empty list after 2xjson.dumps have 4-5 bytes, might cause to lose 1 extra package in truncation process
+        if len(assessment_keep_five) > 0:
+            max_package_list_capacity = max_package_list_capacity - self.__calc_package_payload_size_on_disk(assessment_keep_five)
+
+        # Apply high priority (Failed, Installed) and low priority (Pending, Excluded, Not_Selected) installation logic, and keep min 5 assessment packages
+        if low_pri_index:
+            installation_high_pri = installation_packages[:low_pri_index]
+            installation_low_pri = installation_packages[low_pri_index:]
+
+        installation_high_pri_list, packages_removed_high_pri, remain_list_capacity = self.__apply_truncation(installation_high_pri, max_package_list_capacity)
```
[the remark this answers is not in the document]
rane-rajasi: rename installation_high_pri_list to packages_retained_in_install_high_pri

Similarly for the low pri list

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -795,14 +842,348 @@ def __try_add_error(error_list, detail):
...
+
+        while status_file_size_in_bytes > self.__internal_file_capacity:
+            # Start truncation process
+            packages_retained_in_assessment, assessment_packages_removed_after_truncation, packages_retained_in_installation, installation_packages_removed_after_truncation = \
+                self.__apply_truncation_process(self.__assessment_packages_copy, self.__installation_packages_copy, size_of_max_packages_allowed_in_status, low_pri_index)
+            if len(assessment_packages_removed_after_truncation) > 0:
+                # Update current assessment # of removed packages
+                self.__assessment_packages_removed = assessment_packages_removed_after_truncation
+                # Recompose truncated status file payload (assessment)
+                truncated_status_file = self.__recompose_truncated_status_file(truncated_status_file, packages_retained_in_assessment, self.__assessment_total_error_count,
```
rane-rajasi, line 952: use named parameters

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -795,14 +842,348 @@ def __try_add_error(error_list, detail):
...
+                return index
+        return None
+
+    def __recompose_truncated_status_file(self, truncated_status_file, truncated_package_list, count_total_errors, substatus_json, substatus_name, substatus_index):
+        """ Recompose final truncated status file version """
+        truncated_detail_list = []
+        error_code = Constants.PatchOperationTopLevelErrorCode.ERROR
+        code, errors_details = self.__get_current_complete_status_errors(substatus_json)
+
+        # Check for existing assessment errors before recompose
```
rane-rajasi, line 1101: Check for existing substatus errors before recomposing

This is not specific to assessment now

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -795,14 +842,348 @@ def __try_add_error(error_list, detail):
...
+        if code != error_code:
+            truncated_status_file['status']['substatus'][substatus_index]['status'] = Constants.STATUS_WARNING.lower()      # Update substatus status to warning
+        else:
+            # code == 1 (Error), add everything in the errors['details'] to truncated_detail_list
```
rane-rajasi, line 1105: remove comment

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -795,14 +842,348 @@ def __try_add_error(error_list, detail):
...
+            truncated_detail_list.extend(errors_details)
+
+        return self.__recompose_truncated_substatus_msg(truncated_status_file, truncated_package_list, truncated_detail_list,
```
rane-rajasi, line 1108: use named parameters

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -795,14 +842,348 @@ def __try_add_error(error_list, detail):
...
+        return self.__calc_status_size_on_disk(status_file_no_list_data)
+
+    def __get_substatus_index(self, substatus_name, substatus):
+        """" Get substatus index from the current substatus """
+        for index, sub_name in enumerate(substatus):
+            if sub_name['name'] == substatus_name:
+                return index
+        return None
+
+    def __recompose_truncated_status_file(self, truncated_status_file, truncated_package_list, count_total_errors, substatus_json, substatus_name, substatus_index):
```
rane-rajasi, line 1095: This substatus recomposing implementation (this and all it's successive functions) seems too extensive and can easily lead to errors in future. Think on simplifying this logic

rane-rajasi (changes requested):
The truncation implementation seems to have again diverted a bit from our pseudo code. Please ensure we keep that as a base.

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -795,14 +842,348 @@ def __try_add_error(error_list, detail):
...
+    def size_of_constant_status_data(self, complete_status_file_payload, assessment_status_index, installation_status_index):
+        """ Get the size in bytes of the complete_status_file without package data  """
+        status_file_no_list_data = complete_status_file_payload
+        if assessment_status_index is not None:
+            assessment_substatus_msg_copy = self.__get_substatus_msg_at_index(status_file_no_list_data, assessment_status_index)
+            assessment_msg_empty_list = self.__update_substatus_msg_packages_and_errors(assessment_substatus_msg_copy, [], assessment_substatus_msg_copy['errors'])
+            status_file_no_list_data['status']['substatus'][assessment_status_index]['formattedMessage']['message'] = json.dumps(assessment_msg_empty_list)
+
+        if installation_status_index is not None:
+            installation_substatus_msg_copy = self.__get_substatus_msg_at_index(status_file_no_list_data, installation_status_index)
```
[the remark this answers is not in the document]
feng-j678: use the global var _substatus_msg_copy for storing the substatus msg, reduce repetition

src/core/src/service_interfaces/StatusHandler.py
feng-j678: good catch on this, I meant to move this comment to here. if I don't have that logic then the code highlighted in red will be executed and removing 2-4bytes from capacity, sometimes this possibly truncated 1 more package.

image

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -795,14 +842,348 @@ def __try_add_error(error_list, detail):
...
+        """
+        left_index = 0
+        right_index = len(package_list) - 1
+
+        if len(package_list) == 0:
+            return [], [], capacity
+        if self.__calc_package_payload_size_on_disk(package_list) <= capacity:
+            return package_list, [], capacity - self.__calc_package_payload_size_on_disk(package_list)
+        if self.__calc_package_payload_size_on_disk(package_list[0]) > capacity:
+            return [], package_list, capacity
```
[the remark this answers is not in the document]
feng-j678: did a code walk through meeting to cover this logic, done

src/core/src/service_interfaces/StatusHandler.py
feng-j678: comment removed

src/core/src/service_interfaces/StatusHandler.py
feng-j678: refactor the code in  __recompose_truncated_status_file(). done

src/core/src/CoreMain.py
rane-rajasi: My opinion, this should be added for each status file write, post truncation, because that gives the details of what happened in each truncation (and thus status file writes). Will help in debugging issues in future. @kjohn-msft for more thoughts

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -344,24 +357,33 @@ def __new_assessment_summary_json(self, assessment_packages_json, status, code):
...
-            "lastModifiedTime": str(self.env_layer.datetime.timestamp()),
-            "startedBy": str(started_by),
-            "errors": self.__set_errors_json(self.__assessment_total_error_count, self.__assessment_errors)
-        }
+        # Compose substatus message
+        errors = self.__set_errors_json(self.__assessment_total_error_count, self.__assessment_errors)
+
+        substatus_message = self.__compose_assessment_substatus_msg(
+            activity_id=self.execution_config.activity_id, reboot_pending=self.is_reboot_pending, crit_patch_count=critsec_patch_count,
+            other_patch_count=other_patch_count, assessment_packages=assessment_packages_json, start_time=self.execution_config.start_time,
```
rane-rajasi, line 365: nit: assessment_packages can be renamed to just 'packages', since the function is explicitly for assessment

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -405,20 +427,31 @@ def __new_installation_summary_json(self, installation_packages_json):
         self.__refresh_installation_reboot_status()
 
         # Compose substatus message
+        maintenance_run_id = self.execution_config.maintenance_run_id if self.execution_config.maintenance_run_id is not None else ''
+        errors = self.__set_errors_json(self.__installation_total_error_count, self.__installation_errors)
+        substatus_message = self.__compose_installation_substatus_msg(activity_id=self.execution_config.activity_id, reboot_status=self.__installation_reboot_status,
+            maintenance_window=self.__maintenance_window_exceeded, not_selected=not_selected_patch_count, excluded=excluded_patch_count,
+            pending=pending_patch_count, installed=installed_patch_count, failed=failed_patch_count,
+            installation_packages=installation_packages_json, start_time=self.execution_config.start_time,
```
rane-rajasi, line 435: same comment as above for 'installation_packages'

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -595,7 +634,8 @@ def load_status_file_components(self, initial_load=False):
                 if self.execution_config.exec_auto_assess_only:
                     self.__installation_substatus_json = complete_status_file_data['status']['substatus'][i]
                 else:
-                    self.__installation_summary_json = self.__get_substatus_message(complete_status_file_data, i)
+                    self.__installation_summary_json = self.__get_substatus_msg_at_index(complete_status_file_data, i)
```
rane-rajasi, line 637: nit: function rename is not needed, the original function name also is reflective of fetching a singular substatus from the status file. 
NOTE: Try avoiding changes unrelated to the truncation logic, as much as possible, in this PR since it is pretty extensive on it's own

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -697,10 +739,18 @@ def __write_status_file(self):
...
+        self.env_layer.file_system.write_with_retry_using_temp_file(self.complete_status_file_path, '[{0}]'.format(status_file_payload_json_dumps), mode='w+')
+
+        if Constants.StatusTruncationConfig.TURN_ON_TRUNCATION:
+            status_file_size_in_bytes = self.__calc_status_size_on_disk(status_file_payload_json_dumps)     # calc complete_status_file_payload byte size on disk
 
-        # Write agent status file
-        self.env_layer.file_system.write_with_retry_using_temp_file(self.status_file_path, '[{0}]'.format(json.dumps(complete_status_payload)), mode='w+')
+            if status_file_size_in_bytes > self.__internal_file_capacity:  # perform truncation complete_status_file byte size > 126kb
+                truncated_status_file = self.__create_truncated_status_file(status_file_size_in_bytes, status_file_payload_json_dumps)
+                status_file_payload_json_dumps = json.dumps(truncated_status_file)
```
rane-rajasi, line 750: Everything from line 746 to 750 should go in a function. i.e. if truncation is enabled, call a function that first checks if truncation is needed, and then applies truncation, and returns the truncated status.

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -798,17 +844,213 @@ def __try_add_error(error_list, detail):
...
+            "message": str(formatted_message)
+        }
     # endregion
 
+    # region - Patch Truncation
+    def log_truncated_packages(self):
+        """ log the removed packages from patches in CoreMain after main operation are marked completed """
+        if not len(self.__assessment_packages_removed) == 0:
+            self.composite_logger.log_debug("Packages removed from assessment packages list: {0}".format(self.__assessment_packages_removed))
+        if not len(self.__installation_packages_removed) == 0:
```
rane-rajasi, line 880: No test coverage

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -827,3 +1069,108 @@ def __removed_older_complete_status_files(self, status_folder):
...
+
+        if installation_status_index is not None:
+            installation_msg_without_packages = self.__update_substatus_msg(substatus_msg=self.__installation_substatus_msg_copy, substatus_msg_patches=[])
+            status_file_no_list_data['status']['substatus'][installation_status_index]['formattedMessage']['message'] = json.dumps(installation_msg_without_packages)
+
+        return self.__calc_status_size_on_disk(json.dumps(status_file_no_list_data))
+
+    def __get_substatus_index(self, substatus_name, substatus):
+        """" Get substatus index from the current substatus """
+        for index, sub_name in enumerate(substatus):
```
rane-rajasi, line 1096: sub_name indicates it contains only the substatus name, whereas in here, sub_name contains the entire substatus. So the naming here is not correct. Please rename 'sub_name' to substatus and substatus can be renamed to substatus_list

src/core/src/service_interfaces/StatusHandler.py
feng-j678: done

src/core/src/service_interfaces/StatusHandler.py
feng-j678: done

src/core/src/service_interfaces/StatusHandler.py
feng-j678: the renamed from one of previous comment. change reverted.

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -798,17 +844,213 @@ def __try_add_error(error_list, detail):
...
     # endregion
 
+    # region - Patch Truncation
+    def log_truncated_packages(self):
+        """ log the removed packages from patches in CoreMain after main operation are marked completed """
+        if not len(self.__assessment_packages_removed) == 0:
+            self.composite_logger.log_debug("Packages removed from assessment packages list: {0}".format(self.__assessment_packages_removed))
+        if not len(self.__installation_packages_removed) == 0:
+            self.composite_logger.log_debug("Packages removed from installation packages list: {0}".format(self.__installation_packages_removed))
+        if len(self.__assessment_packages_removed) == 0 and len(self.__installation_packages_removed) == 0:
```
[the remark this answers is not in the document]
feng-j678: done. it is difficult to just test a method that do just log and no return without using mock. maybe in the future, we return a boolean value to make the test easier?

src/core/src/service_interfaces/StatusHandler.py
feng-j678: done. it is difficult to just test a method that do just log and no return without using mock. maybe in the future, we return a boolean value to make the test easier?

src/core/src/service_interfaces/StatusHandler.py
feng-j678: done

src/core/src/service_interfaces/StatusHandler.py
feng-j678: we are passing the new truncated_detail object into the errors json if error code is not error, if it is error then take the existing details object and add truncated code and truncated msg in self__try_add_errors. almost every tests for truncation test the error object and code. every function in recompose_subsatus_error is reused of existing functions, the new change is the True parameter that add extra logics in self.__set_errors to display warning or error code depending on # of error_count_by_operation and increment error count. 

image

kjohn-msft:
@feng-j678 @rane-rajasi 
Where is the conditional writes for performance tracked (write truncated status only once in x minutes)?
I don't see it in the PR description as tracked, and its not superficially clear that it's done.

feng-j678:
> @feng-j678 @rane-rajasi Where is the conditional writes for performance tracked (write truncated status only once in x minutes)? I don't see it in the PR description as tracked, and its not superficially clear that it's done.

checker condition code added.

src/core/src/service_interfaces/StatusHandler.py
feng-j678: done

src/core/src/CoreMain.py
feng-j678: pending for koshy's input on this

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -569,7 +590,10 @@ def load_status_file_components(self, initial_load=False):
 
         self.composite_logger.log_debug("Loading status file components [InitialLoad={0}].".format(str(initial_load)))
 
-        # Verify the status file exists - if not, reset status file
+        # Remove older complete status files
+        self.__removed_older_complete_status_files(self.execution_config.status_folder)
```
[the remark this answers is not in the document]
feng-j678: new master changes are pull into this pr. done

src/core/src/service_interfaces/StatusHandler.py
feng-j678: pending for rajasi to review recompose logic

src/core/src/service_interfaces/StatusHandler.py
feng-j678: pending for completion, the remove the comment

src/core/src/service_interfaces/StatusHandler.py
feng-j678: pending for rajasi review on this new logic

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -697,10 +760,17 @@ def __write_status_file(self):
...
 
         # Write complete status file <seq.no>.complete.status
-        self.env_layer.file_system.write_with_retry_using_temp_file(self.complete_status_file_path, '[{0}]'.format(json.dumps(complete_status_payload)), mode='w+')
+        status_file_payload_json_dumps = json.dumps(complete_status_payload)
+        self.env_layer.file_system.write_with_retry_using_temp_file(self.complete_status_file_path, '[{0}]'.format(status_file_payload_json_dumps), mode='w+')
 
-        # Write agent status file
-        self.env_layer.file_system.write_with_retry_using_temp_file(self.status_file_path, '[{0}]'.format(json.dumps(complete_status_payload)), mode='w+')
+        if Constants.StatusTruncationConfig.TURN_ON_TRUNCATION:
+            truncated_status_file_json_dumps = self.__validate_status_file_for_truncation(status_file_payload_json_dumps)
```
rane-rajasi, line 767: __validate_status_file_for_truncation(): Is this just validating or is this performing truncation also? The name indicates it will simply validate

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -798,17 +864,243 @@ def __try_add_error(error_list, detail):
...
+        if len(self.__assessment_packages_removed) == 0 and len(self.__installation_packages_removed) == 0:
+            self.composite_logger.log_debug("No packages truncated")
+
+    def __check_truncation_time_passed_by_x_sec(self, track_truncation_timestamp):
+        """ check current sys time is more than truncation time stamp by x constant seconds"""
+        if track_truncation_timestamp is None:
+            track_truncation_timestamp = datetime.datetime.now()
+
+        curr_timestamp = datetime.datetime.now()
+        return (curr_timestamp - track_truncation_timestamp).total_seconds() >= Constants.StatusTruncationConfig.SKIP_TRUNCATION_LOGIC_IN_X_SEC
```
rane-rajasi, line 911: why is this skip truncation by secs added?

Also, the function name again does not clearly state what is happening within

rane-rajasi (changes requested):
Did a very short pass, so this is not my final review for the current iteration. 

Please add a brief description on what this PR is trying to achieve with some details. For eg: how is it trying to truncate, when does it truncate, why is the skip by seconds logic added, etc.,

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -798,17 +864,243 @@ def __try_add_error(error_list, detail):
...
+
+    def __set_force_truncation_true(self, status):
```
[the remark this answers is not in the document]
feng-j678: done

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -697,10 +739,14 @@ def __write_status_file(self):
             shutil.rmtree(self.complete_status_file_path)
 
         # Write complete status file <seq.no>.complete.status
-        self.env_layer.file_system.write_with_retry_using_temp_file(self.complete_status_file_path, '[{0}]'.format(json.dumps(complete_status_payload)), mode='w+')
+        status_file_payload_json_dumps = json.dumps(complete_status_payload)
+        self.env_layer.file_system.write_with_retry_using_temp_file(self.complete_status_file_path, '[{0}]'.format(status_file_payload_json_dumps), mode='w+')
+
+        if Constants.StatusTruncationConfig.TURN_ON_TRUNCATION:
+            status_file_payload_json_dumps = self.__check_file_byte_size_for_truncation(status_file_payload_json_dumps)
```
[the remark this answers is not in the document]
feng-j678: method name updated -. __check_status_file_size_perform_truncation_if_needed(). at this point the method name is getting longer like paragraph. looking for suggestion for shorter name. done

src/core/src/service_interfaces/StatusHandler.py
feng-j678: done, see above comment

src/core/src/service_interfaces/StatusHandler.py
feng-j678: add description, done.

rane-rajasi:
> Did a very short pass, so this is not my final review for the current iteration.
> 
> Please add a brief description on what this PR is trying to achieve with some details. For eg: how is it trying to truncate, when does it truncate, why is the skip by seconds logic added, etc.,

PR description can be made further descriptive:
for eg: The last point states there is a delay added in the truncation logic, but does not explain why the delay is introduced.

PR description should briefly state the issue we see today and our solution. See the previous comment on "Please add a brief description on what this PR is trying to achieve with some details. For eg: how is it trying to truncate, when does it truncate, why is the skip by seconds logic added, etc.,"

src/core/src/CoreMain.py
rane-rajasi: This comment is still open and needs feedback. @feng-j678 please make sure all existing comments are addressed

src/core/src/service_interfaces/StatusHandler.py
rane-rajasi: Not a blocker for the first validation round: But the description, within function or in PR, does not explain "why" this skip truncation logic is added.

Also, based on my assumption, of this logic, truncation should be skipped if the remaining time (i.e. current - track_tuncation) <= Constants.StatusTruncationConfig.SKIP_TRUNCATION_LOGIC_IN_X_SEC. Is this correct?

So the last statement should be:
 return (curr_timestamp - track_truncation_timestamp).total_seconds() > Constants.StatusTruncationConfig.SKIP_TRUNCATION_LOGIC_IN_X_SEC

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -798,17 +864,243 @@ def __try_add_error(error_list, detail):
...
+        if self.__is_file_truncated and (status == Constants.STATUS_SUCCESS or status == Constants.STATUS_ERROR):
+            self.__force_truncation_on = True
+
+    def __check_status_file_size_perform_truncation_if_needed(self, status_file_payload_json_dumps):
+        status_file_size_in_bytes = self.__calc_status_size_on_disk(status_file_payload_json_dumps)  # calc complete_status_file_payload byte size on disk
+
+        if status_file_size_in_bytes > self.__internal_file_capacity:  # perform truncation complete_status_file byte size > 126kb
+            is_delay_truncation_by_x_sec = self.__check_truncation_time_passed_by_x_sec(self.__track_truncation_timestamp)
+
+            if is_delay_truncation_by_x_sec or self.__force_truncation_on:
```
rane-rajasi, line 924: Should this condition read as:
If there is no delay required for truncation OR force truncation is on?

If so, the check should be:
 if not is_delay_truncation_by_x_sec or self.__force_truncation_on:

This is why a description conveying the intent/ desired functionality of the code is important, either in function description or PR description

Also, the function and variable names need to be modified to reflect the intent, but leaving that for the next review

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -827,3 +1119,108 @@ def __removed_older_complete_status_files(self, status_folder):
...
+        return substatus_msg['errors']['code'], substatus_msg['errors']['details']
+
+    def __create_assessment_tombstone_list(self, packages_removed_from_assessment):
+        assessment_tombstone_map = {}
+        tombstone_record_list = []
+
+        # Map['classification', count]
+        for package in packages_removed_from_assessment:
+            classifications = package['classifications'][0]
+            assessment_tombstone_map[classifications] = assessment_tombstone_map.get(classifications, 0) + 1
```
rane-rajasi, line 1196: assessment_tombstone_map.get()

Wouldn't this throw an error on empty list?

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -827,3 +1119,108 @@ def __removed_older_complete_status_files(self, status_folder):
...
+
+        # Add assessment tombstone record per classifications except unclassified
+        for tombstone_classification, tombstone_package_count in assessment_tombstone_map.items():
```
rane-rajasi, line 1199: Where are things being written into assessment_tombstone_map?
Line 1188 initializes it to an empty value, line 1194 does a .get() on an empty list. Don't see any values being set in it

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -827,3 +1119,108 @@ def __removed_older_complete_status_files(self, status_folder):
...
+            if not tombstone_classification == Constants.PackageClassification.UNCLASSIFIED:
+                tombstone_record_list.append(self.__create_assessment_tombstone(tombstone_package_count, tombstone_classification))
```
rane-rajasi, line 1201: Are we adding 1 tombstone record for every classification type? I don't remember what was discussed on this previously but I have doubts on the usefulness of this.

rane-rajasi (changes requested):
Left some comments inline, but my overall review is:
- Need a description of the current functionality and why each change is being introduced (i.e. what is the desired outcome)
- The changes are there but due to the enormity of this PR and the number of things being added, it is difficult to ensure that all intended changes are implemented and everything works together as desired.
- we should consider splitting this PR into parts for easier review.

The parts can be: (1 PR for each)
1.	Basic truncation of assessment and installation patches (Simply removing additional patches with No tombstone or error details). This will also include re-composing the status file post truncation. TBD if this can be further split
2.	Adding tombstone records on truncation
3.	Adding error details on truncation
4.	Adding delay on truncation 
5.	Any other changes not covered so far

src/core/src/service_interfaces/StatusHandler.py
feng-j678: done. if that's the case, this code is unreachable:

line 1000
            if len(packages_removed_from_assessment) > 0:
                # Update current assessment # of removed packages
                self.__assessment_packages_removed = packages_removed_from_assessment

src/core/src/service_interfaces/StatusHandler.py
feng-j678: done. if that's the case, this code is unreachable:

line 1000
if len(packages_removed_from_assessment) > 0:
# Update current assessment # of removed packages
self.__assessment_packages_removed = packages_removed_from_assessment
                assessment_tombstone_records = self.__create_assessment_tombstone_list(self.__assessment_packages_removed)
                packages_retained_in_assessment.extend(assessment_tombstone_records)     # Add assessment tombstone record

src/core/src/service_interfaces/StatusHandler.py
feng-j678: done. yes, this was a one of the requirements.

PR #193, opened 2023-06-16
Add ordered dictionary for quicker look up of Assessment packages in StatusHandler

feng-j678:
What is in this PR:
[X] Modify the nested for loop in set_package_assessment_status to use order dictionary for quicker look up
[X] unit test assessment order dictionary

### DIFF
--- a/src/core/tests/Test_StatusHandler.py
+++ b/src/core/tests/Test_StatusHandler.py
@@ -407,10 +407,21 @@ class TestStatusHandler(unittest.TestCase):
 
         test_packages, test_package_versions = self.__set_up_packages_func(patch_count_for_test)
         status_handler.set_package_assessment_status(test_packages, test_package_versions, 'Critical')
+
+        with self.runtime.env_layer.file_system.open(self.runtime.execution_config.status_file_path, 'r') as file_handle:
+            substatus_file_data = json.load(file_handle)[0]["status"]["substatus"][0]
+
+        self.assertTrue(substatus_file_data["name"] == Constants.PATCH_ASSESSMENT_SUMMARY)
+        formatted_message = json.loads(substatus_file_data['formattedMessage']['message'])
+        self.assertEqual(len(formatted_message['patches']), 5)
+        self.assertEqual(formatted_message['patches'][0]['classifications'], ['Critical'])
         self.assertIsNotNone(status_handler._StatusHandler__assessment_packages_map)
+        self.assertEqual(len(status_handler._StatusHandler__assessment_packages), 5)
+        self.assertEqual(status_handler._StatusHandler__assessment_packages[0], expected_value)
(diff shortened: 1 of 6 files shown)


kjohn-msft (dismissed):

src/core/tests/Test_StatusHandler.py
```diff
@@ -396,6 +396,32 @@ def test_sort_packages_by_classification_and_state(self):
...
+        patch_id = 'python-samba0_2:4.4.5+dfsg-2ubuntu5.4_Ubuntu_16.04'
+        expected_value = {'version': '2:4.4.5+dfsg-2ubuntu5.4', 'classifications': ['Critical'], 'name': 'python-samba0', 'patchId': 'python-samba0_2:4.4.5+dfsg-2ubuntu5.4_Ubuntu_16.04'}
+
+        status_handler = StatusHandler(self.runtime.env_layer, self.runtime.execution_config, self.runtime.composite_logger, self.runtime.telemetry_writer, self.runtime.vm_cloud_type)
+        self.runtime.execution_config.operation = Constants.ASSESSMENT
+        self.runtime.status_handler.set_current_operation(Constants.ASSESSMENT)
+
+        test_packages, test_package_versions = self.__set_up_packages_func(patch_count_for_test)
+        status_handler.set_package_assessment_status(test_packages, test_package_versions, 'Critical')
+        self.assertIsNotNone(status_handler._StatusHandler__assessment_packages_map)
```
rane-rajasi, line 410: I don't agree with checking a functions internal variables in asserts this way. My preference would be to check the output of the function that's being tested. Which in this instance would be the output of set_package_assessment_status() 

Adding @kjohn-msft to bring this to attention

rane-rajasi (changes requested):
Comment inline

src/core/tests/Test_StatusHandler.py
feng-j678: add more assertions for checking if status_file has the new package chagnes

rane-rajasi: We don't need to check StatusHandler's internal variables to verify if packages exist. Line 412 is reading contents from the status file, you can directly check if the packages exist within that file.
Please remove lines 418 to 423 and check for packages within status file using the status file read in line 412.

Please see lines 415, 416, 417 for reference.

rane-rajasi (changes requested):
Comment inline

src/core/tests/Test_StatusHandler.py
feng-j678: done, the reference to internal variable is removed

kjohn-msft (approved):

rane-rajasi (approved):
Now that the py27 CI issue is fixed, let that PR go in first, so you can fetch latest from master into this PR, run py27 on this PR and then merge.

PR #199, opened 2023-06-20
add truncation algorithm steps

feng-j678:


src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -783,3 +783,140 @@ def __set_errors_json(self, error_count_by_operation, errors_by_operation):
             "message": message
         }
     # endregion
+
```
kjohn-msft, line 786: The expectation was to have things embedded in existing code. For e.g. new functions may be separate, but the calls to those functions should be annotated as comments in functions that will call them.

It's not easily understandable how the code flows and maps to the basic algo reviewed Friday. Maybe a discussion after addressing this comment helps. Because I am on call, make sure @rane-rajasi is okay with it this week.

rane-rajasi (changes requested):
In the same context as Koshy's comment, some suggestions for this PR:

- Copy a screenshot of the pseudo code we worked on Friday, into the description of this PR
- Address Koshy's comment, following the pseudo code from Friday.

PR #201, opened 2023-06-22
User/jf/add cached ci yml

feng-j678:
- add cached for py27 installation
- add cached for install coverage

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,7 +83,7 @@ jobs:
             ${{ runner.os }}-pip-
       - name: Install dependencies
         shell: cmd
-        run: pip install -r requirements.txt
+        run: pip install -r .github/workflows/requirements.txt
       - name: Run extension tests
         continue-on-error: true
         shell: cmd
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -101,4 +119,5 @@ jobs:
         if: contains( steps.getoutput.outputs.contents, 'FAILED (failures=' )
         run: |
           echo "${{ steps.getoutput.outputs.contents }}"
+          echo "There are failed tests"
```
kjohn-msft, line 122: Why has this changed?

kjohn-msft (commented):
Looks good - one open question

.github/workflows/ci.yml
feng-j678: this message is just to let user know there are some failed tests. I can remove it if you think it's unnecessary.

kjohn-msft: Sorry, when I'd left the comment, it looked like line 121 was replaced (not augmented) with 122. Nvm.

kjohn-msft (approved):

rane-rajasi (approved):

PR #202, opened 2023-06-29
telemetry change to get esm packages selected count

SathishMSFT:
Adding telemetry log to get count of ESM packages selected to install. 

Test:
* Ubuntu 18 Unattached.
* Ubuntu 18 Attached.

### DIFF
--- a/src/core/src/core_logic/PatchAssessor.py
+++ b/src/core/src/core_logic/PatchAssessor.py
@@ -81,7 +81,8 @@ class PatchAssessor(object):
                 self.status_handler.set_package_assessment_status(sec_packages, sec_package_versions, Constants.PackageClassification.SECURITY)
 
                 # Set the security-esm packages in status.
-                self.package_manager.set_security_esm_package_status(Constants.ASSESSMENT)
+                packages = []  # Passing empty packages.
+                self.package_manager.set_security_esm_package_status(Constants.ASSESSMENT, packages)
 
                 # ensure reboot status is set
                 reboot_pending = self.package_manager.is_reboot_pending()
--- a/src/core/src/core_logic/PatchAssessor.py
+++ b/src/core/src/core_logic/PatchAssessor.py
@@ -81,8 +81,7 @@ class PatchAssessor(object):
                 self.status_handler.set_package_assessment_status(sec_packages, sec_package_versions, Constants.PackageClassification.SECURITY)
 
                 # Set the security-esm packages in status.
-                packages = []  # Passing empty packages.
-                self.package_manager.set_security_esm_package_status(Constants.ASSESSMENT, packages)
+                self.package_manager.set_security_esm_package_status(Constants.ASSESSMENT, packages=[])
 
                 # ensure reboot status is set
                 reboot_pending = self.package_manager.is_reboot_pending()
--- a/src/core/src/package_managers/AptitudePackageManager.py
+++ b/src/core/src/package_managers/AptitudePackageManager.py
@@ -684,6 +684,7 @@ class AptitudePackageManager(PackageManager):
         esm_packages = []
         esm_package_versions = []
         esm_packages_found = False
+        esm_packages_selected_to_install = []
 
         for pkg, version in zip(packages, package_versions):
             if version != Constants.UA_ESM_REQUIRED:
@@ -700,6 +701,11 @@ class AptitudePackageManager(PackageManager):
             self.status_handler.add_error_to_status("{0} patches requires Ubuntu Pro for Infrastructure with Extended Security Maintenance".format(len(esm_packages)), Constants.PatchOperationErrorCodes.UA_ESM_REQUIRED) # Set the error status with the esm_package details. Will be used in portal.
             self.telemetry_writer.write_event("Filter esm packages [EsmPackagesCount={0}]".format(len(esm_packages)), Constants.TelemetryEventLevel.Informational)
 
-        self.composite_logger.log_debug("Filter esm packages : [TotalPackagesCount={0}][EsmPackagesCount={1}]".format(len(packages), len(esm_packages)))
+        # Log the Esm packages that will be patched in case of an Ubuntu Pro attached VM.
+        security_esm_update_query_success, security_esm_updates, security_esm_updates_versions = self.get_security_esm_updates()
+        if security_esm_update_query_success:
+            esm_packages_selected_to_install = [package for package in non_esm_packages if package in security_esm_updates]
+
+        self.composite_logger.log_debug("Filter esm packages : [TotalPackagesCount={0}][EsmPackagesCount={1}][SelectedEsmPackagesCount={2}]".format(len(packages), len(esm_packages), len(esm_packages_selected_to_install)))
         return non_esm_packages, non_esm_package_versions, esm_packages, esm_package_versions, esm_packages_found
 
--- a/src/core/src/package_managers/AptitudePackageManager.py
+++ b/src/core/src/package_managers/AptitudePackageManager.py
@@ -656,7 +656,7 @@ class AptitudePackageManager(PackageManager):
             elif operation == Constants.INSTALLATION:
                 if security_esm_update_query_success:
                     esm_packages_selected_to_install = [package for package in packages if package in security_esm_updates]
-                    self.composite_logger.log_debug("[SelectedEsmPackagesCount={0}]".format(len(esm_packages_selected_to_install)))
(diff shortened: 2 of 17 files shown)


nikhim-um (dismissed):

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -700,6 +701,11 @@ def filter_out_esm_packages(self, packages, package_versions):
             self.status_handler.add_error_to_status("{0} patches requires Ubuntu Pro for Infrastructure with Extended Security Maintenance".format(len(esm_packages)), Constants.PatchOperationErrorCodes.UA_ESM_REQUIRED) # Set the error status with the esm_package details. Will be used in portal.
             self.telemetry_writer.write_event("Filter esm packages [EsmPackagesCount={0}]".format(len(esm_packages)), Constants.TelemetryEventLevel.Informational)
 
-        self.composite_logger.log_debug("Filter esm packages : [TotalPackagesCount={0}][EsmPackagesCount={1}]".format(len(packages), len(esm_packages)))
+        # Log the Esm packages that will be patched in case of an Ubuntu Pro attached VM.
+        security_esm_update_query_success, security_esm_updates, security_esm_updates_versions = self.get_security_esm_updates()
```
GAURAVRAMRAKHYANI, line 705: I think there should be separate function for logging ESM packages selected to install. The responsibility of filter_out_esm_packages should only be filtering out ESM required packages from the packages to be installed.

The function filter_out_esm_packages might be called from multiple places in future but we need to log ESM packages to install only once. Creating a separate function will also make code readability better.

Also, the variable names are confusing in this function like for following line esm_packages_selected_to_install contains packages from the list non_esm_packages.

esm_packages_selected_to_install = [package for package in non_esm_packages if package in security_esm_updates]

GAURAVRAMRAKHYANI, line 705: Can we avoid calling self.get_security_esm_updates()?

Currently self.get_security_esm_updates() is called from set_security_esm_package_status() and here. We can call self.get_security_esm_updates() only once and use the list for both: logging and set_security_esm_package_status().

SathishMSFT: Unfortunately No. This method is used to just abstract Pro specific changes from the common code. We will not be reusing anywhere. If not here, we will have to unnecessarily create other set of methods which will be stubbed in YUM and Zypper.

So, I will keep this logic here as of now.

Yes, Variable names seems to be confusing. will update it.

SathishMSFT: For patching, This method is called first and then followed by the "set_security_esm_package_status". 
For Assessment, Only "set_security_esm_package_status" is called.
So, its difficult to put it in one place, as we still have to maintain if the query was success and other details. Hence, have to call placed in this.

SathishMSFT: @GAURAVRAMRAKHYANI : I have made changes to move the code change to call esm_updates to one place. Please check this.

GAURAVRAMRAKHYANI (dismissed):

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -654,6 +654,9 @@ def set_security_esm_package_status(self, operation):
                 if not self.ubuntu_pro_client.is_ubuntu_pro_client_attached:
                     self.status_handler.add_error_to_status("{0} patches requires Ubuntu Pro for Infrastructure with Extended Security Maintenance".format(len(security_esm_updates)), Constants.PatchOperationErrorCodes.UA_ESM_REQUIRED)
             elif operation == Constants.INSTALLATION:
+                if security_esm_update_query_success:
+                    esm_packages_selected_to_install = [package for package in packages if package in security_esm_updates]
+                    self.composite_logger.log_debug("[SelectedEsmPackagesCount={0}]".format(len(esm_packages_selected_to_install)))
```
kjohn-msft, line 659: Setting security ESM package status. [Operation=Installation][SelectedEsmPackagesCount={0}]

src/core/src/core_logic/PatchAssessor.py
```diff
@@ -81,7 +81,8 @@ def start_assessment(self):
                 self.status_handler.set_package_assessment_status(sec_packages, sec_package_versions, Constants.PackageClassification.SECURITY)
 
                 # Set the security-esm packages in status.
-                self.package_manager.set_security_esm_package_status(Constants.ASSESSMENT)
+                packages = []  # Passing empty packages.
+                self.package_manager.set_security_esm_package_status(Constants.ASSESSMENT, packages)
```
kjohn-msft, line 85: Since packages as a local variable is not used aside from the function call, it's better for code clarity to just have the function call like this in one line:

self.package_manager.set_security_esm_package_status(Constants.ASSESSMENT, packages=[])

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -679,27 +682,27 @@ def filter_out_esm_packages(self, packages, package_versions):
...
+        ua_esm_required_packages = []
+        ua_esm_required_package_versions = []
+        ua_esm_required_packages_found = False
 
         for pkg, version in zip(packages, package_versions):
             if version != Constants.UA_ESM_REQUIRED:
-                non_esm_packages.append(pkg)
-                non_esm_package_versions.append(version)
+                selected_packages.append(pkg)
+                selected_packages_versions.append(version)
```
kjohn-msft, line 694: What does selected_packages mean? non_esm_packages, etc seems more quickly intuitive here.

ua_esm_required_packages_* is fine

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -679,27 +682,27 @@ def filter_out_esm_packages(self, packages, package_versions):
...
 
-        esm_packages_found = len(esm_packages) > 0
-        if esm_packages_found:
-            self.status_handler.add_error_to_status("{0} patches requires Ubuntu Pro for Infrastructure with Extended Security Maintenance".format(len(esm_packages)), Constants.PatchOperationErrorCodes.UA_ESM_REQUIRED) # Set the error status with the esm_package details. Will be used in portal.
-            self.telemetry_writer.write_event("Filter esm packages [EsmPackagesCount={0}]".format(len(esm_packages)), Constants.TelemetryEventLevel.Informational)
+        ua_esm_required_packages_found = len(ua_esm_required_packages) > 0
+        if ua_esm_required_packages_found:
+            self.status_handler.add_error_to_status("{0} patches requires Ubuntu Pro for Infrastructure with Extended Security Maintenance".format(len(ua_esm_required_packages)), Constants.PatchOperationErrorCodes.UA_ESM_REQUIRED) # Set the error status with the esm_package details. Will be used in portal.
+            self.telemetry_writer.write_event("Filter esm packages [EsmPackagesCount={0}]".format(len(ua_esm_required_packages)), Constants.TelemetryEventLevel.Informational)
```
rane-rajasi, line 704: Logger should take care of writing to telemetry. In the sense, line 706, will write this to log and to telemetry. See log_debug() in CompositeLogger.py. Line 704 is not needed

rane-rajasi (changes requested):
Comment inline

src/core/src/package_managers/AptitudePackageManager.py
SathishMSFT: Operation=Installation is not necessary as this scenario will only be there in installation scenario. Will add the rest.

src/core/src/package_managers/AptitudePackageManager.py
SathishMSFT: Done

src/core/src/package_managers/AptitudePackageManager.py
SathishMSFT: Sure. Makes sense. Will update accordingly.

src/core/src/core_logic/PatchAssessor.py
SathishMSFT: Will change.

SathishMSFT: Done

src/core/src/package_managers/AptitudePackageManager.py
SathishMSFT: Done

src/core/src/package_managers/AptitudePackageManager.py
SathishMSFT: Done.

GAURAVRAMRAKHYANI (approved):

kjohn-msft (approved):

rane-rajasi (approved):

PR #207, opened 2023-07-10
raise exception if min py version is < 2.7

feng-j678:
[x] address the bug - Terminal exception AttributeError - 'datetime.timedelta' object has no attribute 'total_seconds'
[x]raise exception if min python version is < 2.7

apis call result
image

### DIFF
--- a/src/core/src/bootstrap/Bootstrapper.py
+++ b/src/core/src/bootstrap/Bootstrapper.py
@@ -174,3 +174,4 @@ class Bootstrapper(object):
                                             "Exception details: " + str(exception))
             if raise_if_not_sudo:
                 raise
+
--- a/src/core/src/bootstrap/Bootstrapper.py
+++ b/src/core/src/bootstrap/Bootstrapper.py
@@ -180,5 +180,5 @@ class Bootstrapper(object):
 
     def check_min_python_version(self):
         if sys.version_info < (2, 7):
-            self.composite_logger.log_debug("Python version is below 2.7")
+            self.composite_logger.log_error("Python version is below 2.7")
             raise Exception("Error: minimum python version is not met - Python version is below 2.7")
--- a/src/core/src/bootstrap/Bootstrapper.py
+++ b/src/core/src/bootstrap/Bootstrapper.py
@@ -141,6 +141,9 @@ class Bootstrapper(object):
         self.composite_logger.log("Linux distribution: " + str(self.env_layer.platform.linux_distribution()) + "\n")
         self.composite_logger.log("Process id: " + str(os.getpid()))
 
+        # check machine min python version is met
+        self.check_min_python_version()
+
         # Ensure sudo works in the environment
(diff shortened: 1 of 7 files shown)


src/core/src/bootstrap/Constants.py
```diff
@@ -335,3 +335,6 @@ class UbuntuProClientSettings(EnumBackport):
         MAX_OS_MAJOR_VERSION_SUPPORTED = 18
         MINIMUM_CLIENT_VERSION = "27.14.4"
 
+    class PythonVersionMsg(EnumBackport):
+        PYTHON_NOT_COMPATIBLE_ERROR_MSG = "Unsupported older Python version. Python version is below 2.7. To resolve: https://www.pythoncentral.io/how-to-update-python/"
```
kjohn-msft, line 339: Do not use external links in our code base. We cannot control the lifecycle of the contents at the URL, and we cannot guarantee the long-term quality of the guidance at any non-Microsoft controlled landing page.

Suggested message: "Unsupported older Python version. Minimum Python version required is 2.7. [DetectedPythonVersion={0}]"

We have to be conservative with logging for performance reasons - do not include PYTHON_COMPATIBLE_MSG or its log. For the PYTHON_NOT_COMPATIBLE_ERROR_MSG, since it's singular, no need to use enum. You can use it as a new line at line 303 to keep it close to similar error messages.

Additional note: This iteration ends the file without a new line at the end. That won't apply with the changes requested but it's good to cover that in all PRs.

src/core/src/bootstrap/Bootstrapper.py
```diff
@@ -174,4 +174,3 @@ def check_sudo_status(self, raise_if_not_sudo=True):
                                             "Exception details: " + str(exception))
             if raise_if_not_sudo:
                 raise
-
```
kjohn-msft, line 177: Revert -- please catch things like this with a review in GitHub after pushing iterations.

kjohn-msft (changes requested):
Comments inline.

rane-rajasi (changes requested):
Questions inline

src/core/src/core_logic/PatchAssessor.py
```diff
@@ -43,6 +44,7 @@ def start_assessment(self):
         """ Start a patch assessment """
         self.status_handler.set_current_operation(Constants.ASSESSMENT)
         self.raise_if_telemetry_unsupported()
+        self.raise_if_min_python_version_not_met()
```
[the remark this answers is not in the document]
kjohn-msft: 1. Yes, it's good to replicate similarly to PatchInstaller even though it'll not execute in any path today.

2. No, not in handler. There's nothing in handler that has this dependency, and we need to as gracefully report it in the correct operation log which can currently only happen in these specific places in Core.

kjohn-msft (approved):

rane-rajasi (approved):

PR #176, opened 2023-03-02
Update get security repo grep command

nicolas-morel-claranet:
Link with the issue #175

PR #208, opened 2023-07-13
Adding more sources to security sources list

rane-rajasi:
Changes include:
- adding /etc/apt/sources.list.d/ as well to the security sources list
- Removed error raised on this command failure (NOTE: This comes from the ask of not blocking the operation on this error but issuing a warning to the customer)

This should take care of the change attempted in: https://github.com/Azure/LinuxPatchExtension/pull/176

### DIFF
--- a/src/core/src/package_managers/AptitudePackageManager.py
+++ b/src/core/src/package_managers/AptitudePackageManager.py
@@ -39,7 +39,7 @@ class AptitudePackageManager(PackageManager):
 
         # Support to get updates and their dependencies
         self.security_sources_list = os.path.join(execution_config.temp_folder, 'msft-patch-security-{0}.list'.format(security_list_guid))
-        self.prep_security_sources_list_cmd = 'sudo grep security /etc/apt/sources.list > ' + os.path.normpath(self.security_sources_list)
+        self.prep_security_sources_list_cmd = 'sudo grep -hR security /etc/apt/sources.list /etc/apt/sources.list.d/ > ' + os.path.normpath(self.security_sources_list)
         self.dist_upgrade_simulation_cmd_template = 'LANG=en_US.UTF8 sudo apt-get -s dist-upgrade <SOURCES> '  # Dist-upgrade simulation template - <SOURCES> needs to be replaced before use; sudo is used as sometimes the sources list needs sudo to be readable
         self.single_package_check_versions = 'apt-cache madison <PACKAGE-NAME>'
         self.single_package_find_installed_dpkg = 'sudo dpkg -s <PACKAGE-NAME>'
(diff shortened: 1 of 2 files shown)


src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -182,9 +182,6 @@ def get_security_updates(self):
         code, out = self.env_layer.run_command_output(self.prep_security_sources_list_cmd, False, False)
         if code != 0:
             self.composite_logger.log_warning(" - SLP:: Return code: " + str(code) + ", Output: \n|\t" + "\n|\t".join(out.splitlines()))
-            error_msg = 'Unexpected return code (' + str(code) + ') from command: ' + self.prep_security_sources_list_cmd
-            self.status_handler.add_error_to_status(error_msg, Constants.PatchOperationErrorCodes.PACKAGE_MANAGER_FAILURE)
-            raise Exception(error_msg, "[{0}]".format(Constants.ERROR_ADDED_TO_STATUS))
```
rane-rajasi, line 187: Raising this comment to bring attention to these changes.

These lines were added as a quick fix for the ubuntu issue: https://github.com/Azure/LinuxPatchExtension/pull/162
Should be okay to remove now. 

And, for the other ask of issuing a warning, we do log this as warning currently. (have no other way I can think of, for marking this as a warning)

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -182,9 +182,6 @@ def get_security_updates(self):
         code, out = self.env_layer.run_command_output(self.prep_security_sources_list_cmd, False, False)
         if code != 0:
```
feng-j678, line 183: can we use a constant for the 0 value something like Constants.Status_Success, i think we have a constant in constant.py

src/core/tests/library/LegacyEnvLayerExtensions.py
```diff
@@ -526,7 +526,7 @@ def run_command_output(self, cmd, no_output=False, chk_err=True):
                         output = "      bash | 4.3-14ubuntu1.3 | http://us.archive.ubuntu.com/ubuntu xenial-updates/main amd64 Packages\n" + \
                                  "      bash | 4.3-14ubuntu1.2 | http://security.ubuntu.com/ubuntu xenial-security/main amd64 Packages\n" + \
                                  "      bash | 4.3-14ubuntu1 | http://us.archive.ubuntu.com/ubuntu xenial/main amd64 Packages"
-                    elif cmd.find('sudo grep security /etc/apt/sources.list >') > -1:
+                    elif cmd.find('sudo grep -hR security /etc/apt/sources.list /etc/apt/sources.list.d/ >') > -1:
                         self.write_to_file(os.path.join(self.temp_folder_path, "temp2.list"), "test temp file 2")
                         code = 0
```
feng-j678, line 531: same comment as above

feng-j678 (approved):

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: Checking code with 0 or some numeric value is used in many places, so this will include significant code change. Will have this in a separate PR

src/core/tests/library/LegacyEnvLayerExtensions.py
rane-rajasi: Same reply as for the other one, will include this in a separate PR

feng-j678 (approved):

kjohn-msft (approved):

kjohn-msft (approved):

PR #209, opened 2023-07-25
The output of command "yum install --assumeno" is different in yum 4 than that in yum 3 so adding parsing logic for yum 4

GAURAVRAMRAKHYANI:
**Issue:**
The output of command "yum install --assumeno" is different in yum 4 than that in yum 3. The output of the command "yum install --assumeno" is parsed to get the list of dependencies of a package. The current parsing logic is able to parse the output in case of yum 3 but not for yum 4. Due to this issue, the function get_dependent_list for yum package manager for distros using yum 4 is not returning any package. 

Due to this issue, there is impact on feature of exclude packages. The packages in the list of excluded packages is getting installed if the package is dependent of some other package. Example: If package p1 is dependent on package p2, p1 is in included list and p2 is in excluded list then package p1 should not be installed because installing p1 will install p2 also but as current parsing logic is not able to find dependencies of package, it is not identified that p1 is dependent on p2 and installing p1 will install p2 also. So, the package p1 is installed and p2 is installed by package manager as p1 is dependent on p2.

Out of the distros I tested, following distros have yum v3 installed:
RHEL 7.4, 7.6
CentOS 7.5, 7.6, 7.9

Following have yum v4 installed:
RHEL 8, 8.1, 8.4
CentOS 8.5

**Fix:**
Add parsing logic for output of command "yum install --assumeno" for yum version 4.

**Test**
Manual Testings:
(a) Tested that parsing logic is working properly in distros using yum 4. The function get_dependent_list is returning the expected dependencies.
(b) Did regression testing in distros using yum 3 with the code changes.
(c) Tested that excluded packages are not getting installed if any included package has dependency on it. Tested this for both versions i.e., yum 3 and yum 4.

Automated tests:
Added test case for function get_dependent_list for yum 4. There is already test for get_dependent_list for yum 3 in the test case test_package_manager in Test_YumPackageManager.py.

### DIFF
--- a/src/core/src/package_managers/YumPackageManager.py
+++ b/src/core/src/package_managers/YumPackageManager.py
@@ -332,6 +332,26 @@ class YumPackageManager(PackageManager):
         
         # In sample1, the starting delimeter is "Upgrading:" and in sample2, the starting delimeter is "Installing dependencies:"
         # If there are combinations of updates and new installations, one of the above delimeters will come first and the below logic will work.
+        # Following is the example when both "Upgrading:" and "Installing dependencies:" are present:
+        #
+        #Last metadata expiration check: 0:01:56 ago on Tue 25 Jul 2023 12:01:47 PM UTC.
+        #Dependencies resolved.
+        #================================================================================================
+        # Package             Arch    Version               Repository                               Size
+        #================================================================================================
+        #Upgrading:
+        # kernel-tools        x86_64  4.18.0-372.64.1.el8_6 rhel-8-for-x86_64-baseos-eus-rhui-rpms  8.4 M
+        # kernel-tools-libs   x86_64  4.18.0-372.64.1.el8_6 rhel-8-for-x86_64-baseos-eus-rhui-rpms  8.2 M
+        # openssl             x86_64  1:1.1.1k-9.el8_6      rhel-8-for-x86_64-baseos-eus-rhui-rpms  710 k
+        # openssl-libs        x86_64  1:1.1.1k-9.el8_6      rhel-8-for-x86_64-baseos-eus-rhui-rpms  1.5 M
+        #Installing dependencies:
+        # kernel-core         x86_64  4.18.0-372.64.1.el8_6 rhel-8-for-x86_64-baseos-eus-rhui-rpms   40 M
+        # kernel-modules      x86_64  4.18.0-372.64.1.el8_6 rhel-8-for-x86_64-baseos-eus-rhui-rpms   32 M
+        #
+        #Transaction Summary
+        #================================================================================================
+        #Install  2 Packages
+        #Upgrade  4 Packages
         
         # There is one edge scenario which is also handled in the below parsing logic i.e., the package detail spills over to two lines:
         # Sample output for the cmd 'sudo yum update --assumeno polkit.x86_64' is :
--- a/src/core/src/package_managers/YumPackageManager.py
+++ b/src/core/src/package_managers/YumPackageManager.py
@@ -333,11 +333,34 @@ class YumPackageManager(PackageManager):
         # In sample1, the starting delimeter is "Upgrading:" and in sample2, the starting delimeter is "Installing dependencies:"
         # If there are combinations of updates and new installations, one of the above delimeters will come first and the below logic will work.
         
+        # There is one edge scenario which is also handled in the below parsing logic i.e., the package detail spills over to two lines:
+        # Sample output for the cmd 'sudo yum update --assumeno polkit.x86_64' is :
+        #
+        #Last metadata expiration check: 0:08:47 ago on Tue 25 Jul 2023 02:14:28 PM UTC.
+        #Package polkit-0.115-13.el8_5.2.x86_64 is already installed.
+        #Dependencies resolved.
+        #================================================================================
+        # Package   Arch   Version          Repository                              Size
+        #================================================================================
+        #Upgrading:
+        # polkit    x86_64 0.115-14.el8_6.1 rhel-8-for-x86_64-baseos-eus-rhui-rpms 154 k
+        # polkit-libs
+        #           x86_64 0.115-14.el8_6.1 rhel-8-for-x86_64-baseos-eus-rhui-rpms  77 k
+        #
+        #Transaction Summary
+        #================================================================================
+        #Upgrade  2 Packages
+        #
+        #Total download size: 231 k
+        #Operation aborted.
+
+        package_extensions = ["x86_64", "noarch", "i686"]
         dependencies = []
         lines = output.strip().split('\n')
         upgrading_or_installing_flag = False
 
-        for line in lines:
+        for line_index in range(0, len(lines)):
+            line = lines[line_index]
             if not upgrading_or_installing_flag:
                 if not line.startswith("Upgrading:") and not line.startswith("Installing dependencies:"):
                     self.composite_logger.log_debug(" - Inapplicable line: " + str(line))
@@ -347,12 +370,28 @@ class YumPackageManager(PackageManager):
                     self.composite_logger.log_debug(" - Packages list started: " + str(line))
                     continue
 
+            if line.startswith("Transaction Summary"):
+                break
+
             updates_line = re.split(r'\s+', line.strip())
-            if len(updates_line) != 6:
+            dependent_package_name = ""
+
+            if len(updates_line) == 6 and updates_line[1] in package_extensions:
+                dependent_package_name = updates_line[0] + "." + updates_line[1]
+            elif len(updates_line) == 5 and line_index > 0 and updates_line[0] in package_extensions:
+                # Handling the scenario when package detail is spill over to 2 lines
+                prev_line = lines[line_index-1]
+                prev_updates_line = re.split(r'\s+', prev_line.strip())
+                if len(prev_updates_line) == 1:
+                    # Taking package name from previuos line and package extension from current line.
+                    dependent_package_name = prev_updates_line[0] + "." + updates_line[0]
+                else:
+                    self.composite_logger.log_debug(" - Inapplicable line: " + str(line))
+                    continue
+            else:
                 self.composite_logger.log_debug(" - Inapplicable line: " + str(line))
                 continue
 
-            dependent_package_name = updates_line[0] + "." + updates_line[1]
             if len(dependent_package_name) != 0 and dependent_package_name not in packages:
                 self.composite_logger.log_debug(" - Dependency detected: " + dependent_package_name)
                 dependencies.append(dependent_package_name)
--- a/src/core/src/package_managers/YumPackageManager.py
+++ b/src/core/src/package_managers/YumPackageManager.py
@@ -932,6 +932,14 @@ class YumPackageManager(PackageManager):
                 self.composite_logger.log_debug(" - Inapplicable line: " + str(line))
                 continue
             else:
+                # The first string should be process ID and hence it should be integer.
+                # If first string is not process ID then the line is not for a process detail.
+                try:
+                    int(process_details[0])
+                except Exception:
+                    self.composite_logger.log_debug(" - Inapplicable line: " + str(line))
+                    continue
+
                 self.composite_logger.log_debug(" - Applicable line: " + str(line))
                 process_count += 1
                 process_list_verbose += process_details[1] + " (" + process_details[0] + "), "  # process name and id
--- a/src/core/src/package_managers/YumPackageManager.py
+++ b/src/core/src/package_managers/YumPackageManager.py
@@ -386,6 +386,9 @@ class YumPackageManager(PackageManager):
                     # Taking package name from previuos line and package extension from current line.
                     dependent_package_name = prev_updates_line[0] + "." + updates_line[0]
                 else:
+                    # This is unexpected scenario.
+                    # If update details are divided in two lines then first line should contain package name and second should contain rest of the details.
+                    # Hence previous line should contain package name and current line should contain other 5 details.
                     self.composite_logger.log_debug(" - Inapplicable line: " + str(line))
                     continue
             else:
--- a/src/core/src/package_managers/YumPackageManager.py
+++ b/src/core/src/package_managers/YumPackageManager.py
@@ -54,6 +54,9 @@ class YumPackageManager(PackageManager):
         self.yum_ps_prerequisite = 'sudo yum -y install yum-plugin-ps'
         self.yum_ps = 'sudo yum ps'
 
+        # check yum version
+        self.yum_version_check = "sudo yum --version"
+
         # auto OS updates
         self.current_auto_os_update_service = None
         self.os_patch_configuration_settings_file_path = ''
@@ -96,6 +99,8 @@ class YumPackageManager(PackageManager):
                                        "Error: Failed to download metadata for repo":  self.fix_ssl_certificate_issue}
         
         self.yum_update_client_package = "sudo yum update -y --disablerepo='*' --enablerepo='*microsoft*'"
+
+        self.yum_version = self.get_yum_version()
 
     def refresh_repo(self):
         pass  # Refresh the repo is no ops in YUM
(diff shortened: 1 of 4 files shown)


src/core/src/package_managers/YumPackageManager.py
```diff
@@ -328,7 +463,13 @@ def get_dependent_list(self, packages):
 
         self.composite_logger.log_debug("\nRESOLVING DEPENDENCIES USING COMMAND: " + str(self.single_package_upgrade_simulation_cmd + package_names))
         output = self.invoke_package_manager(self.single_package_upgrade_simulation_cmd + package_names)
-        dependencies = self.extract_dependencies(output, packages)
+        dependencies = []
+
+        if self.yum_version >= 4:
```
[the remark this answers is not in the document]
GAURAVRAMRAKHYANI: We should try parsing dependency in one of the ways if we are not able to find yum version. Better way is to parse using existing parsing logic which should be consistent with current behavior without this code changes.

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -282,7 +287,118 @@ def is_package_version_installed(self, package_name, package_version):
...
+                dependent_package_name = updates_line[0] + "." + updates_line[1]
+            elif len(updates_line) == 5 and line_index > 0 and updates_line[0] in package_extensions:
+                # Handling the scenario when package detail is spill over to 2 lines
+                prev_line = lines[line_index-1]
+                prev_updates_line = re.split(r'\s+', prev_line.strip())
+                if len(prev_updates_line) == 1:
+                    # Taking package name from previuos line and package extension from current line.
+                    dependent_package_name = prev_updates_line[0] + "." + updates_line[0]
+                else:
+                    self.composite_logger.log_debug(" - Inapplicable line: " + str(line))
```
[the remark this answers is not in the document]
GAURAVRAMRAKHYANI: Sure, will add test for this scenario although this is not expected scenario.

src/core/tests/library/RuntimeCompositor.py
```diff
@@ -41,7 +41,7 @@
 
 
 class RuntimeCompositor(object):
-    def __init__(self, argv=Constants.DEFAULT_UNSPECIFIED_VALUE, legacy_mode=False, package_manager_name=Constants.APT, vm_cloud_type=Constants.VMCloudType.AZURE):
+    def __init__(self, argv=Constants.DEFAULT_UNSPECIFIED_VALUE, legacy_mode=False, package_manager_name=Constants.APT, vm_cloud_type=Constants.VMCloudType.AZURE, test_type="HappyPath"):
```
SathishMSFT, line 44: What is the need for this?

src/core/src/package_managers/YumPackageManager.py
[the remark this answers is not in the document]
GAURAVRAMRAKHYANI: Added test case for code coverage.

src/core/tests/library/RuntimeCompositor.py
GAURAVRAMRAKHYANI: The package manager initialization is done in this init function of RuntimeCompositor. The init function of YumPackageManager has function call get_yum_version which runs command "sudo yum --version". The sample output of "sudo yum --version" is added in the test_type "YumVersion4Dependency" for yum version 4. Without this change, the default test_type is "HappyPath" which does not contain output of "sudo yum --version". 

This code changes are done so that RuntimeCompositor can be initialized with input test_type. Initializing RuntimeCompositor with the given test_type will initialize LegacyEnvLayerExtensions with the given test_type and hence  package manager with sample output of given test_type. I have made the required changes in init function of LegacyEnvLayerExtensions also. Without this change, the package managers were always initialized with outputs in the test_type "HappyPath" as per init function of LegacyEnvLayerExtensions. Added parameter test_type in the init function of LegacyEnvLayerExtensions.

SathishMSFT: For this test, i feel simply setting the yum_Version in your test as 4 will do. This seems to be too much change for this scenario.

SathishMSFT: Had offline discussion. This will be need to test Code in the init methods of package manager,.

SathishMSFT (approved):

nikhim-um (approved):

rane-rajasi:
> **Issue:** The output of command "yum install --assumeno" is different in yum 4 than that in yum 3. The output of the command "yum install --assumeno" is parsed to get the list of dependencies of a package. The current parsing logic is able to parse the output in case of yum 3 but not for yum 4. Due to this issue, the function get_dependent_list for yum package manager for distros using yum 4 is not returning any package.
> 
> Due to this issue, there is impact on feature of exclude packages. The packages in the list of excluded packages is getting installed if the package is dependent of some other package. Example: If package p1 is dependent on package p2, p1 is in included list and p2 is in excluded list then package p1 should not be installed because installing p1 will install p2 also but as current parsing logic is not able to find dependencies of package, it is not identified that p1 is dependent on p2 and installing p1 will install p2 also. So, the package p1 is installed and p2 is installed by package manager as p1 is dependent on p2.
> 
> Out of the distros I tested, following distros have yum v3 installed: RHEL 7.4, 7.6 CentOS 7.5, 7.6, 7.9
> 
> Following have yum v4 installed: RHEL 8, 8.1, 8.4 CentOS 8.5
> 
> **Fix:** Add parsing logic for output of command "yum install --assumeno" for yum version 4.
> 
> **Test** Manual Testings: (a) Tested that parsing logic is working properly in distros using yum 4. The function get_dependent_list is returning the expected dependencies. (b) Did regression testing in distros using yum 3 with the code changes. (c) Tested that excluded packages are not getting installed if any included package has dependency on it. Tested this for both versions i.e., yum 3 and yum 4.
> 
> Automated tests: Added test case for function get_dependent_list for yum 4. There is already test for get_dependent_list for yum 3 in the test case test_package_manager in Test_YumPackageManager.py.

[Rajasi] Question in the example given above where p1 is dependent on p2, with p1 in the included list and p2 in the excluded list. If by current logic, p2 is not identified as a dependency of p1. then how is p2 getting installed?
Shouldn't it be, p1 is installed, and since there are no dependencies identified, p2 is not installed?

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -282,7 +287,121 @@ def is_package_version_installed(self, package_name, package_version):
...
+        #Transaction Summary
+        #=============================================================================================
+        #Install  2 Packages
+        #
+        #Total download size: 72 M
+        #Installed size: 93 M
+        #Operation aborted.
+        
+        # In sample1, the starting delimeter is "Upgrading:" and in sample2, the starting delimeter is "Installing dependencies:"
+        # If there are combinations of updates and new installations, one of the above delimeters will come first and the below logic will work.
```
rane-rajasi, line 334: Please list a sample of what is seen when there are combinations of updates and installations.

Something to pay special attention to here is, where does "Transaction summary" appear in this instance. The implementation below breaks out of the for loop upon encountering "Transaction summary". In a case where there are both upgrades and installations and 2 separate transaction summaries (one for each), the loop would stop after the first one and never parse the 2nd list

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -282,7 +287,121 @@ def is_package_version_installed(self, package_name, package_version):
...
+                    continue
+
+            if line.startswith("Transaction Summary"):
+                break
+
+            updates_line = re.split(r'\s+', line.strip())
+            dependent_package_name = ""
+
+            if len(updates_line) == 6 and updates_line[1] in package_extensions:
+                dependent_package_name = updates_line[0] + "." + updates_line[1]
```
rane-rajasi, line 380: Shouldn't this be: dependent_package_name = updates_line[1] + "." + updates_line[2]? 

Since updates_line[1] is the package name and updates_line[2] Arch

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -282,7 +287,121 @@ def is_package_version_installed(self, package_name, package_version):
...
+                    self.composite_logger.log_debug(" - Packages list started: " + str(line))
+                    continue
+
+            if line.startswith("Transaction Summary"):
+                break
+
+            updates_line = re.split(r'\s+', line.strip())
+            dependent_package_name = ""
+
+            if len(updates_line) == 6 and updates_line[1] in package_extensions:
```
rane-rajasi, line 379: In the 2nd condition, we need to look for the package in package_extensions. updates_line[**1**] is returning the arch from this line:
# polkit    x86_64 0.115-14.el8_6.1 rhel-8-for-x86_64-baseos-eus-rhui-rpms 154 k

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -282,7 +287,121 @@ def is_package_version_installed(self, package_name, package_version):
...
+                dependent_package_name = updates_line[0] + "." + updates_line[1]
+            elif len(updates_line) == 5 and line_index > 0 and updates_line[0] in package_extensions:
```
rane-rajasi, line 381: len(updates_line) == 5 matches line 2 in the sample:

# polkit-libs
        #           x86_64 0.115-14.el8_6.1 rhel-8-for-x86_64-baseos-eus-rhui-rpms  77 k

In this case updates_line[0] would return 'x86_64' and not the package name, so the 2nd condition is not what we want to match. Also, a point of consideration in this example is, the second line does not contain the package name

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -282,7 +287,121 @@ def is_package_version_installed(self, package_name, package_version):
...
+                    self.composite_logger.log_debug(" - Inapplicable line: " + str(line))
+                    continue
+            else:
+                self.composite_logger.log_debug(" - Inapplicable line: " + str(line))
+                continue
+
+            if len(dependent_package_name) != 0 and dependent_package_name not in packages:
+                self.composite_logger.log_debug(" - Dependency detected: " + dependent_package_name)
+                dependencies.append(dependent_package_name)
+
```
rane-rajasi, line 401: I have some concerns with this implementation:

1.  The separation of logic for yum3 and yum4: While there are only 2 separate chunks right now, we cannot continue this line of implementation for each new version. There should be a generic way to handle this for all versions. 

2.  Adding to the first point: "All versions" includes implementing a solution that's easy to adapt to for future versions (i.e. any o/p format changes should be easy to incorporate in the existing logic)

3. Assumption on how the o/p text is structured: 

image
We are only considering these 2 samples with the code being designed with these fixed formats. Any change in these o/p formats would need a code change in future, which does not seem feasible.
for eg: 
            - what if the o/p is split in 3 lines? 
            - or the o/p is split in 2 lines but structured something like:
                        polkit-libs x86_64 0.115-14.el8_6.1
                                                                     rhel-8-for-x86_64-baseos-eus-rhui-rpms  77 k

rane-rajasi (changes requested):
Comments inline

GAURAVRAMRAKHYANI:
> > **Issue:** The output of command "yum install --assumeno" is different in yum 4 than that in yum 3. The output of the command "yum install --assumeno" is parsed to get the list of dependencies of a package. The current parsing logic is able to parse the output in case of yum 3 but not for yum 4. Due to this issue, the function get_dependent_list for yum package manager for distros using yum 4 is not returning any package.
> > Due to this issue, there is impact on feature of exclude packages. The packages in the list of excluded packages is getting installed if the package is dependent of some other package. Example: If package p1 is dependent on package p2, p1 is in included list and p2 is in excluded list then package p1 should not be installed because installing p1 will install p2 also but as current parsing logic is not able to find dependencies of package, it is not identified that p1 is dependent on p2 and installing p1 will install p2 also. So, the package p1 is installed and p2 is installed by package manager as p1 is dependent on p2.
> > Out of the distros I tested, following distros have yum v3 installed: RHEL 7.4, 7.6 CentOS 7.5, 7.6, 7.9
> > Following have yum v4 installed: RHEL 8, 8.1, 8.4 CentOS 8.5
> > **Fix:** Add parsing logic for output of command "yum install --assumeno" for yum version 4.
> > **Test** Manual Testings: (a) Tested that parsing logic is working properly in distros using yum 4. The function get_dependent_list is returning the expected dependencies. (b) Did regression testing in distros using yum 3 with the code changes. (c) Tested that excluded packages are not getting installed if any included package has dependency on it. Tested this for both versions i.e., yum 3 and yum 4.
> > Automated tests: Added test case for function get_dependent_list for yum 4. There is already test for get_dependent_list for yum 3 in the test case test_package_manager in Test_YumPackageManager.py.
> 
> [Rajasi] Question in the example given above where p1 is dependent on p2, with p1 in the included list and p2 in the excluded list. If by current logic, p2 is not identified as a dependency of p1. then how is p2 getting installed? Shouldn't it be, p1 is installed, and since there are no dependencies identified, p2 is not installed?

When the command to install only p1 is run, the package manager installs two packages p1 and p2 because p1 is dependent on p2. Ideally, we should not attempt to install p1.

src/core/src/package_managers/YumPackageManager.py
GAURAVRAMRAKHYANI: |	Last metadata expiration check: 0:01:56 ago on Tue 25 Jul 2023 12:01:47 PM UTC.
	|	Dependencies resolved.
	|	================================================================================================
	|	 Package             Arch    Version               Repository                               Size
	|	================================================================================================
	|	Upgrading:
	|	 kernel-tools        x86_64  4.18.0-372.64.1.el8_6 rhel-8-for-x86_64-baseos-eus-rhui-rpms  8.4 M
	|	 kernel-tools-libs   x86_64  4.18.0-372.64.1.el8_6 rhel-8-for-x86_64-baseos-eus-rhui-rpms  8.2 M
	|	 openssl             x86_64  1:1.1.1k-9.el8_6      rhel-8-for-x86_64-baseos-eus-rhui-rpms  710 k
	|	 openssl-libs        x86_64  1:1.1.1k-9.el8_6      rhel-8-for-x86_64-baseos-eus-rhui-rpms  1.5 M
	|	Installing dependencies:
	|	 kernel-core         x86_64  4.18.0-372.64.1.el8_6 rhel-8-for-x86_64-baseos-eus-rhui-rpms   40 M
	|	 kernel-modules      x86_64  4.18.0-372.64.1.el8_6 rhel-8-for-x86_64-baseos-eus-rhui-rpms   32 M
	|	
	|	Transaction Summary
	|	================================================================================================
	|	Install  2 Packages
	|	Upgrade  4 Packages

GAURAVRAMRAKHYANI: Above is the sample of what is seen when there are combinations of updates and installations.
The string "Transaction Summary" comes after both sections "Upgrading" and "Installing dependencies" are completed.

Should I add this sample in the code as a comment? I did not add it because there is already long comment in this function. Might be we can just mention in a comment that "Transaction Summary" comes after both sections of updates and installations are completed.

src/core/src/package_managers/YumPackageManager.py
GAURAVRAMRAKHYANI: For point 1 and 2:
I tried searching other ways but could not find any. The only way I could figure out to fetch details is through parsing output of yum commands. I agree that it is not feasible to add implementation for every new YUM version with different output format but I think we don't have any other option currently.

For point 3:
I did test in multiple RHEL and Oracle VMs and could see only these formats. I have till now tested in 6-7 different VMs. Hence, I did implementation for the formats which I have observed. I agree that any change in these o/p formats would need a code change in future, but it will be very complicated to handle all the possible cases through parsing logic. The cases could be:
(a) o/p split in 3 or more lines.
(b) o/p split in 2 lines but structured differently and there can be many combinations of such cases.
It will be complicated to add parsing logic for all the cases as mentioned in (a) and (b)

The other output parsing logics in this repo are also based on observations and there can be issues if output is structured differently than expected.

src/core/src/package_managers/YumPackageManager.py
GAURAVRAMRAKHYANI: Sorry, I did not get you. What do you mean by "we need to look for the package in package_extensions"?

Second condition is check that the architecture i.e. updates_line[1] = "x86_64" should be one of the package_extensions (["x86_64", "noarch", "i686"])

src/core/src/package_managers/YumPackageManager.py
GAURAVRAMRAKHYANI: In this case, the package name is in first line and rest of the details are in second line. 

We are not extracting package when line = "polkit-libs". We are extracting package when line = "        x86_64 0.115-14.el8_6.1 rhel-8-for-x86_64-baseos-eus-rhui-rpms 77 k"

line = "        x86_64 0.115-14.el8_6.1 rhel-8-for-x86_64-baseos-eus-rhui-rpms 77 k"
prev_line = "polkit-libs"

updates_line[0] = "x86_64"
prev_updates_line[0] = "polkit-libs"
dependent_package_name = prev_updates_line[0] + "." + updates_line[0] = "polkit-libs.x86_64"

src/core/src/package_managers/YumPackageManager.py
rane-rajasi: Better to keep it in code as a comment, will act as a reference for future

src/core/src/package_managers/YumPackageManager.py
rane-rajasi: Q: why do we have the package_extensions list? This wasn't a part of the old function (now extract_dependencies_yum_3_or_lower())

src/core/src/package_managers/YumPackageManager.py
rane-rajasi: This still needs some more thought. I too will spend some time to think over it.

src/core/src/package_managers/YumPackageManager.py
GAURAVRAMRAKHYANI: This is a check to make sure that string is a package extension. As there could be different output structures like package detail spill over to multi lines, to make sure that the string is package extension, I have added this check. Although, till now only observed two cases: package detail in single line or two lines with first string in second line is package extension.

There is similar check in the method extract_packages_and_versions_including_duplicates.

src/core/src/package_managers/YumPackageManager.py
GAURAVRAMRAKHYANI: Added the sample output as comment

src/core/src/package_managers/YumPackageManager.py
rane-rajasi: Adding @kjohn-msft to the discussion...

kjohn-msft (changes requested):
Added comment to section that was raised with me. Please discuss and address.

kjohn-msft:
**NOTES FROM DISCUSSION 08/21 ---**
- Gaurav's busy and is unable to finish this PR in the near term. He's asked if AzGPS can take it.
- AzGPS is back-logged too and may not get to it sooner than him. If AzGPS gets to it, it will be with positive ack to Gaurav first.
- General outcome: work is paused until further notice. Whoever has bandwidth first to notify group.,

kjohn-msft:
Discussed ETA with Gaurav - mid-November.

src/core/src/package_managers/YumPackageManager.py
GAURAVRAMRAKHYANI: @kjohn-msft, @rane-rajasi  :: I wanted to share pros and cons of all the three approaches: approach in existing PR and the two approaches that @kjohn-msft has shared.

The two approaches that Koshy has shared addresses the point 3 but they do not address point 1 and 2 raised by Rajasi.

There are some Pros and Cons for these two approaches in compare to the existing logic in the PR.
Pros:
1. There is no need to find yum version in these two approaches. Finding yum version itself involves parsing output of a command and there could be different output format which are not observed during testing. So, it is good to avoid finding yum version.
2. There is single function for parsing output for both yum v3 and yum v4 output format.

Cons:
1. As there is single function for parsing both yum v3 and yum v4 output format, the function becomes little complex due to more conditions as it needs to handle both kind of output format.

Overall, I think that the approaches that Koshy has shared are better than existing approach in the PR. I will once analyze both approaches that Koshy has shared in terms of implementation as there are different output formats so need to take care of different formats and I will share my findings before raising new iteration with one of the two approaches.

rane-rajasi: @GAURAVRAMRAKHYANI In my opinion, @kjohn-msft suggestions address all of my concerns. 

The main idea behind all 3 points I raised was that we should avoid introducing dependencies, have generic code  that covers all scenarios with as little or possibly no special handling

for eg: in the current PR, there is one block to address yum3 o/p, another to address yum4, later there might be another. This approach goes on to simply add lines of code, which is the code bloat part Koshy mentioned.

We should have a generic solution/function which is able to handle both types of o/p (yum3 and yum4) and should be scalable to address more such instances in future. (which was point 2 in my previous comment)

To summarize, what I'm leaning towards is a code that handles all types of o/p without being dependent on identifying yum versions or a only certain format of multiline o/p.

To move towards implementing these suggestions and removing dependencies (yum version and format independent multi-line), I believe needs a some code change in this PR. Let me know if you would be able to pick these up and when

PR #213, opened 2023-08-09
Extend Composite Logger log functions to have built in buffer

GAURAVRAMRAKHYANI:
Extend Composite Logger functions to have built in buffer. This will help in combining multiple messages in single log and hence decreasing number of logs and hence time loss due to telemetry throttling is decreased.

A new optional parameter is added in log functions. The data type of the parameter is enum whose values can be true, flush and false. The default value is false.

If the buffer is true, keep appending to built-in variable.
The buffer may be flushed to the log in the very last message meant to be part of the concatenation.
If it's false, write whatever is in memory, clear it out, and also separately write the new message that came in.

The buffer line separator is used to separate different messages in the same buffer - "\n|\t".

There is common buffer for different log levels i.e., error, warning, debug and verbose. This is simpler design than having separate buffers for each log level. The issue with having single buffer for all log levels is that if log function with warning level is called 3 times with different messages with buffer=true and then log function with error level is called with buffer=flush then all the 4 messages will be written with error level because there is single buffer for all log levels. The current use cases of decreasing number of logs by combining multiple messages to single log can be achievable with single buffer for all log levels.

If required later, the design can be expanded to have separate buffers each for error, warning, debug and verbose.

Manual testing done:
Tested the different sequences of logging messages with the buffer value true, false and flush. Tested following scenarios with all log levels i.e. error, warning, debug and verbose:
(a) When log function is called 2 times with buffer=true and third time with buffer=flush, then the three logs are combined to single log. 
(b) When log function is called with buffer=false then the message is logged.
(c) When log function is called with buffer=true followed by log function with buffer=false then the message with buffer=true is logged first and then the message with buffer=false is logged as a separate log.

Tested the scenarios with number of characters in message exceed the limit of maximum characters.
(a) If log function is called with buffer=true and then a new log function is called with buffer=flush and if sum of characters of both the messages exceeds the limit then both messages are logged separately.

### DIFF
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -335,3 +335,16 @@ class Constants(object):
         MAX_OS_MAJOR_VERSION_SUPPORTED = 18
         MINIMUM_CLIENT_VERSION = "27.14.4"
 
+    BUFFER_LINE_SEPERATOR = "\n|\t"
+    BUFFER_LINE_SEPERATOR_LENGTH = len(BUFFER_LINE_SEPERATOR)
+
+    class BufferMessage(EnumBackport):
+        TRUE = 0
+        FALSE = 1
+        FLUSH = 2
+
+    class MessageLevel(EnumBackport):  
+        ERROR = 0
+        WARNING = 1
+        DEBUG = 2
+        VERBOSE = 3
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -344,16 +344,8 @@ class Constants(object):
         MAX_OS_MAJOR_VERSION_SUPPORTED = 18
         MINIMUM_CLIENT_VERSION = "27.14.4"
 
-    BUFFER_LINE_SEPERATOR = "\n|\t"
-    BUFFER_LINE_SEPERATOR_LENGTH = len(BUFFER_LINE_SEPERATOR)
-
     class BufferMessage(EnumBackport):
         TRUE = 0
         FALSE = 1
         FLUSH = 2
 
-    class MessageLevel(EnumBackport):  
-        ERROR = 0
-        WARNING = 1
-        DEBUG = 2
-        VERBOSE = 3
--- a/src/core/src/local_loggers/CompositeLogger.py
+++ b/src/core/src/local_loggers/CompositeLogger.py
@@ -34,6 +34,9 @@ class CompositeLogger(object):
         self.TELEMETRY_LOG = "TELEMETRY_LOG:"
         self.current_env = current_env
         self.NEWLINE_REPLACE_CHAR = " "
+        self.MESSAGE_MAX_CHARS = Constants.TELEMETRY_MSG_SIZE_LIMIT_IN_CHARS - Constants.TELEMETRY_BUFFER_FOR_DROPPED_COUNT_MSG_IN_CHARS - Constants.TELEMETRY_EVENT_COUNTER_MSG_SIZE_LIMIT_IN_CHARS
+        self.buffer_message = ""
+
 
     def log(self, message, message_type=Constants.TelemetryEventLevel.Informational):
         """log output"""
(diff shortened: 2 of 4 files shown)


src/core/src/local_loggers/CompositeLogger.py
```diff
@@ -34,6 +34,9 @@ def __init__(self, env_layer=None, file_logger=None, current_env=None, telemetry
         self.TELEMETRY_LOG = "TELEMETRY_LOG:"
         self.current_env = current_env
         self.NEWLINE_REPLACE_CHAR = " "
+        self.MESSAGE_MAX_CHARS = Constants.TELEMETRY_MSG_SIZE_LIMIT_IN_CHARS - Constants.TELEMETRY_BUFFER_FOR_DROPPED_COUNT_MSG_IN_CHARS - Constants.TELEMETRY_EVENT_COUNTER_MSG_SIZE_LIMIT_IN_CHARS
```
SathishMSFT, line 37: This could be moved to constants too.

src/core/src/local_loggers/CompositeLogger.py
```diff
@@ -48,32 +51,74 @@ def log(self, message, message_type=Constants.TelemetryEventLevel.Informational)
...
+            self.flush_error(message)
+        elif message_level == Constants.MessageLevel.WARNING:
+            self.flush_warning(message)
+        elif message_level == Constants.MessageLevel.DEBUG:
+            self.flush_debug(message)
+        elif message_level == Constants.MessageLevel.VERBOSE:
+            self.flush_verbose(message)
+        self.buffer_message = ""
+
+    def buffer_or_flush_message(self, message, buffer, message_level):
```
SathishMSFT, line 65: What if, i keep calling log_error, log_warning etc.. with enum set to true? how will you differentiate this?

src/core/src/bootstrap/Constants.py
```diff
@@ -335,3 +335,16 @@ class UbuntuProClientSettings(EnumBackport):
         MAX_OS_MAJOR_VERSION_SUPPORTED = 18
         MINIMUM_CLIENT_VERSION = "27.14.4"
 
+    BUFFER_LINE_SEPERATOR = "\n|\t"
+    BUFFER_LINE_SEPERATOR_LENGTH = len(BUFFER_LINE_SEPERATOR)
+
+    class BufferMessage(EnumBackport):
```
SathishMSFT, line 341: ReName this appropriately. May be BufferFlushStatus or something similar.

src/core/src/bootstrap/Constants.py
```diff
@@ -335,3 +335,16 @@ class UbuntuProClientSettings(EnumBackport):
...
+        TRUE = 0
+        FALSE = 1
+        FLUSH = 2
+
+    class MessageLevel(EnumBackport):
```
SathishMSFT, line 346: Why not re-use the Telemetry log level enum? I don`t see a need to add this,

src/core/src/local_loggers/CompositeLogger.py
```diff
@@ -48,32 +51,74 @@ def log(self, message, message_type=Constants.TelemetryEventLevel.Informational)
             timestamp = self.env_layer.datetime.timestamp()
             self.file_logger.write("\n" + timestamp + "> " + message.strip(), fail_silently=False)
 
-    def log_error(self, message):
-        """log errors"""
+    def flush(self, message, message_level):
+        if message_level == Constants.MessageLevel.ERROR:
```
SathishMSFT, line 55: This is not needed if you re-use telemetry log level. You could simple call log method with appropriate enum here.

src/core/src/local_loggers/CompositeLogger.py
```diff
@@ -48,32 +51,74 @@ def log(self, message, message_type=Constants.TelemetryEventLevel.Informational)
...
+            self.flush_debug(message)
+        elif message_level == Constants.MessageLevel.VERBOSE:
+            self.flush_verbose(message)
+        self.buffer_message = ""
+
+    def buffer_or_flush_message(self, message, buffer, message_level):
         message = self.__remove_substring_from_message(message, Constants.ERROR_ADDED_TO_STATUS)
+
+        if buffer == Constants.BufferMessage.FALSE:
+            if len(self.buffer_message) > 0:
```
SathishMSFT, line 69: You could simple append to buffer and do one flush here. Not need to check size of the buffer

src/core/src/local_loggers/CompositeLogger.py
```diff
@@ -34,6 +34,9 @@ def __init__(self, env_layer=None, file_logger=None, current_env=None, telemetry
         self.TELEMETRY_LOG = "TELEMETRY_LOG:"
         self.current_env = current_env
         self.NEWLINE_REPLACE_CHAR = " "
+        self.MESSAGE_MAX_CHARS = Constants.TELEMETRY_MSG_SIZE_LIMIT_IN_CHARS - Constants.TELEMETRY_BUFFER_FOR_DROPPED_COUNT_MSG_IN_CHARS - Constants.TELEMETRY_EVENT_COUNTER_MSG_SIZE_LIMIT_IN_CHARS
+        self.buffer_message = ""
```
SathishMSFT, line 38: Rename this to just buffer

src/core/src/local_loggers/CompositeLogger.py
SathishMSFT: Adding to previous point, this will make this option suitable for FLUSH enum type too. So for both options FLUSH and FALSE, you will append to buffer and write. For option TRUE, you can just store in buffer.

src/core/src/local_loggers/CompositeLogger.py
GAURAVRAMRAKHYANI: sure, will move to constants

src/core/src/local_loggers/CompositeLogger.py
GAURAVRAMRAKHYANI: As discussed offline, the log_error, log_warning, etc. functions were doing more preprocessing in the message string. I have renamed that functions to flush_error, flush_warning, etc. and calling them when flush is required.

kjohn-msft:
After some thought on this, I’m refining the [last guidance on this PR](https://github.com/Azure/LinuxPatchExtension/pull/211#issuecomment-1658599973) to:
-	Keep separation of concerns
o	CompositeLogger does not need to know about buffering except to the extent that it needs to pass it through.
o	TelemetryWriter needs to know about buffering because it’s an implementation detail it needs to worry about.
-	Keep it simple – what’s the minimally-invasive change in TelemetryWriter needed? (see detail)
-	Corner case – there’s a message level change case that wasn’t considered (assume calling code is buggy for robustness in implementation).

In detail:
-	Constants
o	The only constant here will be BufferMessage (as it’s used in multiple classes).
o	Internal constants should be left within respective classes where possible to avoid main Constants bloat.
-	CompositeLogger needs to pass through ‘Buffer’ to TelemetryWriter in 2 functions (change in CL restricted to pass through – 2 incoming function signatures, 2 outbound function calls):
o	log
o	log_debug
o	log_error, log_warning shouldn’t use buffering – no change here with default buffer parameter on log.
o	log_verbose doesn’t write to telemetry.
-	TelemetryWriter needs to implement a new function:
o	write_event_with_buffer(<same params as write event>, buffer_msg)
	This will internally abstract buffer functionality – this is the bulk of the change, neatly isolated and abstracted.
•	Constants / Variables (using correct const casing where const --): 
o	(string) telemetry_buffer_store 
o	(Constants.TelemetryEventLevel) last_telemetry_event_level
o	(string, const) telemetry_buffer_delimiter
•	buffer_msg == true and (last_telemetry_event_level == current or last_telemetry_event_level == None)
o	append delimiter + msg to internal buffer.
•	buffer_msg == false or (last_telemetry_event_level != current)
o	append latest message to buffer
o	flush buffer
•	buffer_msg == flush
o	flush buffer. Reset last_telemetry_event_level
	It will call write_event internally as needed (no change on write-event).
o	TelemetryWriter will continue to internally take care of wrapping, etc. There’s no need to worry about buffer overrun.
-	Unit tests
o	Unit tests can focus heavily on write_event_with_buffer in TelemetryWriter with basic coverage beyond that in CompositeLogger UTs.

Things to remove from PR to strip out unnecessary complexity:
-	Other constants like message level introduced to constants that were an indication of conflict between CompositeLogger and TelemetryWriter.
-	Simplify CompositeLogger 
o	It mustn’t buffer, should write to disk immediately. TelemetryWriter should where appropriate.
o	All the complexity around new message levels and mappings can all be removed.
o	Duplication of message length cuts, etc. is unnecessary as it’s already fully handled in TW.

kjohn-msft (dismissed):
Tentative approval on test coverage with product code as in this iteration.

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -156,6 +160,38 @@ def __ensure_message_restriction_compliance(self, full_message):
             self.composite_logger.log_telemetry_module_error("Error occurred while formatting message for a telemetry event. [Error={0}]".format(repr(e)))
             raise
 
+    def write_event_with_buffer(self, message, event_level, buffer_msg):
+        if buffer_msg == Constants.BufferMessage.TRUE and (event_level == self.last_telemetry_event_level or self.last_telemetry_event_level is None):
+            if self.telemetry_buffer_store != "":
+                self.telemetry_buffer_store = self.telemetry_buffer_store + self.TELEMETRY_BUFFER_DELIMETER + message
+            else:
```
rane-rajasi, line 167: Is there a plan to add these UTs later?

rane-rajasi (commented):
Left one question inline. Outside of that, changes look good

src/core/src/service_interfaces/TelemetryWriter.py
GAURAVRAMRAKHYANI: Added UTs.

kjohn-msft (approved):

rane-rajasi (approved):

PR #216, opened 2023-09-06
APT: Patch Mode set to disabled in ConfigurePatching when os patch config file does not exist

rane-rajasi:
Issue:
- ConfigurePatching is marked as failed if '/etc/apt/apt.conf.d/20auto-upgrades' is not found. 

'/etc/apt/apt.conf.d/20auto-upgrades' not existing on a VM has been observed to be an indication that default auto updates are not configured. The config file is only created the very first time auto updates are configured on the machine.  Hence, non existence of the file can be safely considered as default auto updates are disabled. This also aligns with the implementation patterns used in YUM and Zypper package managers where a config not existing is considered as default updates disabled.

Desired behavior from this PR:
If '/etc/apt/apt.conf.d/20auto-upgrades' is not found:
- automaticOSPatchState in Configure Patching is marked as Disabled
- ConfigurePatching status does not fail due to this config not existing

### DIFF
--- a/src/core/src/package_managers/AptitudePackageManager.py
+++ b/src/core/src/package_managers/AptitudePackageManager.py
@@ -476,8 +476,9 @@ class AptitudePackageManager(PackageManager):
     def get_current_auto_os_patch_state(self):
         """ Gets the current auto OS update patch state on the machine """
         self.composite_logger.log("Fetching the current automatic OS patch state on the machine...")
-        self.__get_current_auto_os_updates_setting_on_machine()
-        if int(self.unattended_upgrade_value) == 0:
+        if os.path.exists(self.os_patch_configuration_settings_file_path):
+            self.__get_current_auto_os_updates_setting_on_machine()
+        if not os.path.exists(self.os_patch_configuration_settings_file_path) or int(self.unattended_upgrade_value) == 0:
             current_auto_os_patch_state = Constants.AutomaticOSPatchStates.DISABLED
         elif int(self.unattended_upgrade_value) == 1:
             current_auto_os_patch_state = Constants.AutomaticOSPatchStates.ENABLED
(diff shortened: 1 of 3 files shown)


kjohn-msft:
Please fill in description to describe current behavior and desired behavior.

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -476,8 +476,9 @@ def get_package_size(self, output):
     def get_current_auto_os_patch_state(self):
         """ Gets the current auto OS update patch state on the machine """
         self.composite_logger.log("Fetching the current automatic OS patch state on the machine...")
-        self.__get_current_auto_os_updates_setting_on_machine()
-        if int(self.unattended_upgrade_value) == 0:
+        if os.path.exists(self.os_patch_configuration_settings_file_path):
+            self.__get_current_auto_os_updates_setting_on_machine()
+        if not os.path.exists(self.os_patch_configuration_settings_file_path) or int(self.unattended_upgrade_value) == 0:
             current_auto_os_patch_state = Constants.AutomaticOSPatchStates.DISABLED
```
kjohn-msft, line 482: Verified on a production machine?

kjohn-msft (changes requested):
LGTM, minor questions inline

rane-rajasi:
> Please fill in description to describe current behavior and desired behavior.

done

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: Validated on customer test machines and on our own test machines

kjohn-msft (approved):

feng-j678 (approved):

PR #234, opened 2024-01-29
Fix/unicode in py27 utests

feng-j678:
with the fix for unicodeerror using euro symbol in unit test, py27 will throw the below error when running utest locally because locally py27 will not auto recognize/encode those char as in py3,  though the cicd pipeline will not throw any errors and process the unit test as expected, see pictures below:

image

cicd pipeline before code change using the pr1reduce-status_file as example since it has master code merged in :
image

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -29,6 +29,4 @@ coverage:
         if_not_found: failure
     ignore:
       - "*/tests/*"
-      - "*_test.*"
-      - "*test_*"
-      - "Test_*"
+      - "*/tests/Test_"
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -27,5 +27,5 @@ coverage:
         threshold: 0%
         if_ci_failed: error
         if_not_found: failure
-    ignore:
-      - "**/tests/*"
+ignore:
+  - "src/core/tests/*"
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -27,6 +27,3 @@ coverage:
         threshold: 0%
         if_ci_failed: error
         if_not_found: failure
-    ignore:
-      - "*/tests/*"
-      - "*/tests/Test_"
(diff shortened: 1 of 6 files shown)


src/core/tests/Test_TelemetryWriter.py
```diff
@@ -101,21 +103,15 @@ def test_write_event_msg_size_limit(self):
             self.assertTrue("a"*(len(message.encode('utf-8')) - chars_dropped) + ". [{0} chars dropped]".format(chars_dropped) in events[-1]["Message"])
             f.close()
 
-    def test_write_event_msg_size_limit_char_more_than_1_bytes(self):
-        """ Perform 1 byte truncation on char that is more than 1 byte, use decode('utf-8', errors='replace') to replace bad unicode with a good 1 byte char (�) """
+    def test_write_event_msg_size_limit_char_with_actual_unicode_str(self):
+        """ Perform 1 byte truncation on unicode str that is more than 1 byte, use decode('utf-8', errors='replace') to replace bad unicode with a good 1 byte char (�) """
+        message = u"a€bc"*3074  # €(\xe2\x82\xac) is 3 bytes char can be written in windows console w/o encoding
+        self.__assert_write_event_msg_size_limit_char(message)
```
kjohn-msft, line 109: Why is it saying all of this isn't covered by tests? Can you investigate this and all the other flagging here?

.github/codecov.yml
```diff
@@ -26,4 +26,7 @@ coverage:
         target: 100%
         threshold: 0%
         if_ci_failed: error
-        if_not_found: failure
\ No newline at end of file
+        if_not_found: failure
+    ignore:
+      - "*/tests/*"
+      - "*/tests/Test_"
```
kjohn-msft, line 32: Why is this here? Test code will auto-execute and automatically count to coverage. It will also help flag if any test is not running for any reason, or if any code path in a test is dead-code.

If anything, things to be excluded from coverage should be non-code sections like the tools folder.

PR #242, opened 2024-02-08
Pr1.3/truncation conditonal checker

feng-j678:
[x] add conditional checker logic to delay truncation and reduce truncation overhead - when truncation timestamp tracker outside the 60 sec mark, truncation will apply on upcoming operation else does not, a terminal state (success or error) overwrites this logic and still apply truncation if status file limits > 126kb
[x] unit test

### DIFF
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 #
 # Requires Python 2.7+
+import datetime
 
 
 class Constants(object):
@@ -184,6 +185,7 @@ class Constants(object):
         TRUNCATION_WARNING_MESSAGE = "Package lists were truncated to limit reporting data volume. In-VM logs contain complete lists."
         TURN_ON_TRUNCATION = True
         MIN_TRUNCATION_INTERVAL_IN_SEC = 60
+        EPOCH = datetime.datetime(1971, 1, 1, 0, 0, 0)
 
     # Wrapper-core handshake files
     EXT_STATE_FILE = 'ExtState.json'
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -300,7 +300,7 @@ class Constants(object):
 
     TELEMETRY_NOT_COMPATIBLE_ERROR_MSG = "Unsupported older Azure Linux Agent version. To resolve: http://aka.ms/UpdateLinuxAgent"
     TELEMETRY_COMPATIBLE_MSG = "Minimum Azure Linux Agent version prerequisite met"
-
+    PYTHON_NOT_COMPATIBLE_ERROR_MSG = "Unsupported older Python version. Minimum Python version required is 2.7. [DetectedPythonVersion={0}]"
     UTC_DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
 
     # EnvLayer Constants
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -30,7 +30,7 @@ class Constants(object):
     UNKNOWN = "Unknown"
 
     # Extension version (todo: move to a different file)
-    EXT_VERSION = "1.6.48"
+    EXT_VERSION = "1.6.49"
 
     # Runtime environments
     TEST = 'Test'
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -30,7 +30,7 @@ class Constants(object):
     UNKNOWN = "Unknown"
 
     # Extension version (todo: move to a different file)
-    EXT_VERSION = "1.6.47"
+    EXT_VERSION = "1.6.48"
 
     # Runtime environments
     TEST = 'Test'
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -30,7 +30,7 @@ class Constants(object):
     UNKNOWN = "Unknown"
 
     # Extension version (todo: move to a different file)
-    EXT_VERSION = "1.6.49"
+    EXT_VERSION = "1.6.50"
 
     # Runtime environments
     TEST = 'Test'
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -201,6 +201,7 @@ class Constants(object):
     MAX_IMDS_CONNECTION_RETRY_COUNT = 5
     MAX_ZYPPER_REPO_REFRESH_RETRY_COUNT = 5
     MAX_BATCH_SIZE_FOR_PACKAGES = 3
+    MAX_COMPLETE_STATUS_FILES_TO_RETAIN = 10
 
     class PackageClassification(EnumBackport):
         UNCLASSIFIED = 'Unclassified'
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -51,8 +51,11 @@ class Constants(object):
     MAX_AUTO_ASSESSMENT_LOGFILE_SIZE_IN_BYTES = 5*1024*1024
     MAX_AUTO_ASSESSMENT_WAIT_FOR_MAIN_CORE_EXEC_IN_MINUTES = 3 * 60
 
-    class Paths(EnumBackport):
+    class SystemPaths(EnumBackport):
         SYSTEMD_ROOT = "/etc/systemd/system/"
+
+    class AzGPSPaths(EnumBackport):
+        EULA_SETTINGS = "/var/lib/azure/linuxpatchextension/patch.eula.settings"
 
     class EnvSettings(EnumBackport):
         LOG_FOLDER = "logFolder"
@@ -76,6 +79,11 @@ class Constants(object):
         PATCH_MODE = 'patchMode'
         ASSESSMENT_MODE = 'assessmentMode'
         MAXIMUM_ASSESSMENT_INTERVAL = 'maximumAssessmentInterval'
+
+    class EulaSettings(EnumBackport):
+        ACCEPT_EULA_FOR_ALL_PATCHES = 'AcceptEULAForAllPatches'
+        ACCEPTED_BY = 'AcceptedBy'
+        LAST_MODIFIED = 'LastModified'
 
     TEMP_FOLDER_DIR_NAME = "tmp"
     TEMP_FOLDER_CLEANUP_ARTIFACT_LIST = ["*.list"]
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -208,7 +208,8 @@ class Constants(object):
     MAX_INSTALLATION_RETRY_COUNT = 3
     MAX_IMDS_CONNECTION_RETRY_COUNT = 5
     MAX_ZYPPER_REPO_REFRESH_RETRY_COUNT = 5
-    MAX_BATCH_SIZE_FOR_PACKAGES = 3
+    MAX_BATCH_SIZE_FOR_PACKAGES = 6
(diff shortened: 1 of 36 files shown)


src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -90,6 +91,11 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
 
         self.__current_operation = None
 
+        self.__truncation_timestamp_tracker = None  # To allow truncation on upcoming operation when tracker > 60 sec (tentatively)
```
kjohn-msft, line 94: We had discussed this to be self.__truncation_timestamp yesterday

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -90,6 +91,11 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
...
+        self.__truncated_status_file_json_dumps = None  # To keep status file truncated (not overwritten) when tracker timestamp < 60 sec (tentatively)
+        self.__force_truncation_first_time = True  # Apply truncation only once initially tracker timestamp will be < 60 sec
```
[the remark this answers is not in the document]
feng-j678: hmm I see you are suggesting to use timestamp as conditional check for first time truncation, but this function is called before the condition check, the timestamp will be initialized, so i can't use it in condition check for first time truncation because it is always initialized 
image
image

src/core/src/service_interfaces/StatusHandler.py
feng-j678: removed tracker, and other reference
done

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -333,6 +339,9 @@ def set_assessment_substatus_json(self, status=Constants.STATUS_TRANSITIONING, c
         # Wrap assessment summary into assessment substatus
         self.__assessment_substatus_json = self.__new_substatus_json_for_operation(Constants.PATCH_ASSESSMENT_SUMMARY, status, code, json.dumps(self.__assessment_summary_json))
 
+        # Set force truncation true when final status is success or error
+        self.__set_force_truncation_on_terminal_status(substatus_status=status)
```
rane-rajasi, line 343: Point to note: Force truncation is being set on the basis of a single substatus's status. It is not checking/looking for terminal status within all substatuses's

For eg: at this point in code, PatchAssessmentSummary has a terminal status but PatchInstallationSummary is still transitioning. This code will apply truncation.

Noting it down to ensure this is the desired behavior.

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -853,18 +876,48 @@ def get_num_installation_patches_removed(self):
...
         """ Get truncated status file payload when status file byte size is more than 126kb """
         status_file_size_in_bytes = self.__calc_status_size_on_disk(status_file_payload_json_dumps)  # calc complete_status_file_payload_json byte size on disk
 
         if status_file_size_in_bytes > Constants.StatusTruncationConfig.INTERNAL_FILE_SIZE_LIMIT_IN_BYTES:  # perform truncation complete_status_file byte size > 126kb
-            truncated_status_file = self.__create_truncated_status_file(status_file_size_in_bytes, status_file_payload_json_dumps)
-            status_file_payload_json_dumps = json.dumps(truncated_status_file)
+            is_truncation_allowed = self.__calc_truncation_timestamp_in_sec() > Constants.StatusTruncationConfig.NO_TRUNCATION_LOGIC_IN_SEC
+
+            if self.__force_truncation_first_time or is_truncation_allowed or self.__force_truncation_on:
+                self.composite_logger.log_verbose("[IsTruncationAllowed={0}] [IsForceTruncationOn={1}]".format(str(is_truncation_allowed), str(self.__force_truncation_on)))
```
rane-rajasi, line 894: Log self.__force_truncation_first_time too

src/core/src/service_interfaces/StatusHandler.py
```diff
@@ -853,18 +876,48 @@ def get_num_installation_patches_removed(self):
...
+                if is_truncation_allowed:
+                    self.__truncation_timestamp = datetime.datetime.now()  # Set timestamp to newer time
+
+                if self.__force_truncation_first_time:
+                    self.__force_truncation_first_time = False  # Set to False because first time truncation applied during timestamp < 60 sec
+
+                return self.__truncated_status_file_json_dumps
+
+            # To ensure all newly status file remain in truncated state when complete statusfile > 126kb and timestamp < 60 sec
+            if self.__truncated_status_file_json_dumps is not None and not is_truncation_allowed and not self.__force_truncation_on:
```
rane-rajasi, line 907: This should be else if:


Code can be optimized to:

if self.__force_truncation_first_time or is_truncation_allowed or self.__force_truncation_on:
# truncate status, set self.__truncation_timestamp  and reset self.__force_truncation_first_time. Store truncated status into status_file_payload_json_dumps

else if: self.__truncated_status_file_json_dumps is not None
 status_file_payload_json_dumps = self.__truncated_status_file_json_dumps

return status_file_payload_json_dumps at the end of the function

src/core/src/bootstrap/Constants.py
```diff
@@ -183,6 +183,7 @@ class StatusTruncationConfig(EnumBackport):
         MIN_ASSESSMENT_PATCHES_TO_RETAIN = 5
         TRUNCATION_WARNING_MESSAGE = "Package lists were truncated to limit reporting data volume. In-VM logs contain complete lists."
         TURN_ON_TRUNCATION = True
+        NO_TRUNCATION_LOGIC_IN_SEC = 60
```
rane-rajasi, line 186: rename to TIME_BETWEEN_TRUNCATION_IN_SEC or DELAY_TRUNCATION_BY_SEC

rane-rajasi (changes requested):
Comments inline

src/core/src/service_interfaces/StatusHandler.py
feng-j678: image

src/core/tests/Test_StatusHandlerTruncation.py
```diff
@@ -37,6 +37,7 @@ def setUp(self):
         self.__test_scenario = None
         self.__patch_count_assessment = 0
         self.__patch_count_installation = 0
+        Constants.StatusTruncationConfig.NO_TRUNCATION_LOGIC_IN_SEC = -1  # ignore the cool time 60 sec, ut execution time is linear
```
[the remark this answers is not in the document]
feng-j678: my point of view on the delay is we apply truncation > x_Sec, so reduce the x_sec to -1 and because the difference datetime in sec in below method is very small close to 0 like 0.008msec, but we are more than x_sec timeframe then we should see truncated result, which we do in all truncated status ut. modifying the constant value, we are testing the delay logic. let me know if you have a diffrence idea on testing this. 
modification
image

src/core/src/service_interfaces/StatusHandler.py
feng-j678: done

src/core/src/service_interfaces/StatusHandler.py
feng-j678: I immediately return the truncated_status_json_dumps in else block to avoid more code and variable assignment. 
is this okay?
image


done

src/core/src/bootstrap/Constants.py
feng-j678: hmm i changed the constant to that, but here it is showing the outdated var name:

image

rane-rajasi: DELAY_TRUNCATION_BY_SEC is the updated name correct? what is the outdated var name then?

feng-j678: the outdated var name was NO_TRUNCATION_LOGIC_IN_SEC

src/core/src/bootstrap/Constants.py
```diff
@@ -183,6 +183,7 @@ class StatusTruncationConfig(EnumBackport):
         MIN_ASSESSMENT_PATCHES_TO_RETAIN = 5
         TRUNCATION_WARNING_MESSAGE = "Package lists were truncated to limit reporting data volume. In-VM logs contain complete lists."
         TURN_ON_TRUNCATION = True
+        DELAY_TRUNCATION_BY_SEC = 60
```
kjohn-msft, line 186: MIN_TRUNCATION_INTERVAL_IN_SEC

why - the variable as it is named now sounds like a one time delay. And it's not a constant delay, it's the minimum interval of time that must pass before truncation can happen normally after the last time (exceptions around terminal states exist of course)

src/core/tests/Test_StatusHandlerTruncation.py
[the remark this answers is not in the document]
rane-rajasi: We should have a UT to validate the code change being introduced in this PR. i.e. a scenario where truncation is applied and status is not updated when the time difference with an earlier truncation is < 60 sec.

We do not have any UT currently to validate this scenario

src/core/src/service_interfaces/StatusHandler.py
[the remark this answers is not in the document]
kjohn-msft: This dilemma happens because setting the timestamp to "now" when you see it to be None is incorrect. 

The timestamp should only be set to a time that truncation actually happened.

At the point you are initializing it because it is None, you should initialize it to "Min timestamp" - usually epoch - that's so far in the past that it's obvious the first-time case applies. And this additional variable for first time is completely unnecessary.

kjohn-msft (changes requested):
Inline

src/core/src/bootstrap/Constants.py
feng-j678: done

src/core/src/service_interfaces/StatusHandler.py
feng-j678: add ut for scenario, use the minimum timestamp 2 sec (delay) and reduce MIN_TRUNCATION_INTERVAL_IN_SEC to 2sec
done

src/core/tests/Test_StatusHandlerTruncation.py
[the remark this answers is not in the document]
feng-j678: add ut for scenario, use the minimum timestamp 2 sec (delay) and reduce MIN_TRUNCATION_INTERVAL_IN_SEC to 2sec
done

src/core/src/service_interfaces/StatusHandler.py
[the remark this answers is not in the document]
feng-j678: change applied,
done

kjohn-msft (dismissed):

kjohn-msft (approved):

src/core/tests/Test_StatusHandlerTruncation.py
```diff
@@ -613,6 +615,95 @@ def test_both_assessment_and_installation_with_informational_truncated(self):
...
+        # Assert startTime and lastModifiedTime are equal - first time truncated and remained truncated status file
+        remained_truncated_assessment_msg = self.__get_message_json_from_substatus(remained_truncated_substatus_file)
+        remained_truncated_installation_msg = self.__get_message_json_from_substatus(remained_truncated_substatus_file, installation_substatus_index=1)
+
+        self.assertEqual(first_truncated_assessment_msg['startTime'], remained_truncated_assessment_msg['startTime'])
+        self.assertEqual(first_truncated_assessment_msg['lastModifiedTime'], remained_truncated_assessment_msg['lastModifiedTime'])
+        self.assertEqual(first_truncated_installation_msg['startTime'], remained_truncated_installation_msg['startTime'])
+        self.assertEqual(first_truncated_installation_msg['lastModifiedTime'], remained_truncated_installation_msg['lastModifiedTime'])
+
+        Event().wait(1)  # Use event to pause ut execution for 1 sec, so we will get new lastModifiedTime, this is terminal status truncation
```
rane-rajasi, line 684: Why this wait? Wouldn't we get lastModifiedTime from the status file?

rane-rajasi (approved):

PR #243, opened 2024-02-22
improve unit test time performance

feng-j678:
[x]reduce patches to 10000 to improve ut time performance
[x]prior change total time is 1m3secish
[x]post change total time is 37secish
[x]post change test_statusfiletruncation time is 10secish


src/core/tests/Test_StatusHandlerTruncation.py
```diff
@@ -89,7 +89,7 @@ def test_only_assessment_patches_over_size_limit_truncated(self):
         truncated status file byte size: 126kb. """
 
         self.__test_scenario = 'assessment_only'
-        self.__patch_count_assessment = 100000
+        self.__patch_count_assessment = 10000
```
kjohn-msft, line 92: Why isn't this part of the configuration?

feng-j678: because the input patches size varies by unit test by scenarios and that's a class level var will be used in assertion logic

rane-rajasi (approved):

kjohn-msft (approved):

PR #244, opened 2024-02-27
Bug fix: Auto assessment attempting to update healthstore

rane-rajasi:
Auto assessment attempts to update health store status, since health_store_id is not reset to None when along with other configs for an auto assess operation

### DIFF
--- a/src/core/tests/Test_CoreMain.py
+++ b/src/core/tests/Test_CoreMain.py
@@ -828,7 +828,7 @@ class TestCoreMain(unittest.TestCase):
     def test_auto_assessment_success_with_installation_in_prev_operation_on_same_sequence(self):
         """Unit test for auto assessment request with installation (Auto Patching) completed on the sequence before.
         Result: should contain PatchAssessmentSummary with an updated timestamp after auto assessment, and retain PatchInstallationSummary, ConfigurePatchingSummary and PatchMetadatForHealthStoreSummary from the installation(Auto Patching) operation"""
-        # operation #1: Assessment
+        # operation #1: Installation
         argument_composer = ArgumentComposer()
         argument_composer.operation = Constants.INSTALLATION
         argument_composer.maintenance_run_id = "8/27/2021 02:00:00 PM +00:00"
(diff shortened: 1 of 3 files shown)


src/core/tests/Test_StatusHandler.py
```diff
@@ -64,6 +64,19 @@ def test_set_package_assessment_status_for_auto_assessment(self):
...
+        self.runtime.execution_config.exec_auto_assess_only = True
+        self.runtime.execution_config.maintenance_run_id = None
+        self.runtime.execution_config.health_store_id = "test"
+        self.assertRaises(Exception, lambda: StatusHandler(self.runtime.env_layer, self.runtime.execution_config, self.runtime.composite_logger, self.runtime.telemetry_writer, self.runtime.vm_cloud_type))
+
+        # test with maintenance_run_id not None
+        self.runtime.execution_config.exec_auto_assess_only = True
+        self.runtime.execution_config.maintenance_run_id = "test"
+        self.runtime.execution_config.health_store_id = None
+        self.assertRaises(Exception, lambda: StatusHandler(self.runtime.env_layer, self.runtime.execution_config, self.runtime.composite_logger, self.runtime.telemetry_writer, self.runtime.vm_cloud_type))
```
kjohn-msft, line 78: It should no longer throw an exception because there's nothing to write?

kjohn-msft: Your test should primarily be having **both** maintenance_run_id and health_store_id set to None to reflect what auto_assessment runs as - and that should **not** throw an exception. These other two cases are fine (should throw an exception). It's better for it to be two tests - one that you have now for the negative case, and then the positive case reflecting the fix made here.

kjohn-msft (changes requested):
Comments inline

src/core/tests/Test_StatusHandler.py
rane-rajasi: Default values for both of these is None in ArgumentComposer, which are then used in this (success) test test_set_package_assessment_status_for_auto_assessment()

kjohn-msft: Superficially, if an existing positive test covered this correctly, that test should have failed prior to the product code change (to match what happened in production). Right?

rane-rajasi: Added a positive test in Test_CoreMain to verify maintenance_run_id and health_store_id are reset by code to None. 

That test almost replicates real time behavior, Installation runs initially -> outputs a status, Auto Assessment comes in with the same args as Installation, resets these 2 args to None and updates assessment substatus

rane-rajasi: Status: Comment addressed

src/core/tests/Test_CoreMain.py
```diff
@@ -906,6 +906,88 @@ def test_auto_assessment_success_with_installation_in_prev_operation_on_same_seq
...
+        # check status file
+        with runtime.env_layer.file_system.open(runtime.execution_config.status_file_path, 'r') as file_handle:
+            status_file_data = json.load(file_handle)[0]["status"]
+        self.assertTrue(status_file_data["operation"] == Constants.INSTALLATION)
+        substatus_file_data = status_file_data["substatus"]
+        self.assertEqual(len(substatus_file_data), 4)
+        self.assertTrue(substatus_file_data[0]["name"] == Constants.PATCH_ASSESSMENT_SUMMARY)
+        self.assertTrue(substatus_file_data[0]["status"].lower() == Constants.STATUS_SUCCESS.lower())
+        # check started by set to 'User'
+        self.assertTrue(json.loads(substatus_file_data[0]["formattedMessage"]["message"])['startedBy'], Constants.PatchAssessmentSummaryStartedBy.USER)
```
feng-j678, line 930: please use assertEquals if we are comparing two objects, this way if test failed it output the expected value and actual value instead of true or false

feng-j678 (changes requested):

src/core/tests/Test_CoreMain.py
rane-rajasi: assertTrue is being used in other similar checks all throughout our test suite. This was a conscious decision made long back simce AssertEquals() is deprecated in Python 3x. The choice was between using assertEqual vs assertTrue back then

kjohn-msft (approved):

PR #245, opened 2024-02-28
Remove external dependencies from coverage, increase default target

kjohn-msft:
We should not be measuring coverage on external dependencies that we have to ship.

And in parallel increasing the default target to the base threshold we've already achieved.

image


kjohn-msft:
Coverage increased by 0.73% -- 
image

feng-j678 (approved):

PR #248, opened 2024-03-07
Fix the issue that if there are conflicting packages in the yum install command then installation of all the packages in the command fails.

GAURAVRAMRAKHYANI:
There are two issues:
(a) In the method add_arch_dependencies, there are conflicting packages added in the same yum install command.
(b) If there are conflicting packages in the yum install command then installation of all the packages in the command fails.

Issue:
If there are two updates available in yum with same name but different architecture and different versions
then if they both are attempted to be installed in single command then install fails due to conflict.

For example following are the packages:
subscription-manager-rhsm-certificates-1.28.36-3.el8_8.x86_64
subscription-manager-rhsm-certificates-20220623-1.el8.noarch

Installing both these packages in same command fails with error in the below screenshot.
Here, the package name is same for both the packages: subscription-manager-rhsm-certificates
Architecture is different: x86_64 and noarch
Version is also different: 1.28.36-3 and 20220623-1

The installation would succeed if version is same and only architecture is different.

This issue is not new in parallel patching, this issue should be there before parallel patching was implemented because in yum package manager, we check the packages with same name but different architecture and install them in single commands. We don't check for their versions.

Fix
If there are two packages with same name but different architecture, we should verify that the version of the packages should be same before adding them as dependent package.

The above fix will only fix the case when we are adding the second package with same name as arch dependency in the function add_arch_dependencies. This will not fix the case when both packages are added in the same batch outside the function add_arch_dependencies.
Added --skip-broken parameter in the yum install command to resolve this issue. If there are conflicting packages in the batch and --skip-broken is mentioned, then yum will resolve the conflict and install one of the conflicting package as part of batch patching and skip installation of the other package. This will avoid failure of all the packages in the batch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,9 +11,9 @@ jobs:
       PYTHONTRACEMALLOC: 1
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       - name: Set up Python 3.9
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: 3.9
       - name: Install dependencies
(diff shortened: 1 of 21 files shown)


SathishMSFT (approved):

kjohn-msft (approved):

kjohn-msft:
Please wait for @rane-rajasi 's review / approval to merge

rane-rajasi (approved):

PR #250, opened 2024-03-13
Calculate maximum batch size in batch patching as per time remaining in maintenance window

GAURAVRAMRAKHYANI:
There are four changes at high level:
(a) Calculate maximum batch size in batch patching as per time remaining in maintenance window. The upper limit of max batch size is 300.

(b) If there are some remaining packages after batch patching due to failures in batch patching or maintenance window cutoff exceeded, then do batch patching again with lower batch size (1/10th of original batch size) instead of directly falling back to sequential patching. In case there is again remaining packages after batch patching with lower batch size then do sequential patching.
The code changes to calculate batch size in phase1 and phase2 will be done in later iterations once the code structure is approved.

(c) Use different value for average time to install packages for different package managers as per the telemetry data. Using current average time to install packages, it should improve after this change, and we can update the values later as per the improvement.

(d) In the maintenance window cutoff calculation, assumption of the number of packages in the batch which could take max time to install was 3, changing it to 1. Rest of the packages should take average time to install.

(e) In sequential patching, for the packages which failed to install in first attempt are retried 3 times. So, at max, there are total 4 attempts in sequential patching for a package. Now, with multi-phase batch patching implementation, we have two attempts in batch patching and 4 attempts in sequential patching. So, at max, there are total 6 attempts for a package. These are too many attempts. Each attempt to install package requires running two commands, one to install package and the other to check if package is installed successfully. These commands take up to total 1 second per package. These commands generate some logs which will add in telemetry throttling and slow down. To reduce attempts, removing 3 retries in sequential patching. So, there are total 2 attempts in batch patching and 1 in sequential patching.

Only adding function signatures and their caller from the code in the first iteration. Will implement the functions in later iterations.

Testing:
Manual testing is not done as function implementation is pending.
Test changes will be done once the product changes are approved.

Update:
Function implementation for get_max_batch_size is done.
Completed the manual testings. Fixed some issues in the code changes which were found during unit testing.
Unit tests are not done. Will do it once product code changes are reviewed.

Manual testing done:
(a) Done testing in all the 3 package managers: apt, yum and zypper.

(b) Tested the scenario in which multi-phase batch patching is done in only 1 phase and in 2 phases due to different conditions like remaining packages to install greater than 0 after phase 1 and maintenance window batch cutoff reached with phase 1 so need to do phase2. Also, tested that the packages remaining after phase 2 are attempted in sequential patching.

(c) Done testing for the case when number of packages to install is 0 in the beginning of the patch installation job.


src/core/src/core_logic/PatchInstaller.py
```diff
@@ -736,3 +781,6 @@ def filter_out_excluded_updates(self, included_packages, included_package_versio
         self.composite_logger.log_debug(str(len(new_included_packages)) + " out of " + str(len(included_packages)) + " packages will remain included in the run.")
         return new_included_packages, new_included_package_versions
     # endregion
+
+    def get_max_batch_size(self, maintenance_window, package_manager):
```
SathishMSFT, line 785: Where is the definition?

GAURAVRAMRAKHYANI: The definition will be added in later iterations as we discussed offline. Trying to limit the size of the PR so that it is easier to review.

SathishMSFT (dismissed):

src/core/src/core_logic/PatchInstaller.py
```diff
@@ -275,21 +275,74 @@ def install_updates(self, maintenance_window, package_manager, simulate=False):
...
+
         # Check for packages available and install them in batches. Some packages may not be installed due to:
         # (a) Not enough time remaining in maintenance window
         # (b) Failure during package installation
-        installed_update_count, patch_installation_successful, maintenance_window_batch_cutoff_reached, packages, package_versions = self.install_packages_in_batches(
-            all_packages, all_package_versions, packages, package_versions, maintenance_window, package_manager)
+
+        # Start Batch Patching Phase1
+        successful_parent_package_install_count_in_batch_patching_phase1 = 0
+        if max_batch_size_for_packages_phase1 > 0:
```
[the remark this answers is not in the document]
GAURAVRAMRAKHYANI: I have made the changes to address this. Using for loop for this. I am not using list, just using single variable which will be updated in each phase to determine max batch size for the phase.

src/core/src/core_logic/PatchInstaller.py
```diff
@@ -343,16 +372,7 @@ def install_updates(self, maintenance_window, package_manager, simulate=False):
...
-            for i in range(0, Constants.MAX_INSTALLATION_RETRY_COUNT):
-                install_result = package_manager.install_update_and_dependencies_and_get_status(package_and_dependencies, package_and_dependency_versions, simulate)
-
-                if install_result != Constants.INSTALLED:
-                    if i < Constants.MAX_INSTALLATION_RETRY_COUNT - 1:
-                        time.sleep(i + 1)
-                        self.composite_logger.log_warning("Retrying installation of package. [Package={0}]".format(package_manager.get_product_name(package_and_dependencies[0])))
-                else:
-                    break
+            install_result = package_manager.install_update_and_dependencies_and_get_status(package_and_dependencies, package_and_dependency_versions, simulate)
```
rane-rajasi, line 375: IIUC, this is a part of the sequential flow, how and why does that change?

GAURAVRAMRAKHYANI: I forgot to mention regarding this in PR description. Added point (e) in the PR description for this.
I have removed some retries in sequential patching. This is explained below. Please let me know if it is fine to remove the retries.

In sequential patching, for the packages which failed to install in first attempt are retried 3 times. So, at max, there are total 4 attempts in sequential patching for a package. Now, with multi-phase batch patching implementation, we have two attempts in batch patching and 4 attempts in sequential patching. So, at max, there are total 6 attempts for a package. These are too many attempts. Each attempt to install package requires running two commands, one to install package and the other to check if package is installed successfully. These commands take up to total 1 second per package. These commands generate some logs which will add in telemetry throttling and slow down. To reduce attempts, removing 3 retries in sequential patching. So, there are total 2 attempts in batch patching and 1 in sequential patching.

src/core/src/core_logic/PatchInstaller.py
```diff
@@ -275,21 +275,50 @@ def install_updates(self, maintenance_window, package_manager, simulate=False):
         stopwatch_for_batch_install_process = Stopwatch(self.env_layer, self.telemetry_writer, self.composite_logger)
         stopwatch_for_batch_install_process.start()
 
-        # Check for packages available and install them in batches. Some packages may not be installed due to:
-        # (a) Not enough time remaining in maintenance window
-        # (b) Failure during package installation
-        installed_update_count, patch_installation_successful, maintenance_window_batch_cutoff_reached, packages, package_versions = self.install_packages_in_batches(
-            all_packages, all_package_versions, packages, package_versions, maintenance_window, package_manager)
+        total_packages_to_install_count = len(packages)
+        maintenance_window_batch_cutoff_reached = False
```
rane-rajasi, line 279: This should always be a check, where we validate time remaining, instead of assuming there will be time left since actual package installation hasn't started yet.

Reason for this is, patch operations are often not completed in a single occurence, due to restarts, machine reboots (by an external entity), etc. There could be an instance where maintenance window has exceeded at this point in code because it is a returning operation at this line.

rane-rajasi: nvm, thought this was the overall maintenance window for installation. Closing this comment

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -79,6 +79,8 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
 
         self.ubuntu_pro_client_all_updates_cached = []
         self.ubuntu_pro_client_all_updates_versions_cached = []
+        
+        self.package_install_expected_avg_time_in_minutes = 1.5 # As per telemetry data, the average time to install package is around 81 seconds for apt.
```
rane-rajasi, line 83: nit: using seconds would be better since the value is in decimal

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -762,3 +764,5 @@ def separate_out_esm_packages(self, packages, package_versions):
         self.composite_logger.log_debug("Filter esm packages : [TotalPackagesCount={0}][EsmPackagesCount={1}]".format(len(packages), len(ua_esm_required_packages)))
         return non_esm_packages, non_esm_package_versions, ua_esm_required_packages, ua_esm_required_package_versions, ua_esm_required_packages_found
 
+    def get_package_install_expected_avg_time_in_minutes(self):
+        return self.package_install_expected_avg_time_in_minutes
```
rane-rajasi, line 768: Leave 2 empty lines at the end of each file. See others for reference

src/core/src/package_managers/PackageManager.py
```diff
@@ -470,4 +470,8 @@ def set_security_esm_package_status(self, operation, packages):
     @abstractmethod
     def separate_out_esm_packages(self, packages, package_versions):
         pass
-
+    
+    @abstractmethod
+    def get_package_install_expected_avg_time_in_minutes(self):
+        """Retrieves average time to install package in minutes."""
+        pass
```
rane-rajasi, line 477: Same comment as the other one about empty lines at the end of each file

feng-j678 (changes requested):
please look into why the coverage is so low:
image

the cicd pipeline won't pass unless the code coverage is greater than 90%, please discuss this with Koshy or Rajasi if we can merge in the pr in this state, it will greatly reduce our coverage threshold

image

rane-rajasi (changes requested):
get_max_batch_size() looks good. 

An observation on the changes in install_updates() in PatchInstaller.py, the function is getting longer, please extract them out to another function in your revised code changes.

Also since get_max_batch_size() is an independent function, we can have UTs to validate it. For any future revisions, please try to add UTs to test out the smaller units. Overall UT to validate E2E scenario can be added in the iteration that has all code changes

GAURAVRAMRAKHYANI:
> get_max_batch_size() looks good.
> 
> An observation on the changes in install_updates() in PatchInstaller.py, the function is getting longer, please extract them out to another function in your revised code changes.
> 
> Also since get_max_batch_size() is an independent function, we can have UTs to validate it. For any future revisions, please try to add UTs to test out the smaller units. Overall UT to validate E2E scenario can be added in the iteration that has all code changes

(a) I have addressed the comment that as install_updates is getting longer, extracted some part of it to another function. Did some manual testings and no issues found. Continuing doing more testing around this change.

(b) I have not yet implemented UT for get_max_batch_size(), will implement it tomorrow IST. Meanwhile, can you review the product changes.

src/core/src/package_managers/PackageManager.py
GAURAVRAMRAKHYANI: Addressed it.

src/core/src/package_managers/AptitudePackageManager.py
GAURAVRAMRAKHYANI: Addressed it.

src/core/src/package_managers/AptitudePackageManager.py
GAURAVRAMRAKHYANI: I wanted to double check before making this change. 

As we have all the parameters in minutes in the functions get_max_batch_size() and is_package_install_time_available(), if we use seconds instead of minutes for package_install_expected_avg_time_in_minutes , then we will have to convert it to minutes in the above-mentioned methods. 

Should I still make it to seconds?

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: 2 empty lines

src/core/src/package_managers/PackageManager.py
rane-rajasi: 2 empty lines

rane-rajasi: comment applies everywhere

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: We generally use seconds throughout our code. My suggestion would be to change mins to secs in get_max_batch_size()

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: 2 empty lines

src/core/src/package_managers/PackageManager.py
rane-rajasi: 2 empty lines

rane-rajasi (changes requested):
Responded to asks on previous feedback. Waiting on UTs

feng-j678 (commented):
the cicd pipeline is failing, the code coverage does not meet the threshold, and there are failing tests

image
image

src/core/src/package_managers/AptitudePackageManager.py
GAURAVRAMRAKHYANI: Addressed it. Using seconds

src/core/src/package_managers/AptitudePackageManager.py
GAURAVRAMRAKHYANI: I have added 2 empty lines in all the files. I can see that in my local repo. I also tried to checkout the branch garamrak-multiPhase in another machine and can see the empty lines there also. Not sure why it is not visible in the PR.
Can you please verify that there are empty lines added after checkout to the garamrak-multiPhase branch

src/core/src/core_logic/PatchInstaller.py
```diff
@@ -736,3 +757,32 @@ def filter_out_excluded_updates(self, included_packages, included_package_versio
...
+        available_time_to_install_packages = available_time_to_install_packages - Constants.BUFFER_TIME_FOR_BATCH_PATCHING_START_IN_MINUTES
+
+        self.composite_logger.log_debug("Avaliable time in minutes to install packages (after removing buffer time): {0}".format(available_time_to_install_packages))
+
+        max_batch_size_for_packages = 0
+
+        # Taking assumption that 1 of the packages in the batch takes maximum expected time to install and remaining packages take average time to install.
+        if available_time_to_install_packages > Constants.PACKAGE_INSTALL_EXPECTED_MAX_TIME_IN_MINUTES:
+            available_time_to_install_packages = available_time_to_install_packages - Constants.PACKAGE_INSTALL_EXPECTED_MAX_TIME_IN_MINUTES
+            max_batch_size_for_packages = max_batch_size_for_packages + 1
```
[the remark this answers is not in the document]
GAURAVRAMRAKHYANI: Addressed it

src/core/src/core_logic/PatchInstaller.py
```diff
@@ -736,3 +757,32 @@ def filter_out_excluded_updates(self, included_packages, included_package_versio
...
+
+            # Remaining packages take average expected time to install.
+            package_install_expected_avg_time_in_minutes = package_manager.get_package_install_expected_avg_time_in_minutes()
+            max_batch_size_for_packages += int(math.floor(available_time_to_install_packages / package_install_expected_avg_time_in_minutes))
+
+        if max_batch_size_for_packages > Constants.MAX_BATCH_SIZE_FOR_PACKAGES:
+            max_batch_size_for_packages = Constants.MAX_BATCH_SIZE_FOR_PACKAGES
```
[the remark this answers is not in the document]
GAURAVRAMRAKHYANI: 300 is batch size.
The packages will be installed in batches of 300.
There is loop in install_packages_in_batches function which try to install packages in batches of batch size.
If there are 350 packages, and batch size is 300 then there are two batches, 1 with size 300 and the other with size 50.

In case after first batch of 300 packages, there is not enough time to install 50 packages in a batch then second phase of batch patching will be done in which batch size is reduced to 1/10th. So, the new batch size would be 30 and packages would be installed in batch of 30.

In case there is not enough time to install 30 packages in a batch then we fallback to sequential patching in which packages are installed sequentially one by one.

src/core/src/bootstrap/Constants.py
```diff
@@ -219,10 +219,21 @@ class StatusTruncationConfig(EnumBackport):
...
+    MAX_PHASES_FOR_BATCH_PATCHING = 2
+
+    # Batch size decay factor is factor to which batch size is decreased in batch patching to install remaining packages in case there
+    # are some package install failures with original batch size.
+    BATCH_SIZE_DECAY_FACTOR = 10
+
+    # We need to keep some buffer time between calculation of batch size and starting batch patching because after calculating the batch size,
+    # there would be little time taken before the batch patching is started. The function is_package_install_time_available is called before installing a batch.
+    # If we do not keep buffer then is_package_install_time_available would return false.
+    BUFFER_TIME_FOR_BATCH_PATCHING_START_IN_MINUTES = 5
```
kjohn-msft, line 235: Grouping config by proximity in the file is okay given that's the pattern in master. But everything in this file is being refactored to logically group related config, and you aligning with that model will avoid me having to refactor your code to merge.

See this Constants file from the last published iteration (of the long-delayed PR close) for concrete examples - https://github.com/Azure/LinuxPatchExtension/blob/173c52230f8c0f02f7fbf290006d4b67ff211928/src/core/src/bootstrap/Constants.py

So a class like PackageBatchConfig(EnumBackport) with variables shortened with the class context:
- PackageBatchConfig.MAX_BATCH_SIZE                    (for packages is implicit and understood)
- .DECAY_FACTOR
- .MAX_PHASES
- .BATCHING_START_BUFFER_IN_MINUTES                (name suggestion to shorten)

The comments are valuable and should be retained. They may be combined at the class level as a multi-line if it helps.

Also, the way to save time in doing this is to first define the class in Constants. Then do a string replace (not a symbol replace) on the original constant name with PackageBatchConfig.CONSTANT

kjohn-msft (approved):

kjohn-msft (changes requested):
Please have the automated checks pass (repeat). Code functionality is okay.

kjohn-msft:
Code cov is having a partial service outage. https://status.codecov.com/  
Retry after the status shows green. @feng-j678 for review on code state as now.

kjohn-msft (dismissed):

feng-j678 (approved):

feng-j678 (approved):

feng-j678 (approved):

kjohn-msft (approved):

PR #251, opened 2024-03-19
Satim/version

SathishMSFT:
Distutils.version is unsupported from python 3.11+. There is not direct version compare for the same. As per suggestion in this thread (https://peps.python.org/pep-0632/#migration-advice), we have taken the version.py (and its dependencies) from packaging module.

Verified this to be working in ubuntu 18.04 where it will be used.


kjohn-msft:
Ok to abandon and clean up this now?

PR #252, opened 2024-03-27
 Bugfix for out-of-date package lists causing install failures

kjohn-msft:
There is a non-obvious bug caused by performance-enhancing code that was added 6-7 years ago. 

Local package lists are left in a state inconsistent against subsequent installation intent because of the sequence of execution of code. This results in us invoking installation actions based on what we have in memory from a prior call while the system is not in the same state.

This never affected operations installing security updates only. It sometimes affected operations installing all updates based on the relative state of repositories (security vs. all others). It most recently reproduced consistently on Ubuntu 22.04.


feng-j678 (approved):

kjohn-msft:
Merging as an exception due to hotfix.

PR #260, opened 2024-06-18
Refreshing cache with machine source prior to ending execution

rane-rajasi:
Refreshes cache with machine source before ending execution, since we have set private sources list in our code in places. Reverting to machine default to ensure default machine experience for customers

### DIFF
--- a/src/core/src/CoreMain.py
+++ b/src/core/src/CoreMain.py
@@ -29,6 +29,7 @@ class CoreMain(object):
         stdout_file_mirror = bootstrapper.stdout_file_mirror
         telemetry_writer = bootstrapper.telemetry_writer
         lifecycle_manager = status_handler = execution_config = None
+        package_manager = None
 
         # Init operation statuses
         patch_operation_requested = Constants.UNKNOWN
@@ -101,7 +102,6 @@ class CoreMain(object):
                     patch_installer.mark_installation_completed()
                     overall_patch_installation_operation_successful = True
                 self.update_patch_substatus_if_pending(patch_operation_requested, overall_patch_installation_operation_successful, patch_assessment_successful, configure_patching_successful, status_handler, composite_logger)
-                package_manager.refresh_repo()
         except Exception as error:
             # Privileged operation handling for non-production use
             if Constants.EnvLayer.PRIVILEGED_OP_MARKER in repr(error):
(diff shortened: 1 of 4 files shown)


src/core/src/CoreMain.py
```diff
@@ -101,7 +101,7 @@ def __init__(self, argv):
                     patch_installer.mark_installation_completed()
                     overall_patch_installation_operation_successful = True
                 self.update_patch_substatus_if_pending(patch_operation_requested, overall_patch_installation_operation_successful, patch_assessment_successful, configure_patching_successful, status_handler, composite_logger)
-
+                package_manager.refresh_repo()
```
kjohn-msft, line 104: This will not run in a legitimate exception throw from code higher up.

We may need to extend package manager to have refresh_repo_safely() which is just a solid try catch in the PackageManager class over refresh_repo()   -- concrete method, not abstract.

And then refresh_repo_safely can be called at the top of the finally block (requires package_manager to be init to None outside the coremain try catch, and then null check on it in the finally block.

kjohn-msft (changes requested):
Comment left inline

src/core/src/CoreMain.py
rane-rajasi: Addressed

src/core/src/package_managers/PackageManager.py
```diff
@@ -61,6 +61,13 @@ def refresh_repo(self):
         """Resynchronize the package index files from their sources."""
         pass
 
+    def refresh_repo_safely(self):
+        """Resynchronize the package index files from machine sources."""
+        try:
+            self.refresh_repo()
+        except Exception as error:
+            self.composite_logger.log_error("Error in refreshing cache from machine sources. [Error={0}]".format(repr(error)))
```
kjohn-msft, line 69: log_debug is fine - this error is not customer facing and doesn't need to be reported to them. What we are doing here is best effort.

kjohn-msft (changes requested):
Minor, inline

src/core/src/package_managers/PackageManager.py
rane-rajasi: Agree with the log type. What is your view on reporting this error in status blob? I'm still debating over. It gets logged within ZypperPkgMgr currently. Not the other PkgMgrs

kjohn-msft: That's ok, this is just a quiet in & out operation.. 

The zypper flow is extremely unlikely to ever trigger something new here that hasn't been triggered before and been acted on. 

But - unrelated to this PR, 'Setting force_reboot flag to True' in Zypper needs to be a log_debug too. It's not a customer-facing message. Please make the change here too.

rane-rajasi: Addressed

kjohn-msft (approved):

feng-j678 (approved):

PR #264, opened 2024-09-09
Fix broken 3-byte-unicode-related unit test in Python 2.x

feng-j678:
[x] address the failing test case (test_write_event_msg_size_limit_char_more_than_1_bytes) in py27 that it can't handle (€) 3 bytes ascii char: Non-ASCII (€) character in file line, but no encoding declared;
[x] modify the test using unicode char ("a\u20acbc" ) as input instead of € 
[x] modify the last assertion checking the message length is less than and equal to the accept limit instead of mathematically calculating the expected message vs the actual message content. in py27, it is recognizing the above unicode as 9 bytes, but py3 recognize it as 4 bytes, therefore the calculation in the last assertion never passed.
[x] remove € in __set_up_patches_func() Test_StatusHandlerTruncation.py
[x] the reason for that assertion change was the unicode as 9 bytes in py2, but py3 recognize it as 4 bytes, so the calculation in the that assertion will never passed, it's very difficult to get the exact mathematical result for both. so we went with
using the upper and lower bound limit in encoding to assert the truncation is applied and it's within a reasonable byte range

### DIFF
--- a/src/core/tests/Test_TelemetryWriter.py
+++ b/src/core/tests/Test_TelemetryWriter.py
@@ -17,7 +17,6 @@ import datetime
 import json
 import os
 import re
-import sys
 import time
 import unittest
 from core.src.bootstrap.Constants import Constants
--- a/src/core/tests/Test_TelemetryWriter.py
+++ b/src/core/tests/Test_TelemetryWriter.py
@@ -17,6 +17,7 @@ import datetime
 import json
 import os
 import re
+import sys
 import time
 import unittest
 from core.src.bootstrap.Constants import Constants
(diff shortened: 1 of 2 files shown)


src/core/tests/Test_TelemetryWriter.py
```diff
@@ -102,19 +102,19 @@ def test_write_event_msg_size_limit(self):
...
         self.runtime.telemetry_writer.write_event(message, Constants.TelemetryEventLevel.Error, "Test Task")
         latest_event_file = [pos_json for pos_json in os.listdir(self.runtime.telemetry_writer.events_folder_path) if re.search('^[0-9]+.json$', pos_json)][-1]
         with open(os.path.join(self.runtime.telemetry_writer.events_folder_path, latest_event_file), 'r+') as f:
             events = json.load(f)
             self.assertTrue(events is not None)
             self.assertEqual(events[-1]["TaskName"], "Test Task")
             self.assertTrue(len(events[-1]["Message"]) < len(message.encode('utf-8')))
-            chars_dropped = len(message.encode('utf-8')) - Constants.TELEMETRY_MSG_SIZE_LIMIT_IN_CHARS + Constants.TELEMETRY_BUFFER_FOR_DROPPED_COUNT_MSG_IN_CHARS + Constants.TELEMETRY_EVENT_COUNTER_MSG_SIZE_LIMIT_IN_CHARS
-            self.assertTrue("a€bc" in events[-1]["Message"])
-            self.assertTrue("a€bc" * (len(message) + 1 - chars_dropped) + ". [{0} chars dropped]".format(chars_dropped) in events[-1]["Message"])  # len(message) + 1 due to bad unicode will be replaced by �
```
kjohn-msft, line 117: Why were the additional assertions removed?

kjohn-msft (commented):
Question inline

src/core/tests/Test_TelemetryWriter.py
rane-rajasi: @feng-j678 this is still open

rane-rajasi (changes requested):
Feedback inline

src/core/tests/Test_TelemetryWriter.py
rane-rajasi: @feng-j678 please update what was discussed between you and Koshy about this feedback here. This gives me clarity and will also be useful in future

rane-rajasi: Copying reasoning from PR description

image

rane-rajasi (approved):

kjohn-msft (approved):

PR #266, opened 2024-09-23
Bugfix/disable rhel8+ security plugin

feng-j678:
[x] per BUG item 28214906 - disable yum install security plugin on rhel8+ images
test result:
rhel810 images
image
prior code change assessment call:
image
post code change assessment call;
image
post code change installation call:
image
image

### DIFF
--- a/src/core/src/package_managers/YumPackageManager.py
+++ b/src/core/src/package_managers/YumPackageManager.py
@@ -142,11 +142,11 @@ class YumPackageManager(PackageManager):
 
     def get_security_updates(self):
         """Get missing security updates"""
-        if self.__is_image_rhel8_higher_skip_security_plugin():
-            return [], []
-
         self.composite_logger.log("\nDiscovering 'security' packages...")
-        self.install_yum_security_prerequisite()
+
+        if not self.__is_image_rhel8_higher():
+            self.install_yum_security_prerequisite()
+
         out = self.invoke_package_manager(self.yum_check_security)
         security_packages, security_package_versions = self.extract_packages_and_versions(out)
 
@@ -179,10 +179,10 @@ class YumPackageManager(PackageManager):
         self.composite_logger.log("Discovered " + str(len(other_packages)) + " 'other' package entries.")
         return other_packages, other_package_versions
 
-    def __is_image_rhel8_higher_skip_security_plugin(self):
-        """Check image RHEL8+ to disable install yum-plugin-security """
+    def __is_image_rhel8_higher(self):
+        """ Check if image is RHEL8+ set true else false """
         if self.env_layer.platform.linux_distribution() is not None:
-            os_offer, os_version, os_code = self.env_layer.platform.linux_distribution();
+            os_offer, os_version, os_code = self.env_layer.platform.linux_distribution()
 
             if "Red Hat Enterprise Linux" in os_offer and int(os_version.split('.')[0]) >= 8:
                 self.composite_logger.log_debug("Disable RHEL8+ install yum-plugin-security: Os Version " + str(os_version))
--- a/src/core/src/package_managers/YumPackageManager.py
+++ b/src/core/src/package_managers/YumPackageManager.py
@@ -142,7 +142,7 @@ class YumPackageManager(PackageManager):
 
     def get_security_updates(self):
         """Get missing security updates"""
-        if self.__is_image_rhel8_higher_for_security_plugin():
+        if self.__is_image_rhel8_higher_skip_security_plugin():
             return [], []
 
         self.composite_logger.log("\nDiscovering 'security' packages...")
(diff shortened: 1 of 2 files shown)


src/core/src/package_managers/YumPackageManager.py
```diff
@@ -142,6 +142,9 @@ def get_all_updates(self, cached=False):
 
     def get_security_updates(self):
         """Get missing security updates"""
+        if self.__is_image_rhel8_higher_skip_security_plugin():
+            return [], []
+
         self.composite_logger.log("\nDiscovering 'security' packages...")
         self.install_yum_security_prerequisite()
```
kjohn-msft, line 149: The purpose of the fix is to not run Line 149 (and only Line 149) " self.install_yum_security_prerequisite() " if the RHEL version is 8+.

What the code here does is it skips everything in this function which is incorrect.

So. ---
if not rhel8+:
    self.install...

continue after that regardless

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -176,6 +179,17 @@ def get_other_updates(self):
...
+    def __is_image_rhel8_higher_skip_security_plugin(self):
+        """Check image RHEL8+ to disable install yum-plugin-security """
+        if self.env_layer.platform.linux_distribution() is not None:
+            os_offer, os_version, os_code = self.env_layer.platform.linux_distribution();
+
+            if "Red Hat Enterprise Linux" in os_offer and int(os_version.split('.')[0]) >= 8:
+                self.composite_logger.log_debug("Disable RHEL8+ install yum-plugin-security: Os Version " + str(os_version))
+                return True
+
+        return False
```
kjohn-msft, line 191: Single responsibility - the _only_ responsibility of the code here is to know if it's RHEL 8 or greater.
The name of the function should reflect that. The actions inside the function should reflect that.
There should be no reference to what the calling code wants or will do with the return.

kjohn-msft (changes requested):
Comments inline.

src/core/src/package_managers/YumPackageManager.py
feng-j678: logic is apply on self.install_yum_security_prerequisite

src/core/src/package_managers/YumPackageManager.py
feng-j678: updated

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -143,7 +143,10 @@ def get_all_updates(self, cached=False):
     def get_security_updates(self):
         """Get missing security updates"""
         self.composite_logger.log("\nDiscovering 'security' packages...")
-        self.install_yum_security_prerequisite()
+
+        if not self.__is_image_rhel8_higher():
```
rane-rajasi, line 147: nit: __is_image_rhel8_or_higher

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -176,6 +179,17 @@ def get_other_updates(self):
         self.composite_logger.log("Discovered " + str(len(other_packages)) + " 'other' package entries.")
         return other_packages, other_package_versions
 
+    def __is_image_rhel8_higher(self):
+        """ Check if image is RHEL8+ set true else false """
```
rane-rajasi, line 183: nit: """ If image is RHEL8+ return true else false """

src/core/tests/Test_YumPackageManager.py
```diff
@@ -619,5 +626,55 @@ def test_obsolete_packages_should_not_considered_in_available_updates(self):
...
+    def test_rhel8_image_higher_no_security_plugin(self):
+        """Unit test for yum package manager rhel images >= 8 and Classification = Security"""
+        # mock linux_distribution
+        backup_envlayer_platform_linux_distribution = LegacyEnvLayerExtensions.LegacyPlatform.linux_distribution
+        LegacyEnvLayerExtensions.LegacyPlatform.linux_distribution = self.mock_linux8_distribution_to_return_redhat
+
+        self.__assert_test_rhel8_image()
+
+        # restore linux_distribution
+        LegacyEnvLayerExtensions.LegacyPlatform.linux_distribution = backup_envlayer_platform_linux_distribution
```
rane-rajasi, line 649: Add UTs with rhel7 image scenario as well

rane-rajasi (changes requested):
Comments inline

src/core/tests/Test_YumPackageManager.py
feng-j678: it was added, now i've rename the ut;

image

src/core/src/package_managers/YumPackageManager.py
feng-j678: done

kjohn-msft (approved):

PR #267, opened 2024-09-24
Pr1/increase_RebootManager_and_PatchAssessor_code coverage

feng-j678:
[x] add ut to RebootManager.py
[x] add ut test_raise_add_error_to_status in PatchAssessor.py
image

### DIFF
--- a/src/core/tests/Test_ConfigurePatchingProcessor.py
+++ b/src/core/tests/Test_ConfigurePatchingProcessor.py
@@ -18,7 +18,11 @@ import os
 import re
 import unittest
 import sys
-from io import StringIO
+# Conditional import for StringIO
+try:
+    from StringIO import StringIO  # Python 2
+except ImportError:
+    from io import StringIO  # Python 3
 
 from core.src.CoreMain import CoreMain
 from core.src.bootstrap.Constants import Constants
@@ -317,6 +321,7 @@ class TestConfigurePatchingProcessor(unittest.TestCase):
     def test_configure_patching_raise_exception_auto_os_patch_state(self):
         # arrange capture std IO
         captured_output = StringIO()
+        original_stdout = sys.stdout
         sys.stdout = captured_output
 
         argument_composer = ArgumentComposer()
@@ -334,7 +339,7 @@ class TestConfigurePatchingProcessor(unittest.TestCase):
         runtime.configure_patching_processor.start_configure_patching()
 
         # restore sdt.out ouptput
-        sys.stdout = sys.__stdout__
+        sys.stdout = original_stdout
 
         # assert
         output = captured_output.getvalue()
--- a/src/core/tests/Test_ConfigurePatchingProcessor.py
+++ b/src/core/tests/Test_ConfigurePatchingProcessor.py
@@ -18,11 +18,7 @@ import os
 import re
 import unittest
 import sys
-# Conditional import for StringIO
-try:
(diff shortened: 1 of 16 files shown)


src/core/tests/Test_ConfigurePatchingProcessor.py
```diff
@@ -317,6 +321,7 @@ def test_configure_patching_with_patch_mode_and_assessment_mode_by_platform(self
     def test_configure_patching_raise_exception_auto_os_patch_state(self):
         # arrange capture std IO
         captured_output = StringIO()
+        original_stdout = sys.stdout
```
rane-rajasi, line 324: Checking...
Is this sys.stdout or sys.__stdout__? Since that was removed from line 342 below

src/core/tests/Test_PatchAssessor.py
```diff
@@ -181,9 +180,18 @@ def test_raise_if_min_python_version_not_met(self):
             self.runtime.patch_assessor.start_assessment()
         self.assertEqual(str(context.exception), Constants.PYTHON_NOT_COMPATIBLE_ERROR_MSG.format(sys.version_info))
 
+    def test_raise_add_error_to_status(self):
```
rane-rajasi, line 183: This should be in Test_StatusHandler.py

feng-j678: I don't see why this needs to be moved to statushandler, statushandler capture the output of operation, this test is asserting the exception at time operation starts and the function patch_assessor.start_assessment is a public method in patchassessor, so I believe it's better to have it corresponding to class file

src/core/tests/Test_ConfigurePatchingProcessor.py
feng-j678: this is sys.stdout can be redirect and store terminal output

src/core/tests/Test_PatchAssessor.py
rane-rajasi: In that case, change the test function name to reflect the scenario. For eg: test_patch assessment_throws_exception()
test_raise_add_error_to_status() indicates you are testing **add_error_to_status** which is a functionality of StatusHandler

src/core/tests/Test_PatchAssessor.py
```diff
@@ -181,9 +180,18 @@ def test_raise_if_min_python_version_not_met(self):
...
+        with self.assertRaises(Exception) as context:
+            self.runtime.patch_assessor.start_assessment()
+            
+        self.assertIn(Constants.ERROR_ADDED_TO_STATUS, repr(context.exception))
+        self.assertEqual(context.exception.args[1], "[{0}]".format(Constants.ERROR_ADDED_TO_STATUS))
+
     def raise_ex(self):
         raise Exception()
-
+
```
rane-rajasi, line 194: nit: remove tab spacing

rane-rajasi (changes requested):
Comments inline

src/core/tests/Test_PatchAssessor.py
feng-j678: done

src/core/tests/Test_PatchAssessor.py
feng-j678: done

kjohn-msft (approved):

rane-rajasi (approved):

PR #278, opened 2024-11-10
Add retry to check_sudo_status and unit test

feng-j678:
[x] add retry logic into check_sudo_status that check sudo for 30 mins at interval of 5min (300sec)
[x] add MAX_CHECK_SUDO_RETRY_COUNT, MAX_CHECK_SUDO_INTERVAL_IN_SEC in constant
[x] add add unit test in test_coremain file to test all possible check sudo cases

image
image
image
image


src/core/tests/library/RuntimeCompositor.py
```diff
@@ -41,9 +41,11 @@
 
 
 class RuntimeCompositor(object):
```
kjohn-msft, line 43: Why is the runtime compositor needed to unit test Bootstrapper? Two reasons not to do this:
- Runtime compositor is meant for tests requiring most of the classes to be initialized. Bootstrapper does not need that. I'd just have a new class Test_Bootstrapper and have the sudo tests in there with minimal initialization of bootstrapper.
- The way this is done is it unnecessarily overloads RuntimeCompositor for all tests while only catering to the needs of a single class (and not creating anything truly shared for all the tests).

src/core/src/bootstrap/Bootstrapper.py
```diff
@@ -147,31 +148,46 @@ def basic_environment_health_check(self):
...
+                    if raise_if_not_sudo:
+                        raise
+                    return False
+
+                self.composite_logger.log_error("Retrying sudo status check after a delay of [ElapsedTimeInSeconds={0}][Attempts={1}]".format(Constants.MAX_CHECK_SUDO_INTERVAL_IN_SEC, str(attempts)))
+                time.sleep(Constants.MAX_CHECK_SUDO_INTERVAL_IN_SEC)
+
+    def run_command_output(self, command, no_output=False, chk_err=True):
+        """ Runs a command and returns the output """
+        return self.env_layer.run_command_output(command, no_output, chk_err)
```
kjohn-msft, line 192: Why is this here? For unit testability? With the suggestion not to use runtime compositor, it's possible to mock the env_layer entry itself right?

src/core/src/bootstrap/Bootstrapper.py
```diff
@@ -147,31 +148,46 @@ def basic_environment_health_check(self):
...
+                    self.composite_logger.log_debug("Sudo Check Failed [Attempts={0}][MaxAttempts={1}]".format(str(attempts), Constants.MAX_CHECK_SUDO_RETRY_COUNT))
+                    raise Exception("Unexpected sudo check result. Output: " + " ".join(output.split("\n")))
+            except Exception as exception:
+                self.composite_logger.log_error("Sudo status check failed. Please ensure the computer is configured correctly for sudo invocation. " +
+                                                "Exception details: " + str(exception))
+                if attempts == Constants.MAX_CHECK_SUDO_RETRY_COUNT:
+                    self.composite_logger.log_debug("Sudo Check Failed, reached [MaxAttempts={0}]".format(str(attempts)))
+                    if raise_if_not_sudo:
+                        raise
+                    return False
```
kjohn-msft, line 185: check_sudo_status_with_retry -->
can invoke --  check_sudo_status (unchanged code)

Much easier to read code and also improves unit-testability. Pretty sure this was discussed last week.

kjohn-msft (changes requested):
Comments inline

src/core/src/bootstrap/Bootstrapper.py
```diff
@@ -147,31 +148,46 @@ def basic_environment_health_check(self):
...
+                # output should look like either this (bad):
+                #   [sudo] password for username:
+                #   False
+                # or this (good):
+                #   uid=0(root) gid=0(root) groups=0(root)
+                #   True
+
+                output_lines = output.splitlines()
+                if len(output_lines) < 2:
+                    self.composite_logger.log_debug("Sudo Check Failed [Attempts={0}][MaxAttempts={1}]".format(str(attempts), Constants.MAX_CHECK_SUDO_RETRY_COUNT))
```
rane-rajasi, line 164: This log should give out the exception message as well

src/core/src/bootstrap/Bootstrapper.py
```diff
@@ -147,31 +148,46 @@ def basic_environment_health_check(self):
...
+                    raise Exception("Unexpected sudo check result. Output: " + " ".join(output.split("\n")))
+
+                if output_lines[1] == "True":
+                    self.composite_logger.log_debug("Sudo Check Successfully [Attempts={0}][MaxAttempts={1}]".format(str(attempts), Constants.MAX_CHECK_SUDO_RETRY_COUNT))
+                    return True
+                elif output_lines[1] == "False":
+                    if raise_if_not_sudo:
+                        self.composite_logger.log_debug("Sudo Check Failed [Attempts={0}][MaxAttempts={1}]".format(str(attempts), Constants.MAX_CHECK_SUDO_RETRY_COUNT))
```
rane-rajasi, line 172: add exception message to the log as well

rane-rajasi (changes requested):
Comments inline

src/core/src/bootstrap/Bootstrapper.py
feng-j678: can't add exception to the log here because we're not catching exception, we are raising one

src/core/src/bootstrap/Bootstrapper.py
feng-j678: can't add exception to the log here because we're not catching exception, we are raising one

src/core/src/bootstrap/Bootstrapper.py
feng-j678: move the env_layer mocks to a new Test_Bootstrapper.py

src/core/tests/library/RuntimeCompositor.py
feng-j678: not exactly the ask here, the runtime compositor initialize all necessary objects, this is a set up class so it's bootstrapper. all other test files are using compositor for initiation and for unit testings

src/core/src/bootstrap/Bootstrapper.py
```diff
@@ -142,36 +143,46 @@ def basic_environment_health_check(self):
...
+        if output_lines[1] == "True":
+            return True
+        elif output_lines[1] == "False":
             if raise_if_not_sudo:
-                raise
-
+                raise Exception("Unable to invoke sudo successfully. Output: " + " ".join(output.split("\n")))
+            return False
+        else:
+            raise Exception("Unexpected sudo check result. Output: " + " ".join(output.split("\n")))
```
kjohn-msft, line 188: I'm not seeing a good reason to change any of this - the highlighted note:

image

src/core/src/bootstrap/Bootstrapper.py
```diff
@@ -142,36 +143,46 @@ def basic_environment_health_check(self):
         self.composite_logger.log("Process id: " + str(os.getpid()))
 
         # Ensure sudo works in the environment
-        sudo_check_result = self.check_sudo_status()
+        sudo_check_result = self.check_sudo_status_retry()
         self.composite_logger.log_debug("Sudo status check: " + str(sudo_check_result) + "\n")
 
+    def check_sudo_status_retry(self, raise_if_not_sudo=True):
```
kjohn-msft, line 149: See name of the function in the previous comment:
image

kjohn-msft (changes requested):
Please review original comments again

src/core/src/bootstrap/Bootstrapper.py
```diff
@@ -142,36 +143,46 @@ def basic_environment_health_check(self):
...
+            try:
+                self.composite_logger.log_debug("Performing sudo status check... This should complete within 10 seconds. [Attempts={0}][MaxAttempts={1}]".format(str(attempts), Constants.MAX_CHECK_SUDO_RETRY_COUNT))
+                if self.check_sudo_status(raise_if_not_sudo=raise_if_not_sudo):
+                    self.composite_logger.log_debug("Sudo Check Successfully [Attempts={0}][MaxAttempts={1}]".format(str(attempts), Constants.MAX_CHECK_SUDO_RETRY_COUNT))
+                    return True
+            except Exception as exception:
+                self.composite_logger.log_error("Sudo status check failed. Please ensure the computer is configured correctly for sudo invocation. " + "Exception details: " + str(exception))
+
+                if attempts == Constants.MAX_CHECK_SUDO_RETRY_COUNT:
+                    self.composite_logger.log_debug("Sudo Check Failed, reached [MaxAttempts={0}]".format(str(attempts)))
```
kjohn-msft, line 161: error

src/core/src/bootstrap/Bootstrapper.py
```diff
@@ -142,36 +143,46 @@ def basic_environment_health_check(self):
...
+    def check_sudo_status_retry(self, raise_if_not_sudo=True):
+        """ retry to invoke sudo check """
+        for attempts in range(1, Constants.MAX_CHECK_SUDO_RETRY_COUNT + 1):
+            try:
+                self.composite_logger.log_debug("Performing sudo status check... This should complete within 10 seconds. [Attempts={0}][MaxAttempts={1}]".format(str(attempts), Constants.MAX_CHECK_SUDO_RETRY_COUNT))
+                if self.check_sudo_status(raise_if_not_sudo=raise_if_not_sudo):
+                    self.composite_logger.log_debug("Sudo Check Successfully [Attempts={0}][MaxAttempts={1}]".format(str(attempts), Constants.MAX_CHECK_SUDO_RETRY_COUNT))
+                    return True
+            except Exception as exception:
+                self.composite_logger.log_error("Sudo status check failed. Please ensure the computer is configured correctly for sudo invocation. " + "Exception details: " + str(exception))
```
kjohn-msft, line 158: remove because inner function has this

src/core/src/bootstrap/Bootstrapper.py
```diff
@@ -142,36 +143,46 @@ def basic_environment_health_check(self):
...
         # Ensure sudo works in the environment
-        sudo_check_result = self.check_sudo_status()
+        sudo_check_result = self.check_sudo_status_retry()
         self.composite_logger.log_debug("Sudo status check: " + str(sudo_check_result) + "\n")
 
+    def check_sudo_status_retry(self, raise_if_not_sudo=True):
+        """ retry to invoke sudo check """
+        for attempts in range(1, Constants.MAX_CHECK_SUDO_RETRY_COUNT + 1):
+            try:
+                self.composite_logger.log_debug("Performing sudo status check... This should complete within 10 seconds. [Attempts={0}][MaxAttempts={1}]".format(str(attempts), Constants.MAX_CHECK_SUDO_RETRY_COUNT))
```
kjohn-msft, line 153: remove this as inner function has this

src/core/src/bootstrap/Bootstrapper.py
```diff
@@ -142,36 +143,46 @@ def basic_environment_health_check(self):
...
+                if self.check_sudo_status(raise_if_not_sudo=raise_if_not_sudo):
+                    self.composite_logger.log_debug("Sudo Check Successfully [Attempts={0}][MaxAttempts={1}]".format(str(attempts), Constants.MAX_CHECK_SUDO_RETRY_COUNT))
+                    return True
```
kjohn-msft, line 156: Ignore - only do the function call inside the try catch

src/core/src/bootstrap/Bootstrapper.py
```diff
@@ -142,36 +143,46 @@ def basic_environment_health_check(self):
...
+            except Exception as exception:
+                self.composite_logger.log_error("Sudo status check failed. Please ensure the computer is configured correctly for sudo invocation. " + "Exception details: " + str(exception))
+
+                if attempts == Constants.MAX_CHECK_SUDO_RETRY_COUNT:
```
kjohn-msft, line 160: >= 0  -- safety check for infinite loop

src/core/src/bootstrap/Bootstrapper.py
feng-j678: we need a successful log and a return true here since retry invoke check_sudo_status or else we'll get null exception on this code: sudo_check_result = self.check_sudo_status_with_retry()

src/core/src/bootstrap/Bootstrapper.py
feng-j678: revert check_sudo_status to original state, but i think we should move this self.composite_logger.log("Performing sudo status check... This should complete within 10 seconds.") to retry logic

src/core/src/bootstrap/Bootstrapper.py
```diff
@@ -142,36 +143,46 @@ def basic_environment_health_check(self):
...
+                    self.composite_logger.log_debug("Sudo Check Failed, reached [MaxAttempts={0}]".format(str(attempts)))
+                    if raise_if_not_sudo:
+                        raise
+
+                self.composite_logger.log_error("Retrying sudo status check after a delay of [ElapsedTimeInSeconds={0}][Attempts={1}]".format(Constants.MAX_CHECK_SUDO_INTERVAL_IN_SEC, str(attempts)))
+                time.sleep(Constants.MAX_CHECK_SUDO_INTERVAL_IN_SEC)
```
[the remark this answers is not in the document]
feng-j678: revert check_sudo_status to original state, but i think we should move this self.composite_logger.log("Performing sudo status check... This should complete within 10 seconds.") to retry logic

src/core/src/bootstrap/Bootstrapper.py
feng-j678: done

src/core/src/bootstrap/Bootstrapper.py
feng-j678: this log is captures in check_sudo_status()

src/core/src/bootstrap/Bootstrapper.py
feng-j678: keeping this log because it tracks the # of attempts are we retry

src/core/src/bootstrap/Bootstrapper.py
feng-j678: done

src/core/src/bootstrap/Bootstrapper.py
```diff
@@ -174,4 +192,3 @@ def check_sudo_status(self, raise_if_not_sudo=True):
                                             "Exception details: " + str(exception))
             if raise_if_not_sudo:
                 raise
-
```
kjohn-msft, line 177: missing line

kjohn-msft: All files end with extra blank line at the end

src/core/src/bootstrap/Bootstrapper.py
```diff
@@ -142,9 +143,26 @@ def basic_environment_health_check(self):
...
+        """ retry to invoke sudo check """
+        for attempts in range(1, Constants.MAX_CHECK_SUDO_RETRY_COUNT + 1):
+            try:
+                self.composite_logger.log_debug("Attempt sudo status check [Attempts={0}][MaxAttempts={1}]".format(str(attempts), Constants.MAX_CHECK_SUDO_RETRY_COUNT))
+                if self.check_sudo_status(raise_if_not_sudo=raise_if_not_sudo):
+                    self.composite_logger.log_debug("Sudo Check Successfully [Attempts={0}][MaxAttempts={1}]".format(str(attempts), Constants.MAX_CHECK_SUDO_RETRY_COUNT))
+                    return True
+            except Exception as exception:
+                if attempts >= 0 and attempts >= Constants.MAX_CHECK_SUDO_RETRY_COUNT:
+                    self.composite_logger.log_error("Sudo Check Failed, reached [MaxAttempts={0}][Exception={1}]".format(str(attempts), str(exception)))
```
kjohn-msft, line 159: self.composite_logger.log_error("Customer environment error (sudo failure). [Exception={1}][RetryCount={0}]".format(str(exception), str(attempts)))

src/core/src/bootstrap/Bootstrapper.py
kjohn-msft: You don't need a log.

You just need:
return self.check_sudo_status(raise_if_not_sudo)

src/core/src/bootstrap/Bootstrapper.py
```diff
@@ -142,9 +143,26 @@ def basic_environment_health_check(self):
...
         # Ensure sudo works in the environment
-        sudo_check_result = self.check_sudo_status()
+        sudo_check_result = self.check_sudo_status_with_retry()
         self.composite_logger.log_debug("Sudo status check: " + str(sudo_check_result) + "\n")
 
+    def check_sudo_status_with_retry(self, raise_if_not_sudo=True):
+        """ retry to invoke sudo check """
+        for attempts in range(1, Constants.MAX_CHECK_SUDO_RETRY_COUNT + 1):
+            try:
+                self.composite_logger.log_debug("Attempt sudo status check [Attempts={0}][MaxAttempts={1}]".format(str(attempts), Constants.MAX_CHECK_SUDO_RETRY_COUNT))
```
kjohn-msft: This log line is not required - inner function reports already

kjohn-msft: To be clear: "Ignore - only do the function call inside the try catch"
=>
try:
    return self.check_sudo_status_with_retry()
except Exception as exception:
   <blah>

kjohn-msft (approved):

kjohn-msft (approved):

src/core/src/bootstrap/Bootstrapper.py
```diff
@@ -170,7 +190,7 @@ def check_sudo_status(self, raise_if_not_sudo=True):
             else:
                 raise Exception("Unexpected sudo check result. Output: " + " ".join(output.split("\n")))
         except Exception as exception:
-            self.composite_logger.log_error("Sudo status check failed. Please ensure the computer is configured correctly for sudo invocation. " +
+            self.composite_logger.log_debug("Sudo status check failed. Please ensure the computer is configured correctly for sudo invocation. " +
                                             "Exception details: " + str(exception))
             if raise_if_not_sudo:
                 raise
```
rane-rajasi, line 196: If an exception is raised, this function will return **None** if raise_if_not_sudo=False, since we aren't explicitly returning a bool. and on line 155 (if sudo_status and attempts > 1:) we are checking for a bool. 

Depending on what seems best for all callers of this functions, either make sure it returns False here (i.e. in exception path when raise_if_not_sudo=False) or modify line 155 to accomodate sudo_status having None value as well

rane-rajasi (changes requested):
Left one comment

src/core/src/bootstrap/Bootstrapper.py
feng-j678: yes, you're right. I have added retry logic to handle when sudo status is either none or false

src/core/src/bootstrap/Bootstrapper.py
```diff
@@ -142,10 +143,40 @@ def basic_environment_health_check(self):
...
+                        self.composite_logger.log_debug("Retrying sudo status check after a delay of [ElapsedTimeInSeconds={0}][RetryCount={1}]".format(Constants.MAX_CHECK_SUDO_INTERVAL_IN_SEC, str(attempts)))
+                        time.sleep(Constants.MAX_CHECK_SUDO_INTERVAL_IN_SEC)
+                        continue
+
+                    elif attempts >= Constants.MAX_CHECK_SUDO_RETRY_COUNT:
+                        raise
+
+            except Exception as exception:
+                if attempts >= Constants.MAX_CHECK_SUDO_RETRY_COUNT:
+                    self.composite_logger.log_error("Customer environment error (sudo failure). [Exception={0}][MaxRetryCount={1}]".format(str(exception), str(attempts)))
```
kjohn-msft, line 171: MaxRetryCount was RetryCount in the last iteration I saw. RetryCount is accurate. MaxRetryCount is not (it would imply a constant maximum).

src/core/src/bootstrap/Bootstrapper.py
```diff
@@ -142,10 +143,40 @@ def basic_environment_health_check(self):
...
+                    if raise_if_not_sudo:
+                        raise
+                    return False
```
kjohn-msft, line 174: Leave the return 'None' (similar to the non-retry function) because you don't know if it's false. All you know is that it's not true. At a higher level, the contract should remain the same - it's just doing retries transparently.

rane-rajasi (approved):

kjohn-msft (approved):

rane-rajasi (approved):

PR #287, opened 2025-01-22
Fix excessive disk logging

feng-j678:
[x] pr to address bug item 29588036: Prevent excessive logging to disk in runway OS-error conditions
[x] add truncation logic to handle runway logging to disk on machines when there's error, especially network issue. the truncation will cut the message at a safe point at newline and max log content size in byte is set at 32MB to prevent disk space exhaustion on machine
[x] add unit tests

### DIFF
--- a/src/core/src/local_loggers/FileLogger.py
+++ b/src/core/src/local_loggers/FileLogger.py
@@ -38,8 +38,9 @@ class FileLogger(object):
 
     def write(self, message, fail_silently=True):
         try:
+            truncated_message = self.truncate_message(message)
             if self.log_file_handle is not None:
-                self.log_file_handle.write(message)
+                self.log_file_handle.write(truncated_message)
         except Exception as error:
             # DO NOT write any errors here to stdout
             failure_message = "Fatal exception trying to write to log file: " + repr(error) + ". Attempted message: " + str(message)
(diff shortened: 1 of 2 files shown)


src/core/src/local_loggers/FileLogger.py
```diff
@@ -38,8 +38,9 @@ def __del__(self):
 
     def write(self, message, fail_silently=True):
         try:
+            truncated_message = self.truncate_message(message)
```
kjohn-msft, line 41: There's variable assigment and re-assignment going on in every call to this function (almost all the time unnecessarily). Try to invoke the function only if the max_size exceeds.

if len(message) > max_size:
   message = message[:max_size]

It's unclear why the new line index matters, etc in the function as it's put now.  (raising the question of why the function is needed then too)

feng-j678: i've modified the code to avoid the var assignment by moving the max_msg_size check prior the function called, I still think we should include the newline index cut off because of it might cut off weird 3 bytes char which will cause json serialization error, unless there's no json serialization after, right?

feng-j678: done

kjohn-msft (approved):

rane-rajasi (approved):

PR #280, opened 2024-11-25
Fix/distutil deprecation

feng-j678:
[x] addressed distutil module deprecation in py312, deprecation is causing failure in vms with py312. to address this, created a customize comparator class to compare extension version numbers, extract the version numbers from the config path, and sort all existing extension versions to get the latest version. this comparator class is also being used to check ubuntu pro client machine versions.
image
[x] add ut for comparator class in core and extension
[x] core ut result:
image

[x] extension ut result:
image

image

PR #294, opened 2025-03-11
Fix the version comparator bug for ubuntu pro client version extraction

feng-j678:
[x] address the bug introduced in https://github.com/Azure/LinuxPatchExtension/pull/280/files
[x] address bug in ubuntu pro client vm where version comparator lib unable to extract the os version, so pro client query is not working:
image

core test
image

handler test
image

post code change:
image


src/core/src/core_logic/VersionComparator.py
```diff
@@ -39,10 +39,21 @@ def extract_version_nums(self, path):
...
         Extract the version part from a given path.
-        Input: /var/lib/waagent/Microsoft.CPlat.Core.LinuxPatchExtension-1.2.5/config
-        Return: "1.2.5"
+        Path	                    Extracted Version
+        /var/lib/waagent/Microsoft.CPlat.Core.LinuxPatchExtension-1.2.5/config "1.2.5"
+        "abc-1.2.3-alpha"	        "1.2.3"
+        "xyz-34~20.04"	            "34"
+        "some-27.14-ubuntu"	        "27.14"
+        "random-5.0.1+build"	    "5.0.1"
+        "abc"	                    ""
```
kjohn-msft, line 48: Please include the exact problem seen with pro client as an example here.

src/core/src/core_logic/VersionComparator.py
```diff
@@ -39,10 +39,21 @@ def extract_version_nums(self, path):
...
         """
-        match = re.search(r'([\d]+\.[\d]+\.[\d]+)', path)
+        # extract lpe version
+        if "LinuxPatchExtension" in path:
+            match = re.search(r'LinuxPatchExtension-(\d+(?:\.\d+)*)', path) # extract numbers with optional dot-separated parts
+            return match.group(1).rstrip('.') if match else ""
```
kjohn-msft, line 53: Too specific for a function generically named extract_version_nums.

It's okay for the code paths looking at LPE to have a separate function dedicated to that scenario if that's what's needed.
- like extract_version_nums_for_lpe

src/core/tests/Test_VersionComparator.py
```diff
@@ -25,27 +25,45 @@ def setUp(self):
...
+        self.assertEqual(self.version_comparator.extract_version_nums("Users\Smith~123\AppData\tmp5a42j2ua\Microsoft.CPlat.Core.LinuxPatchExtension-1.6.99"), "1.6.99")
+        self.assertEqual(self.version_comparator.extract_version_nums("Users\Smith~123\AppData\tmp5a42j2ua\Microsoft.CPlat.Core.LinuxPatchExtension-1.6"), "1.6")
+        self.assertEqual(self.version_comparator.extract_version_nums("Users\Smith~123\AppData\tmp5a42j2ua\Microsoft.CPlat.Core.LinuxPatchExtension-1.6."), "1.6")
+        self.assertEqual(self.version_comparator.extract_version_nums("Users\Smith~123\AppData\tmp5a42j2ua\Microsoft.CPlat.Core.LinuxPatchExtension-1.6.."), "1.6")
+        self.assertEqual(self.version_comparator.extract_version_nums("Users\Smith~123\AppData\tmp5a42j2ua\Microsoft.CPlat.Core.LinuxPatchExtension-1.6.abc"), "1.6")
+        self.assertEqual(self.version_comparator.extract_version_nums("Users\Smith~123\AppData\tmp5a42j2ua\Microsoft.CPlat.Core.LinuxPatchExtension-1.6abc"), "1.6")
+        self.assertEqual(self.version_comparator.extract_version_nums("Users\Smith~123\AppData\tmp5a42j2ua\Microsoft.CPlat.Core.LinuxPatchExtension-a.b.c"), "")
```
kjohn-msft, line 42: Why were the previous tests removed? Also what's the relevance of the LPE test cases here? This is not how the function is invoked in for the LPE cases..

src/core/tests/Test_VersionComparator.py
```diff
@@ -25,27 +25,45 @@ def setUp(self):
...
+        self.assertEqual(self.version_comparator.extract_version_nums("27.13.4-ab+18.04.1"), "27.13.4")
+        self.assertEqual(self.version_comparator.extract_version_nums("27.13.4abc-18.04.1"), "27.13.4")
+        self.assertEqual(self.version_comparator.extract_version_nums("27.13.4!@abc"), "27.13.4")
 
-        # Test compare versions logic
+
+        # Test compare versions logic Ubuntuproclient version
+        test_extracted_v1 = self.version_comparator.extract_version_nums("27.13.4~18.04.1")
+        test_extracted_v2 = self.version_comparator.extract_version_nums("27.13.4!27.1.2.3-ab+18.04.1")
+        test_extracted_v3 = self.version_comparator.extract_version_nums("27.13.4~25.6aa#@bc")
```
kjohn-msft, line 66: Where is the unit test for the specific problem in production this code's supposed to fix?

src/core/tests/Test_VersionComparator.py
```diff
@@ -25,27 +25,45 @@ def setUp(self):
...
+        self.assertEqual(self.version_comparator.extract_version_nums("27.1!abc-20"), "27.1")
 
-        self.assertEqual(test_extracted_v1, expected_extracted_version)
-        self.assertEqual(test_extracted_v2, expected_extracted_version)
-        self.assertEqual(test_extracted_v3, expected_extracted_version)
+        self.assertEqual(self.version_comparator.extract_version_nums("27.13.4"), "27.13.4")
+        self.assertEqual(self.version_comparator.extract_version_nums("27.13.4~18.04.1"), "27.13.4")
+        self.assertEqual(self.version_comparator.extract_version_nums("27.13.4-ab+18.04.1"), "27.13.4")
+        self.assertEqual(self.version_comparator.extract_version_nums("27.13.4abc-18.04.1"), "27.13.4")
+        self.assertEqual(self.version_comparator.extract_version_nums("27.13.4!@abc"), "27.13.4")
```
kjohn-msft, line 60: How were these patterns decided on? Is it from a test matrix?

src/core/src/core_logic/VersionComparator.py
feng-j678: i've update the extraction list to reflect the possible all versions

src/core/src/core_logic/VersionComparator.py
feng-j678: create 2 separate extraction methods - lpe and os

src/core/tests/Test_VersionComparator.py
feng-j678: the purposes of these tests are to determine possible lpe version inputs matching to the expected regex result, if they are then the existing testing (test_ubuntuproclient.py and test_actionhandler.py) that go through production code will work as expected

src/core/tests/Test_VersionComparator.py
feng-j678: the version comparator is to replace the unsupported distutil library, we shouldn't be modifying existing test code in test_actionhandler and test_ubuntuproclient. the ut in these 2 files calls the version comparator logic. all testing is passing, this indicates version comparator is working as expected. the reason why previous version comparator failed because test in ubuntuproclient version are 3 digits, and comparator checks only 3 digits. if the actual version is more or less, then comparator will fail. this new regex version checks for all possible digits. I also added 1 more unit test in test_ubuntuproclient, and i don't believe it does much because as long the comparator regex checks for all possible digit's format, we should be good.

src/core/tests/Test_VersionComparator.py
feng-j678: those are possible patterns I can think of

src/core/tests/Test_VersionComparator.py
```diff
@@ -24,53 +24,92 @@ class TestVersionComparator(unittest.TestCase):
...
-        # Test compare versions logic
-        self.assertEqual(self.version_comparator.compare_version_nums(test_extracted_v1, "27.13.4"), 0) # equal
-        self.assertEqual(self.version_comparator.compare_version_nums(test_extracted_v2, "27.13.3"), 1) # greater
-        self.assertEqual(self.version_comparator.compare_version_nums(test_extracted_v3, "27.13.5"), -1) # less
-
-        # Test sort versions logic
-        unsorted_path_versions = [
+    def test_linux_extension_version_extract_comparator(self):
+        """ Test extract version logic on Extension package """
+        self.assertEqual(self.version_comparator.extract_lpe_path_version_num("Users\Smith~123\AppData\tmp5a42j2ua\Microsoft.CPlat.Core.LinuxPatchExtension-1.2.25"), "1.2.25")
```
kjohn-msft, line 29: My question specifically was why is there this "Users\Smith~123\AppData\tmp5a42j2ua\" and similar in all the paths? Is this how the product code is invoking it?

src/core/src/core_logic/VersionComparator.py
```diff
@@ -35,38 +36,93 @@ def compare_version_nums(self, version_a, version_b):
...
         """ Extract version number from input and return int tuple.
-        Input: "Microsoft.CPlat.Core.LinuxPatchExtension-1.6.100"
+        Lpe input: "/var/lib/waagent/Microsoft.CPlat.Core.LinuxPatchExtension-1.6.100"
         Return: (1.6.100)
+
+        os version input: "34~18.04"
+        Return: (34)
         """
-        version_numbers = self.extract_version_nums(path)
+        if "LinuxPatchExtension" in version_input:
```
rane-rajasi, line 121: nit: Add this to Constants

src/core/tests/Test_VersionComparator.py
```diff
@@ -24,54 +24,109 @@ class TestVersionComparator(unittest.TestCase):
...
+
+    def test_linux_os_version_comparison(self):
+        """ Test compare versions logic Ubuntuproclient version with existing vm version """
+        test_extracted_good_version = self.version_comparator.extract_os_version_nums("34.13.4~18.04.1")  # return 34
+
+        self.assertEqual(self.version_comparator.compare_version_nums(test_extracted_good_version, "34.13.4"), 0)  # equal  34.13.4 == 34.13.4
+        self.assertEqual(self.version_comparator.compare_version_nums(test_extracted_good_version, "34.13.3"), 1)  # greater 34.13.4 > 34.13.3
+        self.assertEqual(self.version_comparator.compare_version_nums(test_extracted_good_version, "34.13.5"), -1)  # less 34.13.4 < 34.13.5
+
+        test_extracted_bad_version = self.version_comparator.extract_os_version_nums("abc~18.04.1")  # return ""
```
rane-rajasi, line 107: wouldn't this return 18.04.1? Assert on next line (108) is still correct because 18.04.1 < 34.13.4

rane-rajasi (changes requested):
comments inline

src/core/tests/Test_VersionComparator.py
feng-j678: updated the path to reflect the actual lpe path

src/core/src/core_logic/VersionComparator.py
feng-j678: done

src/core/tests/Test_VersionComparator.py
feng-j678: i'm not sure if such version would exist abc~18.04.1, my assumption is if it does happen, I will see that as invalid image version. also based on my understanding of pro client image 34~18.04.01, 34 is the new image version, 18.04.1 is the older version, so abc will be a bad version img

src/core/src/core_logic/VersionComparator.py
kjohn-msft: Not applicable to new code

src/core/tests/Test_VersionComparator.py
kjohn-msft: This is an invalid test - as the actual expectations for the test input is undefined (or even undefinable). Such an input should never be seen and as such isn't worthwhile expanding test case coverage for, unless the point is to raise an exception in product code for an undefined condition.

kjohn-msft (approved):

src/core/tests/Test_VersionComparator.py
```diff
@@ -24,54 +24,35 @@ class TestVersionComparator(unittest.TestCase):
...
+        self.assertEqual(self.version_comparator.extract_version_from_os_version_nums("34.1~18.04.1"), "34.1")
+        self.assertEqual(self.version_comparator.extract_version_from_os_version_nums("34.13.4"), "34.13.4")
+        self.assertEqual(self.version_comparator.extract_version_from_os_version_nums("34.13.4~18.04.1"), "34.13.4")
+        self.assertEqual(self.version_comparator.extract_version_from_os_version_nums("34.13.4-ab+18.04.1"), "34.13.4")
+        self.assertEqual(self.version_comparator.extract_version_from_os_version_nums("34.13.4abc-18.04.1"), "34.13.4")
+        self.assertEqual(self.version_comparator.extract_version_from_os_version_nums("abc.34.13.4!@abc"), "34.13.4")
+
+    def test_linux_os_version_comparison(self):
+        """ Test compare versions logic Ubuntuproclient version with existing vm version """
+        test_extracted_good_version = self.version_comparator.extract_version_from_os_version_nums("34.13.4~18.04.1")  # return 34
```
michellemcdaniel, line 47: comment looks wrong

michellemcdaniel (approved):

rane-rajasi (approved):

PR #296, opened 2025-03-14
[OE PR Comments] Version Comparator Comments as Code on PR Branch

kjohn-msft:
Merge into hotfix/version_comparator..  as a reflection of comments I have.

Use hotfix/version_comparator after that as final review with 2 reviewers, etc.


src/extension/src/ExtVersionComparator.py
```diff
@@ -92,7 +46,29 @@ def sort_versions_desc_order(self, version_list):
             "32.1~18.04.01",
             "32~18.04.01"]
```
feng-j678, line 47: since we are not using this logic for os version comparison in handler, then we need to remove the os version comment

src/extension/src/ExtVersionComparator.py
```diff
@@ -103,9 +79,5 @@ def __version_key(self, version_input):
         os version input: "34~18.04"
```
feng-j678, line 79: since we are not using this logic for os version comparison in handler, then we need to remove the os version comment

feng-j678 (changes requested):

PR #298, opened 2025-03-18
BugFix: CentOS VMs not installing patches during Auto Patching

rane-rajasi:
Fixes included in this PR:
1. Do not use MaintenanceRunId to validate an incoming Auto Patching request
2. MaxPatchPublishDate fails if healthstoreid contains additional '_'. For eg: Ideal healthstoreid "pub_off_sku_2020.09.29" is supported currently but if a healthstoreid, say "pub_off_sk_u_2020.09.29", comes in, maxpatchpublishdate calculation fails

### DIFF
--- a/src/core/src/core_logic/ExecutionConfig.py
+++ b/src/core/src/core_logic/ExecutionConfig.py
@@ -108,9 +108,9 @@ class ExecutionConfig(object):
         """ Obtains implicit date ceiling for published date - converts pub_off_sku_2024.04.01 to 20240401T000000Z """
         max_patch_publish_date = str()
         if health_store_id is not None and health_store_id != "":
-            split = health_store_id.split("_")
-            if len(split) == 4 and len(split[3]) == 10:
-                max_patch_publish_date = "{0}T000000Z".format(split[3].replace(".", ""))
+            split = health_store_id.rsplit("_", 1)
+            if len(split) == 2 and len(split[1]) == 10:
+                max_patch_publish_date = "{0}T000000Z".format(split[1].replace(".", ""))
 
         self.composite_logger.log_debug("[EC] Getting max patch publish date. [MaxPatchPublishDate={0}][HealthStoreId={1}]".format(str(max_patch_publish_date), str(health_store_id)))
         return max_patch_publish_date
--- a/src/core/src/core_logic/ExecutionConfig.py
+++ b/src/core/src/core_logic/ExecutionConfig.py
@@ -108,9 +108,9 @@ class ExecutionConfig(object):
         """ Obtains implicit date ceiling for published date - converts pub_off_sku_2024.04.01 to 20240401T000000Z """
         max_patch_publish_date = str()
         if health_store_id is not None and health_store_id != "":
-            split = health_store_id.rsplit("_", 1)
-            if len(split) == 2 and len(split[1]) == 10:
-                max_patch_publish_date = "{0}T000000Z".format(split[1].replace(".", ""))
+            date_candidate = str(health_store_id)[-10:].replace(".", "")  # last 10 characters and remove '.'
+            if len(date_candidate) == 8 and date_candidate.isdigit() and str(health_store_id)[-11:-10] == "_":
+                max_patch_publish_date = date_candidate + "T000000Z"
 
         self.composite_logger.log_debug("[EC] Getting max patch publish date. [MaxPatchPublishDate={0}][HealthStoreId={1}]".format(str(max_patch_publish_date), str(health_store_id)))
         return max_patch_publish_date
--- a/src/core/src/package_managers/YumPackageManager.py
+++ b/src/core/src/package_managers/YumPackageManager.py
@@ -84,7 +84,7 @@ class YumPackageManager(PackageManager):
 
         # if an Auto Patching request comes in on a CentOS machine with Security and/or Critical classifications selected, we need to install all patches
         installation_included_classifications = [] if execution_config.included_classifications_list is None else execution_config.included_classifications_list
-        if execution_config.maintenance_run_id is not None and execution_config.operation.lower() == Constants.INSTALLATION.lower() \
+        if execution_config.health_store_id is not str() and execution_config.operation.lower() == Constants.INSTALLATION.lower() \
                 and 'CentOS' in str(env_layer.platform.linux_distribution()) \
                 and 'Critical' in installation_included_classifications and 'Security' in installation_included_classifications:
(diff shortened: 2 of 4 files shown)


src/core/src/package_managers/YumPackageManager.py
```diff
@@ -84,7 +84,7 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
 
         # if an Auto Patching request comes in on a CentOS machine with Security and/or Critical classifications selected, we need to install all patches
         installation_included_classifications = [] if execution_config.included_classifications_list is None else execution_config.included_classifications_list
-        if execution_config.maintenance_run_id is not None and execution_config.operation.lower() == Constants.INSTALLATION.lower() \
+        if execution_config.health_store_id is not str() and execution_config.operation.lower() == Constants.INSTALLATION.lower() \
```
rane-rajasi, line 87: @kjohn-msft, I've used healthstoreid for auto patch validation instead of maxpatchpublishdate since the presence of healthstoreid indicates an AutoPatch request vs maxpatchpublishdate can also come in as an API input

src/core/src/core_logic/ExecutionConfig.py
```diff
@@ -108,9 +108,9 @@ def __get_max_patch_publish_date(self, health_store_id):
         """ Obtains implicit date ceiling for published date - converts pub_off_sku_2024.04.01 to 20240401T000000Z """
         max_patch_publish_date = str()
         if health_store_id is not None and health_store_id != "":
-            split = health_store_id.split("_")
-            if len(split) == 4 and len(split[3]) == 10:
-                max_patch_publish_date = "{0}T000000Z".format(split[3].replace(".", ""))
+            split = health_store_id.rsplit("_", 1)
+            if len(split) == 2 and len(split[1]) == 10:
+                max_patch_publish_date = "{0}T000000Z".format(split[1].replace(".", ""))
```
kjohn-msft, line 113: Can you please write a debug log if it reaches this far and the conditions to convert successfully aren't met. e.g. split happens and then the data doesn't look right.

src/core/tests/Test_CoreMain.py
```diff
@@ -526,6 +525,55 @@ def test_install_all_packages_for_centos_autopatching_as_warning_with_never_rebo
 
         LegacyEnvLayerExtensions.LegacyPlatform.linux_distribution = backup_envlayer_platform_linux_distribution
 
+    def test_install_all_packages_for_centos_autopatching_differentversionformat_in_healthstore(self):
```
kjohn-msft, line 528: Please unit test (in the appropriate test class / not core main) various call combinations (input->output) into the private method __get_max_patch_publish_date(self, health_store_id)

kjohn-msft (changes requested):
Comments inline

src/core/src/core_logic/ExecutionConfig.py
```diff
@@ -108,9 +108,9 @@ def __get_max_patch_publish_date(self, health_store_id):
         """ Obtains implicit date ceiling for published date - converts pub_off_sku_2024.04.01 to 20240401T000000Z """
         max_patch_publish_date = str()
         if health_store_id is not None and health_store_id != "":
-            split = health_store_id.split("_")
-            if len(split) == 4 and len(split[3]) == 10:
-                max_patch_publish_date = "{0}T000000Z".format(split[3].replace(".", ""))
+            split = health_store_id.rsplit("_", 1)
+            if len(split) == 2 and len(split[1]) == 10:
```
feng-j678, line 112: is there a case if health_store_id is like this pub_off_sk_u2020.09.29, then wouldn't the rsplit produce [pub_off_sk_, u2020.09.29]?

is it possible to add a debug on ln 113 smth like "debug("Parse health_store_id into list. [HealthStoreIdList={0}]".format(str(split)))

feng-j678 (approved):
adding debug to ln 113 is not required, but nice to have in case of future failures relate to health_store_id format

src/core/src/core_logic/ExecutionConfig.py
rane-rajasi: In this scenario, line 115 would log -> [EC] Getting max patch publish date. [MaxPatchPublishDate=][HealthStoreId=pub_off_sku_2020.9.29]

src/core/tests/Test_CoreMain.py
rane-rajasi: done

src/core/src/core_logic/ExecutionConfig.py
rane-rajasi: See my response to the other comment

rane-rajasi:
@feng-j678, @kjohn-msft code coverage shows changes in this PR are missing coverage, however the detailed report does not have anything missing within these changes. Why is this happening?
image

image

src/core/src/core_logic/ExecutionConfig.py
kjohn-msft: Sounds good - closing.

feng-j678 (approved):

src/core/src/core_logic/ExecutionConfig.py
michellemcdaniel: wait, you want to use date_candadte if it's not a digit and not len 8?

kjohn-msft: Sorry, was writing that comment out while in a call. @michellemcdaniel edited inline - updated line is if len(date_candidate) == 8 and date_candidate.isdigit():

src/core/tests/Test_CoreMain.py
kjohn-msft: What was meant by this (invoke private method, with call combinations) is in this PR to your branch --
https://github.com/Azure/LinuxPatchExtension/pull/300/files

We should try to use these techniques to make our test code more maintainable (existing test code - @feng-j678 for visibility). Specifically having the minimum code to exercise the unit being tested. Vastly helps with initial review, later changes, etc.

src/core/src/core_logic/ExecutionConfig.py
michellemcdaniel: That makes much more sense

src/core/tests/Test_CoreMain.py
feng-j678: yes the coverage for this pr is high

image

kjohn-msft (approved):

feng-j678 (approved):

PR #305, opened 2025-04-10
Feature: Increase max OS version to support Ubuntu Pro Client 22.04

feng-j678:
[x] Increase max os version to support for ubuntu pro client 22.04
[x] core ut:
image
[x] handler ut:
image
image


kjohn-msft:
The screenshot in the PR actually shows 1 of 2 reasons why we were holding back expanding it - 
image

And one part of a 2+ part customer CRI generator from last year and ongoing. Which is why this is being done --
https://github.com/Azure/LinuxPatchExtension/pull/308

kjohn-msft:
Cancelling PR as coverage exists already.

PR #273, opened 2024-10-21
Apt debstyle882 support, bug-fixes in cache evaluation & manipulation

kjohn-msft:
1. Moves away from using UUID for custom sources because it makes diagnosis very hard - moving to a fixed timestamp model + unique characteristics model to help with logs.
2. Adds additional logging to help understand what the cache state is at any point.
3. Adds full support for debstyle882 sources.
4. Adds full support for sourceparts to ensure support in Ubuntu 23.10, 24.04, 24.10+.
5. Reworked getting custom sources to spec to better and correctly benefit from the cached state.
6. Snapshot support is integrated deeper into the stack and supports more repo types as part of this rework.
7. Security repo support extends fully to debstyle882 sources.
8. Pro client logging is improved to help with clarity on discrepancies in lists.
9. Integrates relevant code from the code debt erasure PR to align with the north star monitoring, alerting and investigation story.
          - Stable code is deprioritized.
          - Log origination is made clear with tagging & without having to look at code.
          - IDE hinting on types are introduced.
          - Customer environment errors are called out explicitly as such.

PR #308, opened 2025-04-14
Bugfix: Mitigate external Ubuntu Pro Client issues

kjohn-msft:
**Background:** For VMs that are Ubuntu Pro client capable, there are 2 sets of issues that manifest as required security updates not getting installed on machines_ (some machines).

1. Canonical is aware of packages normally seen as updates in the default flow not being shown as required in pro client scans. There is an explanation for this but the way we have taken a tight dependency on pro client when it's functional causes these updates no to get installed.

2. There are cases where wide swathes of security updates are not getting detected by pro client. It was not clear if this was a pro client issue in the past or an issue with our code. The additional code that went in November helped identify that this was a pro client issue when newer reports came in: https://github.com/Azure/LinuxPatchExtension/pull/273

Both problems listed above are being resolved by not fully relying on pro client and using a combined overlay of the default scanning mechanism with whatever pro client reports. This is the 'best of both worlds' approach. Extensive logging additions will help further reviews with Canonical on pro client behaviors without affecting any customer while a multi-stage resolution is ironed out.

### DIFF
--- a/src/core/src/package_managers/AptitudePackageManager.py
+++ b/src/core/src/package_managers/AptitudePackageManager.py
@@ -358,6 +358,7 @@ class AptitudePackageManager(PackageManager):
 
         # Query pro client if prerequisites are met
         if self.__pro_client_prereq_met:
+            self.refresh_repo()
             self.composite_logger.log_verbose("[APM-Pro][Sec] Discovering 'security' packages (pro client)...")
             ubuntu_pro_client_security_updates_query_success, ubuntu_pro_client_security_packages, ubuntu_pro_client_security_package_versions = self.ubuntu_pro_client.get_security_updates()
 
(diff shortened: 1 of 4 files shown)


src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -348,28 +348,46 @@ def get_security_updates(self):
...
-            if len(pro_client_missed_updates) > 0:       # not good, needs investigation
-                self.composite_logger.log_debug("[APM-Pro][Sec][!] Pro client missed updates found. [Count={0}][Updates={1}]".format(len(pro_client_missed_updates), pro_client_missed_updates))
-            if len(sec_updates_missed_updates) > 0:     # interesting, for review
-                self.composite_logger.log_debug("[APM-Pro][Sec] Pro client only updates found. [Count={0}][Updates={1}]".format(len(sec_updates_missed_updates), sec_updates_missed_updates))
 
-        if ubuntu_pro_client_security_updates_query_success:     # this needs to be revisited based on logs above
-            return ubuntu_pro_client_security_packages, ubuntu_pro_client_security_package_versions
-        else:
+        # Only use non-pro client results if either pre-reqs are not met or if the query fails
+        if not ubuntu_pro_client_security_updates_query_success:
```
feng-j678, line 366: what about adding an additional condition checking len(ubuntu_pro_client_security_packages) > 0 if there's no ubuntu_pro_client_security_packages  we just return security_packages, security_package_versions, this helps avoiding and computation from line 370-389?

if not ubuntu_pro_client_security_updates_query_success: or len(ubuntu_pro_client_security_packages) <= 0

feng-j678 (approved):

rane-rajasi (approved):

PR #310, opened 2025-04-16
Bugfix: Reboot Manager behavior - multiple bugfixes & error rate reduction

kjohn-msft:
Fixing issues seen while testing:
[Bugfix: Mitigate external Ubuntu Pro Client issues](https://github.com/Azure/LinuxPatchExtension/pull/308)

The main issue was that my test infra for that code failed to complete patch installation with this error message:
_"Reboot failed to proceed on the machine in a timely manner."_

The machine then failed to accept ssh logins for 20+ minutes. I thought the machine was dead, but it eventually came back suggesting a much longer reboot time than expected. Tracing this further in production, it suggested there were 500+ operations per week hitting the same issue and getting reported as a terminal failure.

This prompted a closer look at all the Reboot Manager code, and PR addresses every issue identified with it.

The changes in this PR:
1. Differentiating between all of the following:
(a) Reboot buffer in minutes = minimum time required to consider a reboot. This was being overloaded incorrectly to also broadcast the time delay to starting the reboot, which _could_ cause silent maintenance window exceeds. 
(b) Reboot notify timeout in minutes = introduced new to be deliberate about the duration of the notification window.
(c) Reboot wait timeout in minutes (min & max) = time duration to wait before considering an attempt to reboot a failure.

2. The effective reboot wait timeout is dynamic now - it's sits somewhere between the min and max allowed and uses as much as the remaining time in the maintenance window allows to allow for success.

3. Reboot manager code has been cleaned up to meet current coding standards for the code base.

4. Reduce likelihood of timeout at the Compute RP by refreshing the status as long as we're still waiting for a reboot and our process is still running. 

The configured values may be adjusted in the future to account for what is seen at scale.

### DIFF
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -197,7 +197,7 @@ class Constants(object):
     RED_HAT = 'Red Hat'
     SUSE = 'SUSE'
     CENTOS = 'CentOS'
-    AZURE_LINUX = 'Microsoft Azure Linux'
+    AZURE_LINUX = ['Microsoft Azure Linux', 'Common Base Linux Mariner']
 
     # Package Managers
     APT = 'apt'
@@ -382,7 +382,7 @@ class Constants(object):
     class UbuntuProClientSettings(EnumBackport):
         FEATURE_ENABLED = True
         MINIMUM_PYTHON_VERSION_REQUIRED = (3, 5)  # using tuple as we can compare this with sys.version_info. The comparison will happen in the same order. Major version checked first. Followed by Minor version.
-        MAX_OS_MAJOR_VERSION_SUPPORTED = 20
+        MAX_OS_MAJOR_VERSION_SUPPORTED = 24
         MINIMUM_CLIENT_VERSION = "27.14.4"
 
     class BufferMessage(EnumBackport):
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -31,7 +31,7 @@ class Constants(object):
     UNKNOWN = "Unknown"
 
     # Extension version (todo: move to a different file)
-    EXT_VERSION = "1.6.58"
+    EXT_VERSION = "1.6.59"
 
     # Runtime environments
     TEST = 'Test'
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -261,10 +261,11 @@ class Constants(object):
         'IfRequired': REBOOT_IF_REQUIRED,
         'Always': REBOOT_ALWAYS
     }
-    REBOOT_BUFFER_IN_MINUTES = 15           # minimum MW time required to consider rebooting if required (notify - 3, wait - 7, machine - 5)
-    REBOOT_NOTIFY_TIMEOUT_IN_MINUTES = 3       # time to broadcast reboot notification to other processes
-    REBOOT_WAIT_TIMEOUT_IN_MINUTES_MIN = 7     # minimum time to wait for a reboot to have started in the current execution context
-    REBOOT_WAIT_TIMEOUT_IN_MINUTES_MAX = 40    # maximum possible** time to wait for a reboot to have started in the current execution context (**IF MW time remaining allows it)
+    REBOOT_BUFFER_IN_MINUTES = 15                # minimum MW time required to consider rebooting if required (notify - 3, wait - 7, machine - 5)
+    REBOOT_NOTIFY_TIMEOUT_IN_MINUTES = 3         # time to broadcast reboot notification to other processes
+    REBOOT_WAIT_TIMEOUT_IN_MINUTES_MIN = 7       # minimum time to wait for a reboot to have started in the current execution context
+    REBOOT_WAIT_TIMEOUT_IN_MINUTES_MAX = 40      # maximum possible** time to wait for a reboot to have started in the current execution context (**IF MW time remaining allows it)
+    REBOOT_TO_MACHINE_READY_TIME_IN_MINUTES = 5  # time to wait for the machine to be ready after a reboot actually happens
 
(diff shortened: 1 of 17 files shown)


feng-j678 (approved):

src/core/src/core_logic/RebootManager.py
```diff
@@ -14,116 +14,167 @@
...
+                self.__maintenance_window_exceeded_flag = True
                 return False
 
-    def is_reboot_pending(self):
-        return self.package_manager.force_reboot or (self.status_handler and self.status_handler.is_reboot_pending)
+        # No-op - This code should never be reached. If seen, it indicates a bug in the code.
+        self.composite_logger.log_error('[RM] Bug-check: Unexpected code branch reached. [RebootSetting={0}][RebootPending={1}]'.format(str(self.__reboot_setting), str(reboot_pending)))
+        return False
+
+    def __start_reboot(self, message="Azure VM Guest Patching initiated a reboot after an 'InstallPatches' operation.", maintenance_window_available_time_in_minutes=0):
```
rane-rajasi, line 127: Not necessarily "after an 'InstallPatches' operation". Can happen at the start or in between too

src/core/src/core_logic/RebootManager.py
```diff
@@ -14,116 +14,167 @@
...
+        # type: () -> bool
+        """ Check if a reboot is pending either from the package manager or the status handler """
+        return self.package_manager.force_reboot or (self.status_handler and self.status_handler.is_reboot_pending)
+
+    def has_maintenance_window_exceeded_at_reboot_manager(self):
+        # type: () -> bool
+        """ Check if the maintenance window was exceeded at reboot manager """
+        return self.__maintenance_window_exceeded_flag
+
+    def get_reboot_setting_sanitized(self):
```
rane-rajasi, line 55: Why is this named get_reboot_setting_**sanitized** if it doesn't process/sanitize/format self.__reboot_setting ?

src/core/src/core_logic/RebootManager.py
```diff
@@ -14,116 +14,167 @@
...
+            # 3. The time required to handle changes prior to reboot is greater than the time we've allocated. (action on AzGPS if seen at scale in Azure)
+            self.status_handler.set_installation_reboot_status(Constants.RebootStatus.FAILED)
+            error_msg = "Customer environment issue: Reboot failed to proceed on the machine in a timely manner. Please retry the operation."
+            self.status_handler.add_error_to_status(error_msg, Constants.PatchOperationErrorCodes.DEFAULT_ERROR)
+            raise Exception(error_msg, "[{0}]".format(Constants.ERROR_ADDED_TO_STATUS))
+
+    def __reboot_wait_pulse(self, elapsed_time_in_minutes, max_allowable_time_to_reboot_in_minutes, maintenance_window_available_time_in_minutes):
+        # type: (int, int, int) -> None
+        self.composite_logger.log("[RM] Waiting for machine reboot. [ElapsedTimeInMinutes={0}][MaxTimeInMinutes={1}][MaintenanceWindowAllowable={2}]"
+                                  .format(str(elapsed_time_in_minutes), str(max_allowable_time_to_reboot_in_minutes), str(maintenance_window_available_time_in_minutes - elapsed_time_in_minutes)))
```
rane-rajasi, line 161: str(maintenance_window_available_time_in_minutes - elapsed_time_in_minutes)

This should come from the function call (i.e. line 145) for maintenance_window_available_time_in_minutes

src/core/src/bootstrap/Constants.py
```diff
@@ -261,8 +261,12 @@ class PackageClassification(EnumBackport):
         'IfRequired': REBOOT_IF_REQUIRED,
         'Always': REBOOT_ALWAYS
     }
-    REBOOT_BUFFER_IN_MINUTES = 15
-    REBOOT_WAIT_TIMEOUT_IN_MINUTES = 5
+    REBOOT_BUFFER_IN_MINUTES = 15                # minimum MW time required to consider rebooting if required (notify - 3, wait - 7, machine - 5)
+    REBOOT_NOTIFY_TIMEOUT_IN_MINUTES = 3         # time to broadcast reboot notification to other processes
```
rane-rajasi, line 265: This reads as it is the timeout for notification but is also used as time to reboot or shutdown. Should be renamed to REBOOT_IN_MINUTES or REBOOT_VM_IN_MINUTES?

rane-rajasi (changes requested):
Comments inline

src/core/src/core_logic/RebootManager.py
kjohn-msft: The variable itself is set sanitized and protected. The function name here is to reflect that status outward.
image

The ideal thing to do would be to name the variable to match that.

src/core/src/core_logic/RebootManager.py
kjohn-msft: Good callout; will adjust the text.

src/core/src/bootstrap/Constants.py
kjohn-msft: Updated to this -- REBOOT_NOTIFY_WINDOW_IN_MINUTES = 3          # time to broadcast reboot notification to other processes as part of the reboot command

It's definitely not a timeout so removed that. Updated the comment to explain where it's used.

feng-j678 (dismissed):

src/core/src/core_logic/RebootManager.py
kjohn-msft: Text is now: "Azure VM Guest Patching initiated a reboot as part of an 'InstallPatches' operation."

src/core/src/core_logic/RebootManager.py
kjohn-msft: Addressed and renamed variable for clarity

rane-rajasi (approved):

feng-j678 (approved):

PR #317, opened 2025-04-28
Bugfix: Apply SSL expired cmd fix to RHEL images only

feng-j678:
[x] apply ssl expired cmd fix only to rhel images
[x] pending unit tests
image


src/core/src/bootstrap/EnvLayer.py
```diff
@@ -298,7 +298,7 @@ def linux_distribution(self):
             else:
                 code, output = self.__read_record(operation)
                 return eval(output)
-
+
```
kjohn-msft, line 301: Revert - avoid file touches that are not really changes

PR #318, opened 2025-04-30
Feature: Azure Linux 3.0: Enhanced Support

rane-rajasi:
This PR contains changes to support strict SDP in Azure Linux.

**How it works**:
Strict SDP in auto patching ensures the same set of patches are updated in all regions during a rollout. This is implemented using max_patch_publish_date. An incoming PatchInstallation request, with strict sdp, will receive a max_patch_publish_date, we need to fetch all patches published on or before this date and update them if necessary.

**Pending**:
[x] BVT test plan

**In VM Tests**:
[x] Azure Linux 3.0 base support: [2.core.log](https://github.com/user-attachments/files/21928845/2.core.log) [2.status.txt](https://github.com/user-attachments/files/21928848/2.status.txt)
[x] Strict SDP support restricted to Azure Linux 3.0 with min requirements (tdnf support)
        &nbsp;&nbsp;&nbsp;&nbsp; 1. don't run SDP on 2. (which would have corresponding in vm tests) - [3.core.log](https://github.com/user-attachments/files/21928935/3.core.log) [3.status.txt](https://github.com/user-attachments/files/21928937/3.status.txt)
        &nbsp;&nbsp;&nbsp;&nbsp; 2. if AzLinux 3 with min requirements not met (in vm tests):
                  &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; 2.1. not achievable - fail - [3.core.log](https://github.com/user-attachments/files/21928904/3.core.log) [3.status.txt](https://github.com/user-attachments/files/21928907/3.status.txt)
                  &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; 2.2 achievable - make the prerequisites met [2.core.log](https://github.com/user-attachments/files/21928909/2.core.log) [2.status.txt](https://github.com/user-attachments/files/21928912/2.status.txt)
                            &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; - if that fails, fail - refer to results in 2.1
        &nbsp;&nbsp;&nbsp;&nbsp; 3. if AzLinux with min requirements met (either through action or by default), perform action - (in vm tests)  [2.core.log](https://github.com/user-attachments/files/21928921/2.core.log) [2.status.txt](https://github.com/user-attachments/files/21928923/2.status.txt)


src/core/src/package_managers/TdnfPackageManager.py
```diff
@@ -89,6 +94,39 @@ def refresh_repo(self):
         self.invoke_package_manager(self.cmd_clean_cache)
         self.invoke_package_manager(self.cmd_repo_refresh)
 
+    # region Fetch SnapshotTime
+    def __get_posix_time_from_health_store_id(self, execution_config):
+        """Converts date str received (in format: yyyy.mm.dd) to POSIX time string"""
+        # eg: Input: 2024.12.20 (str) -> Output: 1734681600 (str)
+        posix_time = str()
+        health_store_id = execution_config.health_store_id
+        date_in_health_store_id = execution_config.validate_date_in_health_store_id(health_store_id)
```
kjohn-msft, line 103: Breaks separation of concerns. I'd get what's needed here from what is already persisted in ExecutionConfig as maxpatchpublishdate.

Because maxpatchpublishdate can be set in other ways and include more than just a date with no time. 

The package manager does not need to know anything beyond the fully qualified timestamp (including HHMMSS) - the fact that they are 00000 in default patching is incidental and not to be defaulted to.

rane-rajasi: My understanding here was: maxpatchpublishdate can contain value set from inclusions. Default Patching sends in healtstoreId, that's the reason behind using healthstoreId here to compute posix time.

kjohn-msft: Maxpatchpublishdate is level of abstraction visible outside. Everything below that is supposed to be invisible to package managers for SDP (where the data comes from, what it contains before being surfaced as maxpatchpublishdate, etc.).

kjohn-msft: Specifically look at how this is implemented in Apt - APM has zero knowledge of where maxpatchpublishdate is coming from at any point. There's a value there or there isn't - and it accommodates both and behaves accordingly. Healthstore id is completely an unknown entity to the specialized APM class. It means nothing to the level of knowledge it needs to have.

rane-rajasi: @kjohn-msft please review the latest changes

kjohn-msft:
Marking todos discussed offline:

- [ ] Distinguish between AzLinux 2.0 and 3.0. Validate in-VM with new code to fork behaviors.
- [ ] Distinguish between min-support prereqs within 3.0 and a VM without it in code.
- [ ] Implement in-VM update path to get minimum prereqs (without breaking SDP) to get to a supported state to continue the operation as strictly SDP-based. Test in-VM on pre-req-meeting VM and non-pre-req-meeting VM.
- [ ] Manage BVT test plan for the above mainline conditions.

src/core/src/package_managers/TdnfPackageManager.py
```diff
@@ -89,6 +94,31 @@ def refresh_repo(self):
...
+        datetime_to_convert_format = '%Y%m%dT%H%M%SZ'
+        self.composite_logger.log_debug("[TDNF] Getting POSIX time from given datetime. [DateTimeToConvert={0}][DateTimeStringFormat={1}]".format(str(datetime_to_convert), datetime_to_convert_format))
+        try:
+            if datetime_to_convert != str():
+                posix_time = env_layer.datetime.datetime_string_to_posix_time(datetime_to_convert, datetime_to_convert_format)
+        except Exception as error:
+            self.composite_logger.log_debug("[TDNF] Could not fetch POSIX time from given datetime. [DateTimeToConvert={0}][DateTimeStringFormat={1}][ComputedPosixTime={2}][Error={3}]".format(str(datetime_to_convert), datetime_to_convert_format, posix_time, repr(error)))
+
+        self.composite_logger.log_debug("[TDNF] Computed POSIX time from given datetime. [DateTimeToConvert={0}][DateTimeStringFormat={1}][ComputedPosixTime={2}]".format(str(datetime_to_convert), datetime_to_convert_format, posix_time))
+        return posix_time
```
kjohn-msft, line 110: Isn't this a terminal failure? Why would we continue forward breaking SDP?

src/core/src/package_managers/TdnfPackageManager.py
```diff
@@ -33,14 +35,17 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
         self.cmd_clean_cache = "sudo tdnf clean expire-cache"
         self.cmd_repo_refresh = "sudo tdnf -q list updates"
 
+        # fetch snapshottime from health_store_id
+        self.snapshot_posix_time = self.__get_posix_time(execution_config.max_patch_publish_date, env_layer)
```
kjohn-msft, line 39: self.env_layer is being set in the parent PackageManager

The pattern in the entire code base is to use that directly instead of passing it in as arguments within the scope of the class.

kjohn-msft: Also, why is this being done in the constructor prior to configurepatching, assessment, etc.?

If the conversion fails, isn't it better to be specifically scoped to within the installpatches context?

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -534,6 +534,14 @@ def utc_to_standard_datetime(utc_datetime):
         def standard_datetime_to_utc(std_datetime):
             """ Converts datetime object to string of format '"%Y-%m-%dT%H:%M:%SZ"' """
             return std_datetime.strftime("%Y-%m-%dT%H:%M:%SZ")
+
+        @staticmethod
+        def datetime_string_to_posix_time(datetime_string, format_string):
+            """ Converts string of given format to posix datetime string. """
```
kjohn-msft, line 540: Include type hinting

src/core/src/package_managers/TdnfPackageManager.py
```diff
@@ -33,14 +35,17 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
...
-        self.single_package_check_installed = 'sudo tdnf list installed <PACKAGE-NAME>'
-        self.single_package_upgrade_simulation_cmd = 'sudo tdnf install --assumeno --skip-broken '
+        self.tdnf_check = self.__generate_command_with_snapshottime('sudo tdnf -q list updates <SNAPSHOTTIME>', self.snapshot_posix_time)
+        self.single_package_check_versions = self.__generate_command_with_snapshottime('sudo tdnf list available <PACKAGE-NAME> <SNAPSHOTTIME>', self.snapshot_posix_time)
+        self.single_package_check_installed = self.__generate_command_with_snapshottime('sudo tdnf list installed <PACKAGE-NAME> <SNAPSHOTTIME>', self.snapshot_posix_time)
+        self.single_package_upgrade_simulation_cmd = self.__generate_command_with_snapshottime('sudo tdnf install --assumeno --skip-broken <SNAPSHOTTIME>', self.snapshot_posix_time)
 
         # Install update
-        self.single_package_upgrade_cmd = 'sudo tdnf -y install --skip-broken '
+        self.single_package_upgrade_cmd = self.__generate_command_with_snapshottime('sudo tdnf -y install --skip-broken <SNAPSHOTTIME>', self.snapshot_posix_time)
```
kjohn-msft, line 48: self.__ ? Are these expected to be accessible outside the class?

src/core/src/package_managers/TdnfPackageManager.py
```diff
@@ -89,6 +94,31 @@ def refresh_repo(self):
         self.invoke_package_manager(self.cmd_clean_cache)
         self.invoke_package_manager(self.cmd_repo_refresh)
 
+    # region Strict SDP using SnapshotTime
+    def __get_posix_time(self, datetime_to_convert, env_layer):
```
kjohn-msft, line 98: Not convinced this function needs to exist at all. 

I'd just write this in the constructor** where the function is currently called:

self.snapshot_posix_time = None if self.execution_config.maxpatchpublishdate == str() else self.env_layer.datetime_string_to_posix_time ...

If the conversion fails, it's okay to fail.

** = the other comment of that conversion only getting done in the install patches flow to avoid failures in the CP, AP.

src/core/src/package_managers/TdnfPackageManager.py
kjohn-msft: Assessment is not supposed to be affected by being part of an SDP install patches flow.
It'll just yo-yo the assessment information between that operation and the next CP, AP or non-SDP IP operation.
Please review the Apt implementation a bit closer for the strict SDP parallels.

src/core/src/package_managers/TdnfPackageManager.py
```diff
@@ -89,6 +94,31 @@ def refresh_repo(self):
...
+        return posix_time
+
+    @staticmethod
+    def __generate_command_with_snapshottime(command_template, snapshotposixtime=str()):
+        # type: (str, str) -> str
+        """ Prepares a standard command to use snapshottime."""
+        if snapshotposixtime == str():
+            return command_template.replace('<SNAPSHOTTIME>', str())
+        else:
+            return command_template.replace('<SNAPSHOTTIME>', ('--snapshottime={0}'.format(str(snapshotposixtime))))
```
kjohn-msft, line 119: Something to review especially because these commands need to be run differently based on:
- Am I 2.0?
- Am I 3.0 without support?
- Am I 3.0 with support running an operation that shouldn't see any of this? (CP, AP)
- Am I 3.0 with support running an operation sub-flow that should see this? (IP) - this is the only path that should see this; vs. all paths right now.

In simplifying achieving the state behavior variations above, consider if generation is truly just appending the param to existing supported functions, and if that helps simplify some aspects of this.

src/core/tests/Test_TdnfPackageManager.py
```diff
@@ -41,6 +41,55 @@ def mock_write_with_retry_raise_exception(self, file_path_or_handle, data, mode=
...
+        self.assertTrue(package_manager.snapshot_posix_time is str())
+
+        # health_store_id is random string
+        self.runtime.stop()
+        argument_composer = ArgumentComposer()
+        argument_composer.health_store_id = "pub_offer_sku_20.312.543"
+        self.runtime = RuntimeCompositor(argument_composer.get_composed_arguments(), True, Constants.TDNF)
+        self.container = self.runtime.container
+        package_manager = self.container.get('package_manager')
+        self.assertTrue(package_manager.snapshot_posix_time is str())
```
kjohn-msft, line 70: Please simplify with input / output matrix and deduped code.

src/core/tests/Test_TdnfPackageManager.py
```diff
@@ -41,6 +41,55 @@ def mock_write_with_retry_raise_exception(self, file_path_or_handle, data, mode=
...
+
+    def test_valid_datetime_to_convert_to_posix_time(self):
+        # valid health_store_id
+        self.runtime.stop()
```
kjohn-msft, line 74: why is runtime.stop sitting at the top of the test code block here and in other places?

kjohn-msft (changes requested):
I'd suggest moving up a level in the execution flow to delineate which code is strictly required to run in which scenario and isolate accordingly. The lower-level details can then be re-filled in incrementally with that clarity set.

src/core/tests/Test_TdnfPackageManager.py
```diff
@@ -41,6 +41,55 @@ def mock_write_with_retry_raise_exception(self, file_path_or_handle, data, mode=
...
 
+    def test_invalid_health_store_id_posix_time(self):
+        # health_store_id is None
+        self.runtime.stop()
+        argument_composer = ArgumentComposer()
+        argument_composer.health_store_id = None
+        self.runtime = RuntimeCompositor(argument_composer.get_composed_arguments(), True, Constants.TDNF)
+        self.container = self.runtime.container
+        package_manager = self.container.get('package_manager')
+        self.assertTrue(package_manager.health_store_id_in_posix_time is str())
```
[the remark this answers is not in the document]
kjohn-msft: Not reflective of the intent here.

src/core/src/package_managers/TdnfPackageManager.py
rane-rajasi: @kjohn-msft I think snapshottime should also include snapshottime, because the list of patches available changes based on snapshottime and this is majorly be relevant in later SDP stages. For eg: A rollout starting 5/23, will have that snapshottime affixed. Now, say, this will reach say phase 10 after 8 days, there may be more patches available to install after 8 days than on 5/23. In this case, if we don't use snapshottime in all assessment queries, we end up listing more patches as available to the customer than what we install. This is raise CRIs on not all patches are being installed.

rane-rajasi:
> @rane-rajasi can the comment threads be updated with past discussions outside the PR? It's good to get to a list of:
> 
> * outstanding questions
> * already addressed items (reviewable)
> * pending work based on resolved conversations
> 
> And a consolidated plan to finish from there.

Ack. The major outstanding asks in here are:

1. Distinguishing between AzLinux 2.0 and 3.0
2. Distinguish between azlinux 3.0 supported version vs not supported versions
3. If AzLinux version is supported and tdnf version used in an old one (without snapshot), upgrade tdnf to support snapshot
4. Review Canonical strict SDP to delineate which code is strictly required to run in which scenario and isolate accordingly

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -396,4 +401,15 @@ def utc_to_standard_datetime(utc_datetime):
         def standard_datetime_to_utc(std_datetime):
             """ Converts datetime object to string of format '"%Y-%m-%dT%H:%M:%SZ"' """
             return std_datetime.strftime("%Y-%m-%dT%H:%M:%SZ")
+
+        @staticmethod
+        def datetime_string_to_posix_time(datetime_string, format_string):
+            """ Converts string of given format to posix datetime string.
+                type: (str, str) -> str"""
```
kjohn-msft, line 408: Excluding calling context as this is in a base layer - conformant posix time is int, not str

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -242,6 +242,11 @@ def cpu_arch():  # architecture
         @staticmethod
         def vm_name():     # machine name
             return platform.node()
+
+        @staticmethod
+        def version():
+            return str() if (EnvLayer.get_python_major_version() == 2) else distro.os_release_attr('version')
```
kjohn-msft, line 248: We carry our own copy of distro - why is this necessary?

Also, this needs to be more reliable --
return (distro.os_release_attr('version') or
        distro.os_release_attr('version_id') or
        distro.version() or
        Constants.Version.Unknown)   // use correct constant equivalent - this is just illustrative

Aside from that, a general comment at envlayer - the code must be deterministic for a caller. Hidden 2 vs 3 behavior needs to always be thought through more.

src/core/src/core_logic/ExecutionConfig.py
```diff
@@ -107,23 +107,13 @@ def __get_max_patch_publish_date(self, health_store_id):
...
-        """ Returns date in format: %Y%m%d i.e 20240401 in the above example """
         if health_store_id is not None and health_store_id != "":
             date_candidate = str(health_store_id)[-10:].replace(".", "")  # last 10 characters and remove '.'
             if len(date_candidate) == 8 and date_candidate.isdigit() and str(health_store_id)[-11:-10] == "_":
-                return date_candidate
-        return ""
+                max_patch_publish_date = date_candidate + "T000000Z"
+
+        self.composite_logger.log_debug("[EC] Getting max patch publish date. [MaxPatchPublishDate={0}][HealthStoreId={1}]".format(str(max_patch_publish_date), str(health_store_id)))
+        return max_patch_publish_date
```
kjohn-msft, line 116: Why is this changed?

src/core/tests/Test_CoreMain.py
```diff
@@ -668,16 +668,16 @@ def test_install_all_packages_for_azure_linux_autopatching(self):
...
         self.assertEqual(json.loads(substatus_file_data[1]["formattedMessage"]["message"])["patches"][0]["name"], "azurelinux-release.noarch")
-        self.assertTrue("Other" in str(json.loads(substatus_file_data[1]["formattedMessage"]["message"])["patches"][0]["classifications"]))
+        self.assertTrue("Security" in str(json.loads(substatus_file_data[1]["formattedMessage"]["message"])["patches"][0]["classifications"]))
         self.assertTrue("Installed" == json.loads(substatus_file_data[1]["formattedMessage"]["message"])["patches"][0]["patchInstallationState"])
         self.assertEqual(json.loads(substatus_file_data[1]["formattedMessage"]["message"])["patches"][3]["name"], "python3.x86_64")
-        self.assertTrue("Other" in str(json.loads(substatus_file_data[1]["formattedMessage"]["message"])["patches"][3]["classifications"]))
+        self.assertTrue("Security" in str(json.loads(substatus_file_data[1]["formattedMessage"]["message"])["patches"][3]["classifications"]))
         self.assertTrue("Installed" == json.loads(substatus_file_data[1]["formattedMessage"]["message"])["patches"][3]["patchInstallationState"])
```
kjohn-msft, line 681: Context to this?

src/core/src/core_logic/PatchInstaller.py
```diff
@@ -79,6 +79,7 @@ def start_installation(self, simulate=False):
 
         if self.execution_config.max_patch_publish_date != str():
             self.package_manager.set_max_patch_publish_date(self.execution_config.max_patch_publish_date)
+            # todo: validate if this meets the Az Linux and tdnf requirements
```
kjohn-msft, line 82: Marking todo comment spot for PR evolution.

src/core/src/package_managers/TdnfPackageManager.py
```diff
@@ -80,7 +76,8 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
         self.STR_TOTAL_DOWNLOAD_SIZE = "Total download size: "
         self.version_comparator = VersionComparator()
 
-        # if an Auto Patching request comes in on a Azure Linux machine with Security and/or Critical classifications selected, we need to install all patches, since classifications aren't available in Azure Linux repository
+        # Todo: Q: Do we need this now that get_security_updates() returns all updates?
```
kjohn-msft, line 79: Marking todo comment spot for PR evolution.

kjohn-msft: Behavior must be identical to CentOS - only default patching calls can get autoupgraded to Crit+Sec+Other. And sec updates returns everything. For any other type of call, the customer must explicitly specify everything or the request should fail.

src/core/src/package_managers/TdnfPackageManager.py
```diff
@@ -155,20 +130,20 @@ def get_all_updates(self, cached=False):
...
         self.all_updates_cached, self.all_update_versions_cached = self.extract_packages_and_versions(out)
-
         self.composite_logger.log_debug("[TDNF] Get all updates : [Cached={0}][PackagesCount={1}]]".format(str(False), len(self.all_updates_cached)))
         return self.all_updates_cached, self.all_update_versions_cached
 
     def get_security_updates(self):
         """Get missing security updates. NOTE: Classification based categorization of patches is not available in Azure Linux as of now"""
-        self.composite_logger.log_verbose("[TDNF] Discovering 'security' packages...")
-        security_packages, security_package_versions = [], []
+        self.composite_logger.log_verbose("[TDNF] Discovering all packages as 'security' packages, since Azure Linux does not support package classification...")
```
kjohn-msft, line 140: TDNF does not support security classifications.
It's good to not be overly Azure Linux coupled in a generic Tdnf class from the get go even if the first consumer is Azure Linux.

kjohn-msft: I'd actually take a closer look at the YumPackageManager implementation for a blueprint on how to think about this.

src/core/src/core_logic/ExecutionConfig.py
rane-rajasi: Nothing's changed here, don't know why this is showing up. Ref to code in master: https://github.com/Azure/LinuxPatchExtension/blob/4e31176f7d1c666508d1c6ab27164ea1bd2b9382/src/core/src/core_logic/ExecutionConfig.py#L106

src/core/tests/Test_CoreMain.py
rane-rajasi: This is an Az Linux test, all patches were earlier being marked as Other, which we have now changed to be marked as 'Security'

src/core/src/bootstrap/EnvLayer.py
rane-rajasi: This was added to the PR in error, I was trying something specific to Azure Linux back then

src/core/src/bootstrap/EnvLayer.py
rane-rajasi: Done

src/core/src/bootstrap/EnvLayer.py
rane-rajasi: Removed

src/core/src/core_logic/PatchInstaller.py
rane-rajasi: Removed

src/core/src/package_managers/TdnfPackageManager.py
rane-rajasi: Modifed

src/core/src/package_managers/TdnfPackageManager.py
rane-rajasi: Done

src/core/src/package_managers/TdnfPackageManager.py
```diff
@@ -75,7 +76,7 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
         self.STR_TOTAL_DOWNLOAD_SIZE = "Total download size: "
         self.version_comparator = VersionComparator()
 
-        # if an Auto Patching request comes in on a Azure Linux machine with Security and/or Critical classifications selected, we need to install all patches, since classifications aren't available in Azure Linux repository
+        # if an Auto Patching request comes in on an Azure Linux machine with Security and/or Critical classifications selected, we need to install all patches, since classifications aren't available in Azure Linux repository
```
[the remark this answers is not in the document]
rane-rajasi: Is should be added in our public doc: https://learn.microsoft.com/en-us/azure/virtual-machines/automatic-vm-guest-patching.
We had something similar for CentOS before EOL. @kjohn-msft to bring to your notice and also correct me if I'm wrong

src/core/src/package_managers/TdnfPackageManager.py
```diff
@@ -117,32 +128,38 @@ def get_all_updates(self, cached=False):
...
+        """Get missing security updates. NOTE: Classification based categorization of patches is not available in TDNF as of now"""
+        self.composite_logger.log_verbose("[TDNF] Discovering all packages as 'security' packages, since TDNF does not support package classification...")
+        security_packages, security_package_versions = self.get_all_updates(cached=False)
         self.composite_logger.log_debug("[TDNF] Discovered 'security' packages. [Count={0}]".format(len(security_packages)))
         return security_packages, security_package_versions
 
     def get_other_updates(self):
-        """Get missing other updates.
-        NOTE: This function will return all available packages since Azure Linux does not support package classification in it's repository"""
+        """Get missing other updates."""
```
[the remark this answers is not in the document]
rane-rajasi: @nikhim-um see the comment here for reference on the desired approach: 

- https://github.com/Azure/LinuxPatchExtension/pull/318#discussion_r2198886277
-

src/core/src/bootstrap/Constants.py
```diff
@@ -205,6 +205,8 @@ class StatusTruncationConfig(EnumBackport):
     YUM = 'yum'
     ZYPPER = 'zypper'
 
+    TDNF_MINIMUM_VERSION_FOR_STRICT_SDP = "3.5.8-3.azl3" # minimum version of tdnf required to support Strict SDP
```
kjohn-msft, line 208: This is the minimum version of the Azure Linux specific build of the TDNF package (.azl3 suffix in the version)  

1. Is this better positioned as a constant within the only class the uses it?

2. The implementation of TdnfPackageManager here, in general is an AzLinux specific flavor, and it may be worthwhile to qualify it as such so that we don't end up with a smattering of variables and classes that are ambiguous about their applicability when expanding out of that scope. (Along the lines of AzL3TdnfPackageManager or TdnfPackageManager being the parent class with Azl3TdnfPackageManager as the derivative with clear/distinct AzL3 specializations on top.
- For all consumers of this code (right now), nothing changes except the class from which instantiation happens (bulk name change only).
- The parent child relationship or other doesn't need to be in this PR but can be set on that path right away.

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -46,6 +47,15 @@ def __init__(self):
...
         return any(x in distro_name for x in Constants.AZURE_LINUX)
 
+    @staticmethod
+    def is_distro_azure_linux_3_or_beyond():
+        # type: () -> bool
+        """ Checks if the current distro is Azure Linux 3 """
+
+        version = distro.os_release_attr('version')
+        major = version.split('.')[0] if version else None
+        return major is not None and int(major) >= 3
```
kjohn-msft, line 57: Function body contents only do a part of what the function name promises outwards?

Basically the major version of whatever OS the system may be running is coming back as >=3. This could be anything right, outside the specific way it's used (which is incidental)?

And if the code doesn't work in some case on a matching AzLinux system, it says' No'?

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -396,4 +407,18 @@ def utc_to_standard_datetime(utc_datetime):
...
+            """ Converts string of given format to posix datetime string.
+                type: (str, str) -> int"""
+            # eg: Input: datetime_string: 20241220T000000Z (str), format_string: '%Y%m%dT%H%M%SZ' -> Output: 1734681600 (str)
+
+            # Parse the datetime string
+            dt = datetime.datetime.strptime(datetime_string, format_string)
+
+            # Convert to POSIX time assuming UTC
+            epoch = datetime.datetime(1970, 1, 1)
+            return int((dt - epoch).total_seconds())
```
kjohn-msft, line 422: Are inputs guaranteed safe? Or it is expected to throw on bad input?

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -396,4 +407,18 @@ def utc_to_standard_datetime(utc_datetime):
         def standard_datetime_to_utc(std_datetime):
             """ Converts datetime object to string of format '"%Y-%m-%dT%H:%M:%SZ"' """
             return std_datetime.strftime("%Y-%m-%dT%H:%M:%SZ")
+
+        @staticmethod
+        def datetime_string_to_posix_time(datetime_string, format_string):
+            """ Converts string of given format to posix datetime string.
+                type: (str, str) -> int"""
```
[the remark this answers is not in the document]
kjohn-msft: `# type: content`

Github tries to autoformat that

src/core/src/package_managers/TdnfPackageManager.py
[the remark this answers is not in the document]
kjohn-msft: image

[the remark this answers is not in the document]
kjohn-msft: Note that CentOS only has it for _platform-driven_ updates.

Customers doing it themselves need to be explicit in their choices (has been true for a very long time - AUM V1+) -- 
image

src/core/src/package_managers/TdnfPackageManager.py
```diff
@@ -90,6 +91,16 @@ def refresh_repo(self):
         self.invoke_package_manager(self.cmd_clean_cache)
         self.invoke_package_manager(self.cmd_repo_refresh)
 
+    # region Strict SDP using SnapshotTime
+    @staticmethod
+    def __generate_command(command_template, snapshotposixtime=str()):
```
kjohn-msft, line 96: Ambiguous naming - __generate_command_with_snapshotposixtime_if_specified ?

src/core/src/package_managers/TdnfPackageManager.py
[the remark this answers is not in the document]
kjohn-msft: Please reference this comment chain - expectation is full CentOS alignment. https://github.com/Azure/LinuxPatchExtension/pull/318/files#r2241041968

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -396,4 +407,18 @@ def utc_to_standard_datetime(utc_datetime):
...
+            # eg: Input: datetime_string: 20241220T000000Z (str), format_string: '%Y%m%dT%H%M%SZ' -> Output: 1734681600 (str)
+
+            # Parse the datetime string
+            dt = datetime.datetime.strptime(datetime_string, format_string)
+
+            # Convert to POSIX time assuming UTC
```
michellemcdaniel, line 420: Is there no built-in way to get posix time in python where we don't have to manually specific epoch?

src/core/src/package_managers/TdnfPackageManager.py
```diff
@@ -117,32 +128,38 @@ def get_all_updates(self, cached=False):
...
         all_packages, all_package_versions = self.get_all_updates(True)
+        security_packages, security_package_versions = self.get_security_updates()
+        for index, package in enumerate(all_packages):
+            if package not in security_packages:
+                other_packages.append(package)
+                other_package_versions.append(all_package_versions[index])
 
         self.composite_logger.log_debug("[TDNF] Discovered 'other' packages. [Count={0}]".format(len(other_packages)))
-        return all_packages, all_package_versions
+        return other_packages, other_package_versions
```
kjohn-msft, line 156: This code does nothing - can just return [], []

src/core/src/package_managers/YumPackageManager.py
```diff
@@ -266,6 +266,9 @@ def install_updates_fail_safe(self, excluded_packages):
 
     def install_security_updates_azgps_coordinated(self):
         pass
+
+    def meets_azgps_coordinated_requirements(self):
+        return
```
kjohn-msft, line 271: This will fail when true type definitions are added to this. Must return a bool and in this case, False.

src/core/src/package_managers/TdnfPackageManager.py
```diff
@@ -215,7 +232,70 @@ def install_updates_fail_safe(self, excluded_packages):
...
+
+    def get_tdnf_version(self):
+        """Get the version of TDNF installed on the system"""
+        self.composite_logger.log_debug("[TDNF] Getting tdnf version...")
+        cmd = "rpm -q tdnf | sed -E 's/tdnf-([0-9]+\\.[0-9]+\\.[0-9]+-[0-9]+).*/\\1/'"
+        code, output = self.env_layer.run_command_output(cmd, False, False)
+        if code == 0:
+            # Sample output: 3.5.8-3
+            version = output.split()[0] if output else None
+            self.composite_logger.log_debug("[TDNF] Installed TDNF version [Version={0}]".format(version))
```
kjohn-msft, line 282: Ambiguous in isolation - this suggests you installed that TDNF version.

'TDNF version detected'  for e.g. is more obvious

src/core/src/package_managers/TdnfPackageManager.py
```diff
@@ -215,7 +232,70 @@ def install_updates_fail_safe(self, excluded_packages):
...
+            return False
+        elif not self.version_comparator.compare_versions(tdnf_version, minimum_tdnf_version_for_strict_sdp) >= 0:
+            self.composite_logger.log_warning("[TDNF] TDNF version installed is less than the minimum required version. [InstalledVersion={0}][MinimumRequiredVersion={1}]".format(tdnf_version, Constants.TDNF_MINIMUM_VERSION_FOR_STRICT_SDP))
+            return False
+        return True
+
+    def get_tdnf_version(self):
+        """Get the version of TDNF installed on the system"""
+        self.composite_logger.log_debug("[TDNF] Getting tdnf version...")
+        cmd = "rpm -q tdnf | sed -E 's/tdnf-([0-9]+\\.[0-9]+\\.[0-9]+-[0-9]+).*/\\1/'"
```
kjohn-msft, line 277: You are dropping the distro-specific suffix (.azl3), which is actually critical to this code path, unless Microsoft's changes are accepted upstream (in whatever version).

src/core/src/package_managers/TdnfPackageManager.py
```diff
@@ -215,7 +232,70 @@ def install_updates_fail_safe(self, excluded_packages):
...
+        code, output = self.env_layer.run_command_output(cmd, False, False)
+        if code == 0:
+            # Sample output: 3.5.8-3
+            version = output.split()[0] if output else None
```
kjohn-msft, line 281: Don't discard distro specific suffix in check - include it explicitly. And consequently adjust tests.

src/core/src/package_managers/TdnfPackageManager.py
```diff
@@ -215,7 +232,70 @@ def install_updates_fail_safe(self, excluded_packages):
...
+                    error_msg = "Failed to meet minimum TDNF version requirement for strict safe deployment. Defaulting to regular upgrades."
+                    self.composite_logger.log_error(error_msg + "[Error={0}]".format(repr(error_msg)))
+                    self.status_handler.add_error_to_status(error_msg)
+                    self.set_max_patch_publish_date()  # fall-back
+                    return False
+                return True
+
+    def is_minimum_tdnf_version_for_strict_sdp_installed(self):
+        """Check if  at least the minimum required version of TDNF is installed"""
+        self.composite_logger.log_debug("[TDNF] Checking if minimum TDNF version required for strict safe deployment is installed...")
```
kjohn-msft, line 263: Please share log file from executions - there seem to be many redundant log lines hidden by nested function calls. While each function looks ok in isolation, the incremental downward nesting creates verbosity. Take 263 + 277..

Now, is the verbosity necessary too - in more than one way. What lines warrant a debug vs. what is okay as verbose? I'd lean towards 263 being verbose as an example. It's useful in a per-VM scenario (which we should defer to physical logs more), not so much at scale.

src/core/src/package_managers/TdnfPackageManager.py
```diff
@@ -215,7 +232,70 @@ def install_updates_fail_safe(self, excluded_packages):
...
+        tdnf_version = self.get_tdnf_version()
+        minimum_tdnf_version_for_strict_sdp = re.match(r"(\d+\.\d+\.\d+-\d+)", Constants.TDNF_MINIMUM_VERSION_FOR_STRICT_SDP).group(1)
+        if tdnf_version is None:
+            self.composite_logger.log_error("[TDNF] Failed to get TDNF version. Cannot proceed with strict safe deployment. Defaulting to regular upgrades.")
+            return False
+        elif not self.version_comparator.compare_versions(tdnf_version, minimum_tdnf_version_for_strict_sdp) >= 0:
+            self.composite_logger.log_warning("[TDNF] TDNF version installed is less than the minimum required version. [InstalledVersion={0}][MinimumRequiredVersion={1}]".format(tdnf_version, Constants.TDNF_MINIMUM_VERSION_FOR_STRICT_SDP))
```
kjohn-msft, line 270: Less ambiguity needed in isolation... minimum required version for 'what' (from the customer perspective)

src/core/src/package_managers/TdnfPackageManager.py
```diff
@@ -215,7 +232,70 @@ def install_updates_fail_safe(self, excluded_packages):
...
+        """ Check if the system meets the requirements for Azure Linux strict safe deployment and attempt to update TDNF if necessary """
+        self.composite_logger.log_debug("[TDNF] Checking if system meets Azure Linux security updates requirements...")
+        # Check if the system is Azure Linux 3.0 or beyond
+        if not self.env_layer.is_distro_azure_linux_3_or_beyond():
+            self.composite_logger.log_error("[TDNF] The system does not meet minimum Azure Linux requirement of 3.0 or above for strict safe deployment. Defaulting to regular upgrades.")
+            self.set_max_patch_publish_date()  # fall-back
+            return False
+        else:
+            if self.is_minimum_tdnf_version_for_strict_sdp_installed():
+                self.composite_logger.log_debug("[TDNF] Minimum tdnf version for strict safe deployment is installed.")
```
kjohn-msft, line 250: The code structuring when viewed from the lens of how, when and what logs get emitted can use a bit more work.
I'm noticing this a static reading, but it'll be a lot more obvious when seen in output. For a higher quality review, , it's good to share 'true logs' as reference. (or run through a log adjustment iteration, and then share logs from the code in that iteration)

src/core/src/package_managers/TdnfPackageManager.py
```diff
@@ -215,7 +232,70 @@ def install_updates_fail_safe(self, excluded_packages):
         return
 
     def install_security_updates_azgps_coordinated(self):
-        pass
+        """Install security updates in Azure Linux following strict SDP"""
+        command = self.__generate_command(self.install_security_updates_azgps_coordinated_cmd, self.max_patch_publish_date)
+        out, code = self.invoke_package_manager_advanced(command, raise_on_exception=False)
+        return code, out
+
+    def meets_azgps_coordinated_requirements(self):
```
kjohn-msft, line 240: A function named 'meets_azgps_coordinated_requirements'  (implicit observational check)
doing this 'attempt_tdnf_update_to_meet_strict_sdp_requirements()' inside (state change) is incorrectly representing itself outwards.

src/core/src/package_managers/TdnfPackageManager.py
```diff
@@ -117,32 +128,38 @@ def get_all_updates(self, cached=False):
...
         self.composite_logger.log_debug("[TDNF] Discovered 'other' packages. [Count={0}]".format(len(other_packages)))
-        return all_packages, all_package_versions
+        return other_packages, other_package_versions
 
     def set_max_patch_publish_date(self, max_patch_publish_date=str()):
-        pass
+        """Set the max patch publish date in POSIX time for strict SDP"""
+        self.composite_logger.log_debug("[TDNF] Setting max patch publish date. [MaxPatchPublishDate={0}]".format(str(max_patch_publish_date)))
+        self.max_patch_publish_date = str(self.env_layer.datetime.datetime_string_to_posix_time(max_patch_publish_date, '%Y%m%dT%H%M%SZ')) if max_patch_publish_date != str() else max_patch_publish_date
+        self.composite_logger.log_debug("[TDNF] Set max patch publish date. [MaxPatchPublishDatePosixTime={0}]".format(str(self.max_patch_publish_date)))
```
kjohn-msft, line 162: Too many debug logs - be frugal with these and consolidate and triage what needs to be debug and what needs to be verbose. Comment applies everywhere.

src/core/src/package_managers/TdnfPackageManager.py
```diff
@@ -117,32 +128,38 @@ def get_all_updates(self, cached=False):
...
 
     def get_security_updates(self):
-        """Get missing security updates. NOTE: Classification based categorization of patches is not available in Azure Linux as of now"""
-        self.composite_logger.log_verbose("[TDNF] Discovering 'security' packages...")
-        security_packages, security_package_versions = [], []
+        """Get missing security updates. NOTE: Classification based categorization of patches is not available in TDNF as of now"""
+        self.composite_logger.log_verbose("[TDNF] Discovering all packages as 'security' packages, since TDNF does not support package classification...")
+        security_packages, security_package_versions = self.get_all_updates(cached=False)
         self.composite_logger.log_debug("[TDNF] Discovered 'security' packages. [Count={0}]".format(len(security_packages)))
         return security_packages, security_package_versions
```
kjohn-msft, line 141: This is my mistake from earlier high-level discussions - the correct CentOS alignment is this -- https://github.com/Azure/LinuxPatchExtension/pull/318/files#r2241037265

The effective outcome is sort of the same but not quite. Make note of both scenarios - default patching, and customer-initiated and how they differ.

src/core/src/package_managers/TdnfPackageManager.py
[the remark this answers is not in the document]
rane-rajasi: Agreed and this behavior remains the same for Azure Linux

src/core/src/package_managers/TdnfPackageManager.py
rane-rajasi: All in VM test logs files are attached in the first comment of this PR: https://github.com/Azure/LinuxPatchExtension/pull/318#issue-3031705551

src/core/src/package_managers/TdnfPackageManager.py
rane-rajasi: All in VM test logs files are attached in the first comment of this PR: https://github.com/Azure/LinuxPatchExtension/pull/318#issue-3031705551

Will also update logs based on changes from this iteration, but for the purposes of this review, existing logs should work. Also, a note, all in VM test logs will always be linked in that first comment

src/core/src/package_managers/TdnfPackageManager.py
rane-rajasi: The linked CentOS alignment does exist for default patching within TDNF as well. IIRC, our past discussion and conclusion to fetch from get_all_updates() within security, was to ensure these patches are tagged as 'Security' in status, which is needed to mark them all as Security updates for the customer

src/core/src/bootstrap/Constants.py
```diff
@@ -347,7 +349,7 @@ class TelemetryTaskName(EnumBackport):
     TELEMETRY_NOT_COMPATIBLE_ERROR_MSG = "Unsupported older Azure Linux Agent version. To resolve: http://aka.ms/UpdateLinuxAgent"
     TELEMETRY_COMPATIBLE_MSG = "Minimum Azure Linux Agent version prerequisite met"
     PYTHON_NOT_COMPATIBLE_ERROR_MSG = "Unsupported older Python version. Minimum Python version required is 2.7. [DetectedPythonVersion={0}]"
-    INFO_STRICT_SDP_SUCCESS = "Success: Safely patched your VM in a AzGPS-coordinated global rollout. https://aka.ms/AzGPS/StrictSDP [Target={0}]"
+    INFO_STRICT_SDP_SUCCESS = "Success: Safely patched your VM in a AzGPS-coordinated global rollout. https://aka.ms/AzGPS/StrictSDP [Target={0}]" # TBD: the link is specific to Canonical, we need a generic one or one link of each distro
```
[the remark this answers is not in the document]
rane-rajasi: Done

src/core/src/bootstrap/EnvLayer.py
rane-rajasi: Added an Azure Linux check before verifying the major version

src/core/src/bootstrap/EnvLayer.py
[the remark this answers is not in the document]
rane-rajasi: Done in all new functions added in this PR

src/core/src/package_managers/TdnfPackageManager.py
rane-rajasi: Done

src/core/src/package_managers/YumPackageManager.py
rane-rajasi: Done wherever applicable

src/core/src/package_managers/TdnfPackageManager.py
```diff
@@ -325,8 +405,9 @@ def get_dependent_list(self, packages):
                 package_names += ' '
             package_names += package
 
-        self.composite_logger.log_verbose("[TDNF] Resolving dependencies. [Command={0}]".format(str(self.single_package_upgrade_simulation_cmd + package_names)))
-        output = self.invoke_package_manager(self.single_package_upgrade_simulation_cmd + package_names)
+        cmd = self.single_package_upgrade_simulation_cmd + package_names
+        self.composite_logger.log_verbose("[TDNF] Resolving dependencies. [Command={0}]".format(str(cmd)))
```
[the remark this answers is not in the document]
rane-rajasi: Done

src/core/src/package_managers/TdnfPackageManager.py
```diff
@@ -215,7 +232,70 @@ def install_updates_fail_safe(self, excluded_packages):
...
+        if code == 0:
+            # Sample output: 3.5.8-3
+            version = output.split()[0] if output else None
+            self.composite_logger.log_debug("[TDNF] Installed TDNF version [Version={0}]".format(version))
+            return version
+        else:
+            self.composite_logger.log_error("[TDNF] Failed to get TDNF version. [Command={0}][Code={1}][Output={2}]".format(cmd, code, output))
+            return None
+
+    def attempt_tdnf_update_to_meet_strict_sdp_requirements(self):
```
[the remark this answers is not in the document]
rane-rajasi: Done

src/core/src/package_managers/TdnfPackageManager.py
rane-rajasi: Added 'for strict SDP' to these

src/core/src/package_managers/TdnfPackageManager.py
rane-rajasi: In this particular case, datetime_string_to_posix_time() can potentially throw an exception, hence I feel we need to add both logs as debug, 

Reviewing in other places...

kjohn-msft (changes requested):
Two classes of comments 
- one must address (architecturally) but not in this PR - has to be the immediately following PR though. This is to progressively get the code in.
- one set of easily addressable comments that can go in immediately

No major flags in iteration 15.

src/core/tests/Test_CoreMain.py
```diff
@@ -27,6 +27,7 @@
 from core.tests.library.ArgumentComposer import ArgumentComposer
 from core.tests.library.LegacyEnvLayerExtensions import LegacyEnvLayerExtensions
 from core.tests.library.RuntimeCompositor import RuntimeCompositor
+from core.src.external_dependencies import distro
```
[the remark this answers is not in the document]
rane-rajasi: Done

src/core/src/package_managers/TdnfPackageManager.py
```diff
@@ -748,3 +829,9 @@ def separate_out_esm_packages(self, packages, package_versions):
     def get_package_install_expected_avg_time_in_seconds(self):
         return self.package_install_expected_avg_time_in_seconds
 
+    # region - AzLinux specilizations
```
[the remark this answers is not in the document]
rane-rajasi: Done

src/core/src/package_managers/TdnfPackageManager.py
```diff
@@ -90,6 +94,16 @@ def refresh_repo(self):
         self.invoke_package_manager(self.cmd_clean_cache)
         self.invoke_package_manager(self.cmd_repo_refresh)
 
+    # region Strict SDP using SnapshotTime
+    @staticmethod
+    def __generate_command_with_snapshotposixtime_if_specified(command_template, snapshotposixtime=str()):
```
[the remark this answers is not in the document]
rane-rajasi: Done

src/core/src/package_managers/TdnfPackageManager.py
rane-rajasi: Done

src/core/src/package_managers/TdnfPackageManager.py
```diff
@@ -215,7 +225,78 @@ def install_updates_fail_safe(self, excluded_packages):
...
+                    self.set_max_patch_publish_date()  # fall-back
+                    return False
+                return True
+
+    def is_minimum_tdnf_version_for_strict_sdp_installed(self):
+        # type: () -> bool
+        """Check if  at least the minimum required version of TDNF is installed"""
+        self.composite_logger.log_debug("[TDNF] Checking if minimum TDNF version required for strict safe deployment is installed...")
+        tdnf_version = self.get_tdnf_version()
+        minimum_tdnf_version_for_strict_sdp = self.azl3_tdnf_packagemanager.TDNF_MINIMUM_VERSION_FOR_STRICT_SDP
```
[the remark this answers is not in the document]
rane-rajasi: PBI: 34535982

PR #336, opened 2026-03-02
Bugfix: Fix yum output parsing to avoid false reboot detection

sadiksubhani9-sudo:
The leading number (`62`) was incorrectly interpreted as a process ID, which
resulted in false positives during reboot-required checks and incorrect update
status reporting in Azure Update Manager.

### References
- Incident 447902558  
- ADO Bug: [AzGPS][Linux][Guest][Arc][AWS][CRI] Yum incorrectly parsing data for reboot check in AWS  
- Related Arc fix: PR 8873077 (format validation for reboot check)

---

### Impact
- Prevents false reboot detection caused by yum informational output
- Improves accuracy of update status and counters reported by AUM
- No functional change for valid yum process parsing

### DIFF
--- a/src/core/src/package_managers/YumPackageManager.py
+++ b/src/core/src/package_managers/YumPackageManager.py
@@ -1062,7 +1062,7 @@ class YumPackageManager(PackageManager):
                     continue
 
             process_details = re.split(r'\s+', line.strip())
-            if len(process_details) < 7:
+            if len(process_details) < 7 or line.find("packages excluded") >= 0:
                 self.composite_logger.log_verbose("[YPM] > Inapplicable line: " + str(line))
                 continue
             else:


src/core/src/package_managers/YumPackageManager.py
```diff
@@ -1062,7 +1062,7 @@ def do_processes_require_restart(self):
                     continue
 
             process_details = re.split(r'\s+', line.strip())
-            if len(process_details) < 7:
+            if len(process_details) < 7 or line.find("packages excluded") >= 0:
```
rane-rajasi, line 1065: Please share an example of why this change is needed. In the sense, what was the current behavior seen on a VM, with the help of logs and how does this change correct it, again with the help of in VM logs

rane-rajasi (changes requested):

src/core/src/package_managers/YumPackageManager.py
sadiksubhani9-sudo: Before:
The customer saw wrong update status and higher machine counts because some AWS Linux machines printed yum messages that were misread as reboot‑required. This caused machines to show Pending reboot and made the counters add up to more than the actual number of servers.
After:
Those yum informational lines are now ignored and no longer treated as reboot signals.
This fixes the false Pending reboot status and brings the machine counters back to the correct total.

rane-rajasi: I'm asking for logs from a VM for the before and after scenarios. Create a VM, reproduce the before scenario which show the false reboot detection and then another set of logs that show this issue fixed with the help of your code change

sadiksubhani9-sudo: Before (VM logs):
The yum output contained informational lines such as  
`62 packages excluded due to repository priority protections`  
which were parsed as process entries. In the extension logs this caused the reboot‑required process count to increment, resulting in the VM being marked as Pending reboot even though no reboot was actually needed.

After (VM logs):  
With the change, the same line is logged as  
`[YPM] > Inapplicable line: 62 packages excluded due to repository priority protections`  
and is skipped during parsing. The reboot‑required process count is no longer incremented, and the VM correctly reports No pending reboot.

rane-rajasi: Your response is just a statement on what the before and after **should** look like, this is **not** however an actual log fetched from a VM. Repeating my ask from before:  
image

@sadiksubhani9-sudo, you need to create a VM, run this code in that VM, fetch logs from the VM that demonstrates the before scenario. Then run your changed code in the VM, fetch logs that demonstrate the after scenario

sadiksubhani9-sudo: Before : https://github.com/user-attachments/assets/24c618de-e446-4942-85b5-785c917bd18e
After:  https://github.com/user-attachments/assets/d59fed9c-49f1-4bea-b362-8ba7268d1c62

rane-rajasi: Which VM was this run on? Which patch operation was invoked? Please attach the complete set of extension logs from this run:

image
https://github.com/Azure/LinuxPatchExtension?tab=readme-ov-file#3-troubleshooting

sadiksubhani9-sudo: https://microsoft-my.sharepoint.com/:w:/p/skarumanchi/IQABVL8mAG34Q4vm6F6J_J2TAUEl-Y-jnXO7gAY4WicoyTo?e=xR4KAD

sadiksubhani9-sudo: @rane-rajasi sending a reminder for the PR review...

sadiksubhani9-sudo: @kjohn-msft, Please take a look when you get a chance. Thanks!

kjohn-msft: @rane-rajasi requested the contents of a specific folder (ideally zipped) from the test machine - prior to the change and after the change. I don't see that in the internal folder. The goal is to understand the dev-test validation flow associated with any change (this change implicitly being simple). Which means:
- demonstrating prior behavior and new behavior with _full execution log files_ from a representative image.
- having corresponding unit tests based on those samples.

rane-rajasi: @sadiksubhani9-sudo please share the contents of the extension logs folder, path shared here:

image

Koshy is referring the same in this feedback

sadiksubhani9-sudo: [after-fix-logs.tar.gz](https://github.com/user-attachments/files/26993839/after-fix-logs.tar.gz)
[before-fix-logs.tar.gz](https://github.com/user-attachments/files/26993840/before-fix-logs.tar.gz)
 
https://github.com/user-attachments/assets/f670dccc-2022-4104-a52c-d760e1800be0

Please take a look - @rane-rajasi 
cc- @kjohn-msft

Issue #337, opened 2026-03-23
Azure Update Manager disregards FIPS repositories for Ubuntu 22.04


Rkoffer-SMX:
**Describe the bug**
Due to the filter in AptitudePackageManager, updates provided by the FIPS repo https://esm.ubuntu.com/fips-updates/ubuntu jammy-updates does not pass the filter and is dropped.  If the Ubuntu Pro client is enabled, this problem is worse and will ultimately lead to non-FIPS packages being installed causing conflicts.

Azure Update Manager installs non-FIPS openssh-server on FIPS-enabled Ubuntu 22.04
==================================================================================

Root Cause
----------
The LinuxPatchExtension (v1.6.64) has two independent mechanisms for discovering
available updates:

  1. apt-get -s dist-upgrade  (using a custom filtered source list)
  2. Ubuntu Pro Client API    (via uaclient.api.u.pro.packages.updates.v1)

When building the security-classification source list, the extension filters each
deb line with a simple substring check:

  AptitudePackageManager.py:186
    if base_classification == Constants.PackageClassification.SECURITY and "security" not in line:
        continue

The FIPS updates repo uses suite name "jammy-updates", not "jammy-security":

  deb https://esm.ubuntu.com/fips-updates/ubuntu jammy-updates main

So it gets excluded from the security source list. Other ESM repos survive because
they have "-security" suites (jammy-apps-security, jammy-infra-security).

During DISCOVERY, the Pro Client independently finds the FIPS openssh packages
(classified as "standard-security") and they get merged into the combined package
list. The log shows them as "Pro Client only updates."

During INSTALLATION, the Pro Client is not involved. The extension calls:

  AptitudePackageManager.py:486-490
    install_security_updates_azgps_coordinated() ->
      apt-get -y --only-upgrade true dist-upgrade <SOURCES>

This rebuilds the same filtered security source list (without the FIPS repo).
apt resolves openssh-server from jammy-security instead, installing the non-FIPS
version.

Summary: discovery uses Pro Client (FIPS-aware), installation uses apt with a
filtered source list (not FIPS-aware). The two paths are not coordinated.

Rkoffer-SMX:
[...]

PR #338, opened 2026-03-23
Fix: Include FIPS-updates repo in security source list (#337)

Rkoffer-SMX:
The security source list filter excluded FIPS-updates repos because they use suite names like "jammy-updates" instead of "jammy-security". This caused apt to resolve packages from jammy-security during installation, replacing FIPS-certified packages with non-FIPS versions.

Broadens the filter in both one-line and DEB882-style source list readers to also retain lines/stanzas containing "fips-updates".

### DIFF
--- a/src/core/src/package_managers/AptitudePackageManager.py
+++ b/src/core/src/package_managers/AptitudePackageManager.py
@@ -183,7 +183,7 @@ class AptitudePackageManager(PackageManager):
 
         for line in sources_content_lines:
             if len(line.strip()) != 0 and not line.strip().startswith("#"):
-                if base_classification == Constants.PackageClassification.SECURITY and "security" not in line:
+                if base_classification == Constants.PackageClassification.SECURITY and "security" not in line and "fips-updates" not in line:
                     continue
                 std_source_list_content += "\n" + self.__apply_max_patch_publish_date(sources_content=line, max_patch_publish_date=max_patch_published_date)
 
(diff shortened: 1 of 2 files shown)


Rkoffer-SMX:
@microsoft-github-policy-service agree company="SMX"

kjohn-msft (approved):

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -183,7 +183,7 @@ def __read_one_line_style_list_format(self, file_path, max_patch_published_date,
 
         for line in sources_content_lines:
             if len(line.strip()) != 0 and not line.strip().startswith("#"):
-                if base_classification == Constants.PackageClassification.SECURITY and "security" not in line:
+                if base_classification == Constants.PackageClassification.SECURITY and "security" not in line and "fips-updates" not in line:
```
yashnap, line 186: Do we know if there are any other repositories, similar to FIPS, that might not follow the expected naming convention for security updates?

rane-rajasi (approved):

src/core/src/package_managers/AptitudePackageManager.py
kjohn-msft: This is a one-off exception we're taking.

kjohn-msft (approved):

yashnap (approved):

kjohn-msft:
Due to tooling issues needing more investigation, this approved change was taken as in in a separate PR by an internal Microsoft employee. https://github.com/Azure/LinuxPatchExtension/pull/348

PR #129, opened 2022-03-21
Fix failing GitHub workflow tests

benank:
Disabled 4 tests that were failing 100% of the time on GitHub runners. Fixed temporary directory creation to use the runner's temporary directory. Also updated the workflow with newer versions for actions and split up large operations into multiple steps.

PR #340, opened 2026-04-21
Feature: Sanitize credential-like URLs in telemetry events to avoid False CredScan detection

yashnap:
### **Problem Statement :**  

An automated Credential Scan runtime telemetry scan raised an ICM against Azure Guest Patching Service indicating the presence of credential‑like values in extension status telemetry uploaded to Geneva.

CredScan periodically scans Microsoft‑owned telemetry storage and raises incidents when credential‑like patterns are detected. 

Analysis indicates that package manager stdout/stderr (e.g., apt, yum, dnf) emits customer‑configured repository authentication values in URI userinfo format (e.g., username:secret@host). When included in LPE extension payloads, these values get uploaded to Geneva telemetry, where CredScan runtime scanning subsequently classifies them as potential credential leaks. Certain format :password that seemed to have been deprecated but the validations surrounding the previous checks are still present flagging the values that are in similar format. 

While these values originate from customer‑owned VM configuration and are required for local package installation and diagnostics, their propagation into Microsoft‑owned telemetry systems results in recurring automated CredScan incidents which are false positives.

### 
**SOLUTION** 

This change introduces targeted sanitization at the telemetry export boundary:
- Added new class CredentailSanitizer.py which has the logic to filter out sensitive information.
- Applied sanitization before telemetry JSON writes in both telemetry writers (extension and core paths)
- Local logs are unchanged since these events are directly written to events folders which is read by CredScan.
- Added UT's to support the change.

### 
**TESTING STRATEGY**

- Created VM ( RHEL in my case) and verified the linux agent is running on it. ( systemctl status waagent)
- Created Zip from my local code changes and followed steps to run it on the VM
- Added more verbose logging in the yum update, install commands because it was redacting the URL's itself before emitting locally. (Temporary local setup)
- Simulated a failure using Wrong URL with credentials exposed that runs when the extension is enabled
- Verified in /events and /logs ( My handlerEnvironment.json had events located at /tmp/x/events and logs at tmp/x/log) that the log folder ihas the credentials still present vs events folder *.json have them redacted.
### 
**Logs Folder Output** 

logsFolder

### **Events Folder Output**

eventsFolder

**APPROACH**

- Created Sanitization class in both core and extension package and injected the Dependency in respective Telemetry class. 
- Good if we want to add more patterns in the futurem, no consumer impact and backward compatible.

- [X] I have tested out the changes on my local using both RHEL/Linux VM machine.
- [X] I have considered approach where logic can be stay in one place (i.e Utility class) but Utility approach is discouraged.
- [X] I have considered adding the sanitizer logic in individual classes instead of creating separate class for Sanitization but from extensibility perspective the current approach I have is better.


yashnap:
@microsoft-github-policy-service agree company="Microsoft"

rane-rajasi:
@yashnap code coverage checks are failing for this one. We will need those to succeed for a review process to begin. Fyi, we check code coverage in python 3 **and** python 2. Run the test suites for both of these in your local, you can use any minor versions within python 2 and 3

rane-rajasi (changes requested):
Left a comment here: https://github.com/Azure/LinuxPatchExtension/pull/340#issuecomment-4296855976

src/core/src/Utility.py
```diff
@@ -0,0 +1,52 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Requires Python 2.7+
+
+import re
+
+
+class Utility(object):
```
kjohn-msft: We moved away from a Utility class as it becomes a dumping ground for functions. Ask this question to the engineering agent: "What are the drawbacks of having a Utility class in code? From a best practices perspective."

Also review SOLID principles in programming, with S = Single-responsibility principle.

kjohn-msft (changes requested):
One comment inline. Separate: Always get @rane-rajasi's sign off first.

rane-rajasi:
https://github.com/Azure/LinuxPatchExtension/pull/340#issue-4305734406
@yashnap Remove the email and ADO item links from PR description. This is an open-source code base, no confidential MS sources should be shared. If you feel these are needed, add these references in your PR review request post, never share them in here

src/core/src/service_interfaces/CredentialSanitizer.py
```diff
@@ -0,0 +1,50 @@
+# Copyright 2020 Microsoft Corporation
```
rane-rajasi: nit: Year should be 2026

src/core/src/service_interfaces/CredentialSanitizer.py
```diff
@@ -0,0 +1,50 @@
...
+import re
+
+
+class CredentialSanitizer(object):
+    """Sanitizes credential-like values from URIs. Removes password/token from URI userinfo."""
+
+    @staticmethod
+    def sanitize(message):
+        """Sanitizes credential-like values from URIs.
+
```
rane-rajasi, line 21: nit: Remove extra line breaks and spacing. This could be replaced as:

"""Sanitizes credential-like values from URIs. Removes password/token from URI userinfo.
```
- Args:
  - message: The message to sanitize
- Returns: Message with credentials removed from URIs"""
```

No empty line breaks in between

rane-rajasi: The class description and function description is the same. 
Function → “does something”
Class → “represents something” 

Update their description accrodingly

src/core/src/service_interfaces/CredentialSanitizer.py
```diff
@@ -0,0 +1,50 @@
...
+            # Handles credentials containing special characters (except @, /, whitespace)
+            # Groups:
+            # (1) scheme: https://, http://, or ftp://
+            # (2) username: one or more non-whitespace, non-slash, non-colon, non-@ characters
+            # (3) password: zero or more non-whitespace, non-slash, non-@ characters
+            sanitized_message = re.sub(
+                r'(https?://|ftp://)([^:/@\s]+):([^@/\s]*)@',
+                r'\1\2@',
+                message
+            )
```
rane-rajasi: We have a 120 char limit for a single line, these could all go in a single line

i.e. sanitized_message = re.sub(r'(https?://|ftp://)([^:/@\s]+):([^@/\s]*)@', r'\1\2@', message)

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -23,21 +23,23 @@
...
+from core.src.service_interfaces.CredentialSanitizer import CredentialSanitizer
 
 
 class TelemetryWriter(object):
     """Class for writing telemetry data to data transports"""
 
     TELEMETRY_BUFFER_DELIMETER= "\n|\t"
 
-    def __init__(self, env_layer, composite_logger, events_folder_path, telemetry_supported):
+    def __init__(self, env_layer, composite_logger, events_folder_path, telemetry_supported, credential_sanitizer=None):
```
rane-rajasi: LPE uses a Bootstrapper-ConfigurationFactory setup for initializations. Follow the code path from CoreMain.py __init() lines ~26-30, to understand how bootstrapper works. All changes to classes will need to be made following those examples

rane-rajasi: Let me know if you need help understanding this setup

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -161,6 +168,7 @@ def __ensure_message_restriction_compliance(self, full_message):
             self.composite_logger.log_telemetry_module_error("Error occurred while formatting message for a telemetry event. [Error={0}]".format(repr(e)))
             raise
 
+
```
rane-rajasi: Remove extra line space.

Generic guide for all PRs: avoid making any changes to existing code that's not a part of your PR

src/core/tests/Test_TelemetryWriter.py
```diff
@@ -311,5 +314,121 @@ def test_write_event_with_buffer_true_and_empty_string_and_then_flush_with_non_e
             f.close()
             self.assertTrue(text_found.string.startswith("Message 1"))
 
+    # ==================== Unit Tests for Credential Sanitization ====================
+    def test_sanitize_credentials_from_uri_https(self):
+        """ Test sanitization of HTTPS URIs with credentials """
+        message = "Error connecting to https://testuser:TESTTOKEN123456@invalid.repo.example/rpm/repodata/repomd.xml"
+        sanitized = CredentialSanitizer.sanitize(message)
```
rane-rajasi: Refer RuntimeCompositor.py, ArgumentComposer.py on how initializations are setup for the test suite and follow those

src/core/tests/Test_TelemetryWriter.py
```diff
@@ -311,5 +314,121 @@ def test_write_event_with_buffer_true_and_empty_string_and_then_flush_with_non_e
...
+        self.assertEqual(sanitized, expected_message)
+
+    def test_sanitize_credentials_from_uri_http(self):
+        """ Test sanitization of HTTP URIs with credentials """
+        message = "Connection failed to http://user123:password123@example.com/path"
+        sanitized = CredentialSanitizer.sanitize(message)
+        # Password should be removed
+        self.assertNotIn("password123", sanitized)
+        # Username should be preserved
+        self.assertIn("user123@example.com", sanitized)
```
rane-rajasi: Use assertEqual with the sanitized and expected message

src/core/tests/Test_TelemetryWriter.py
```diff
@@ -311,5 +314,121 @@ def test_write_event_with_buffer_true_and_empty_string_and_then_flush_with_non_e
...
+    def test_sanitize_credentials_multiple_urls(self):
+        """ Test sanitization with multiple URLs containing credentials """
+        message = "Failed to fetch from https://user1:pass1@host1.com/api and http://user2:pass2@host2.com/data"
+        sanitized = CredentialSanitizer.sanitize(message)
+        # Passwords should be removed
+        self.assertNotIn("pass1", sanitized)
+        self.assertNotIn("pass2", sanitized)
+        # Usernames should be preserved
+        self.assertIn("user1@host1.com", sanitized)
+        self.assertIn("user2@host2.com", sanitized)
```
rane-rajasi: Try using assertEqual if you can, it gives more confidence/credibility to the test, since it checks the complete string/input

src/core/tests/Test_TelemetryWriter.py
```diff
@@ -311,5 +314,121 @@ def test_write_event_with_buffer_true_and_empty_string_and_then_flush_with_non_e
...
+
+    def test_sanitize_credentials_jfrog_repo_error(self):
```
rane-rajasi: The test name could be: test_sanitize_credentials_with_no_credentials_in_message() OR test_sanitize_credentials_with_no_secrets_in_input_str()

src/core/tests/Test_TelemetryWriter.py
```diff
@@ -311,5 +314,121 @@ def test_write_event_with_buffer_true_and_empty_string_and_then_flush_with_non_e
...
+        """  ERROR with 401 status code from jfrog.io """
+        message = "ERROR: Failed to download metadata for repo 'packages-microsoft-com-prod': Status code: 401 for https://cec-aa.jfrog.io/artifactory/glib-rpm-hel9-lts-microsoft-com/repodata/repomd.xml"
+        sanitized = CredentialSanitizer.sanitize(message)
+        expected_message = "ERROR: Failed to download metadata for repo 'packages-microsoft-com-prod': Status code: 401 for https://cec-aa.jfrog.io/artifactory/glib-rpm-hel9-lts-microsoft-com/repodata/repomd.xml"
+        self.assertEqual(sanitized, expected_message)
+
+    def test_sanitize_credentials_curl_error_buildbot_token(self):
```
rane-rajasi: Alternate naming options:

test_sanitize_credentials_with_error_and_secrets_in_input_str()

src/core/src/service_interfaces/CredentialSanitizer.py
```diff
@@ -0,0 +1,50 @@
...
+            return sanitized_message
+        except Exception:
```
rane-rajasi: Since this is a catch-all Exception and you are returning the original input as is, we should log the exception. In real time, if the code flow reaches here, this log would give details on why the input was never sanitized

src/core/tests/Test_TelemetryWriter.py
```diff
@@ -311,5 +314,121 @@ def test_write_event_with_buffer_true_and_empty_string_and_then_flush_with_non_e
...
+                           "Cannot download repomd.xml: All mirrors were tried")
+        self.assertEqual(sanitized, expected_message)
+
+    def test_sanitize_credentials_exception_handling(self):
+        """ Test exception handling: passing None should return the input unchanged """
+        result = CredentialSanitizer.sanitize(None)
+        self.assertIsNone(result)
+
+    def test_inject_fake_sanitizer_and_verify_invocation(self):
+        """ Integration Test: Can inject a fake sanitizer and verify it was invoked during write_event """
```
rane-rajasi: What is the need for this fake injection testing? Do you have any examples of when this could happen?

src/extension/src/CredentialSanitizer.py
```diff
@@ -0,0 +1,50 @@
...
+# limitations under the License.
+#
+# Requires Python 2.7+
+
+import re
+
+
+class CredentialSanitizer(object):
+    """Sanitizes credential-like values from URIs. Removes password/token from URI userinfo."""
+
```
rane-rajasi, line 25: Follow the same instructions/feedback as core/ for extension/ (All PR comments)

extension/ does not use Bootstrapper-ConfigurationFactory like core but uses a similar precedent of managing all class initializations together in code flow. See __main__.py for extension/

src/extension/tests/Test_TelemetryWriter.py
```diff
@@ -158,10 +161,170 @@ def test_events_deleted_outside_of_extension_while_extension_is_running(self):
...
+     unittest.TextTestRunner(verbosity=2).run(SUITE)
+
+
+
+
+
+
+
+
+
```
rane-rajasi, line 213: Remove all extra lines

rane-rajasi (changes requested):

src/extension/src/CredentialSanitizer.py
yashnap: Working on this

src/core/src/service_interfaces/CredentialSanitizer.py
yashnap: I can inline it but since it is not a simple regex I kept it multi-line for readability.  Is there a hard rule for keeping everything under 120 characters inline ?

src/core/tests/Test_TelemetryWriter.py
yashnap: It is to make sure that dependency injections work and always sanitizes before persisting. Basically it is more for Invocation and wiring logic check.

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2020 Microsoft Corporation
+# Copyright 2026 Microsoft Corporation
```
rane-rajasi: Keep this as 2020, only update the year for any new files you've added

src/core/src/service_interfaces/CredentialSanitizer.py
```diff
@@ -0,0 +1,49 @@
...
+            # (2) username: one or more non-whitespace, non-slash, non-colon, non-@ characters
+            # (3) password: zero or more non-whitespace, non-slash, non-@ characters
+            sanitized_message = re.sub(
+                r'(https?://|ftp://)([^:/@\s]+):([^@/\s]*)@',
+                r'\1\2@',
+                message
+            )
+            return sanitized_message
+        except Exception as error:
+            logging.error("Error occurred while sanitizing credentials from message: %s", repr(error))
```
rane-rajasi: Which logging mechanism is this using? Use LPE's own logging setup, refer how CompositeLogger is used in all classes

src/core/src/bootstrap/ConfigurationFactory.py
```diff
@@ -151,9 +152,14 @@ def new_bootstrap_configuration(config_env, log_file_path, events_folder, teleme
                     'telemetry_writer': None  # Has to be initialized without telemetry_writer to avoid running into a circular dependency loop. Telemetry writer within composite logger will be set later after telemetry writer has been initialized
                 }
             },
+            'credential_sanitizer': {
+                'component': CredentialSanitizer,
+                'component_args': [],
```
rane-rajasi: This will need composite logger for logging

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -23,21 +23,23 @@
...
-    def __init__(self, env_layer, composite_logger, events_folder_path, telemetry_supported):
+    def __init__(self, env_layer, composite_logger, events_folder_path, telemetry_supported, credential_sanitizer=None):
         self.env_layer = env_layer
         self.composite_logger = composite_logger
         self.__operation_id = str(datetime.datetime.utcnow())
         self.__task_name_watermark = "_" + str(datetime.datetime.utcnow().hour) + ":" + str(datetime.datetime.utcnow().minute) + ":" + str(datetime.datetime.utcnow().second) + "_" + str(os.getpid())
         self.__task_name = Constants.TelemetryTaskName.STARTUP + self.__task_name_watermark
         self.events_folder_path = None
         self.__telemetry_event_counter = 1  # will be added at the end of each event sent to telemetry to assist in tracing and identifying event/message loss in telemetry
+        self.credential_sanitizer = credential_sanitizer or CredentialSanitizer()
```
rane-rajasi: Bootstrapper-Container is responsible for initializing CredentialSanitizer(). You won't need to initialize it here

This will just be: self.credential_sanitize = credential_sanitizer

src/core/src/service_interfaces/TelemetryWriter.py
rane-rajasi: Bootstrapper-Container is responsible for initializing credential_sanitizer. With proper setup this will never be None. Incase python is forcing you to set a default value, it's because this comes after keyword arguments. Change the order to:
def __init__(self, env_layer, composite_logger, credential_sanitizer, events_folder_path, telemetry_supported):

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -23,21 +23,23 @@
 import time
 
 from core.src.bootstrap.Constants import Constants
+from core.src.service_interfaces.CredentialSanitizer import CredentialSanitizer
```
rane-rajasi: This import should also go away once the Bootstrapper setup is complete

src/core/src/service_interfaces/CredentialSanitizer.py
```diff
@@ -0,0 +1,49 @@
...
+            # Handles credentials containing special characters (except @, /, whitespace)
+            # Groups:
+            # (1) scheme: https://, http://, or ftp://
+            # (2) username: one or more non-whitespace, non-slash, non-colon, non-@ characters
+            # (3) password: zero or more non-whitespace, non-slash, non-@ characters
+            sanitized_message = re.sub(
+                r'(https?://|ftp://)([^:/@\s]+):([^@/\s]*)@',
+                r'\1\2@',
+                message
+            )
```
rane-rajasi: Add a log describing what was achieved here, something like:

self.composite_logger.log_verbose("Input message was sanitized to remove sensitive information. [InputMessage={0}][SanitizedMessage={1}]".format(str(message), str(sanitized_message)))

Few pointers:
1. "__ emove sensitive information.__  [InputMessage={0}][SanitizedMessage={1}]" <- This is the logging format used in LPE. All the data to be logged is within [] at the end of the statement.
2. Since this statement logs the input which contains credentials, only use composite_logger.log_verbose() since all other logging functions send data to telemetry as well. log_verbose is the only one not sending to telemetry and will only log in the local log file

src/core/src/service_interfaces/CredentialSanitizer.py
rane-rajasi: In addition, this log statement should read something like:

self.composite_logger.log_error("Error occurred while sanitizing credentials from message. [Error={0}]".format(repr(error)))

NOTE: Do not add input message to this log statement because this will go in telemetry.

rane-rajasi: More thoughts:
Exception's error text might also contain credentials, see if you can reproduce this scenario on a VM to identify what exception is thrown.

@kjohn-msft, what are your thoughts on adding this to error objects? Imo, this should not show up in error objects because it's not a customer issue/ask/something they can resolve. This is a safety initiative from our end. On the other hand, if sanitizing errors out and it's not brought to our attention, we keep missing it and leaking credentials.

src/core/src/service_interfaces/CredentialSanitizer.py
rane-rajasi: More code lines, especially empty ones, end up adding to the file size which can also affect readability. Plus, it adds to the overall build size. LPE build is shipped to and resides within a VM, and so we try to keep our overall size as low as possible, to ensure we don't end up consuming too much space on customer machines. 

In my opinion, the comment on lines 35-40 explains what this regex is trying to accomplish and that should be comprehensible for anyone reading this code. So, I would reformat this to a single line. 
Leaving it up to you to decide, in either case, don't keep the closing bracket (line 45) on a separate line

src/core/src/service_interfaces/CredentialSanitizer.py
```diff
@@ -0,0 +1,49 @@
...
+            return sanitized_message
+        except Exception as error:
+            logging.error("Error occurred while sanitizing credentials from message: %s", repr(error))
+            return message
```
rane-rajasi, line 43: An LPE standard, we keep some empty lines at the end of the file (ideally 2 but atleast 1)

This case is an exception to the standard of keeping empty lines to a minimum. It's because of the way LPE build is structured where all the core classes are stacked together in one python file, which is why we add the line separation.

If you want to see how the build is, follow the first instruction under "Build and Test locally" in README and see the files that are generated.

src/core/tests/Test_TelemetryWriter.py
```diff
@@ -71,7 +74,7 @@ def test_write_event(self):
 
         self.assertTrue(telemetry_event_counter_in_first_test_event is not None)
         self.assertTrue(telemetry_event_counter_in_second_test_event is not None)
-        self.assertTrue(int(telemetry_event_counter_in_second_test_event) - int(telemetry_event_counter_in_first_test_event) == 1)
+        self.assertTrue(int(telemetry_event_counter_in_second_test_event) - int(telemetry_event_counter_in_first_test_event) == 1 if telemetry_event_counter_in_first_test_event and telemetry_event_counter_in_second_test_event else False)
```
rane-rajasi: Why was this changed? There are already not None assertions for both of these in the lines above

src/core/tests/Test_TelemetryWriter.py
```diff
@@ -311,5 +314,180 @@ def test_write_event_with_buffer_true_and_empty_string_and_then_flush_with_non_e
...
+
+        latest_event_file = [pos_json for pos_json in os.listdir(self.runtime.telemetry_writer.events_folder_path) if re.search('^[0-9]+.json$', pos_json)][-1]
+        with open(os.path.join(self.runtime.telemetry_writer.events_folder_path, latest_event_file), 'r+') as f:
+            events = json.load(f)
+            self.assertTrue(events is not None)
+            self.assertEqual(events[-1]["TaskName"], "Test Task")
+            # Password should be removed
+            self.assertNotIn("password123", events[-1]["Message"])
+            # Username should be preserved
+            self.assertIn("user123@example.com", events[-1]["Message"])
```
rane-rajasi: Instead of these multiple assertions, you can just have one that matches the complete message

Applicable for other tests too

src/core/tests/Test_TelemetryWriter.py
```diff
@@ -311,5 +314,180 @@ def test_write_event_with_buffer_true_and_empty_string_and_then_flush_with_non_e
...
+            self.assertNotIn("pass2", events[-1]["Message"])
+            # Usernames should be preserved
+            self.assertIn("user1@host1.com", events[-1]["Message"])
+            self.assertIn("user2@host2.com", events[-1]["Message"])
+            # Verify entire message matches expected output (excluding TC counter)
+            message_without_tc = events[-1]["Message"][:events[-1]["Message"].rfind(" [TC=")]
+            self.assertIn("Failed to fetch from https://user1@host1.com/api and http://user2@host2.com/data", message_without_tc)
+            f.close()
+
+    def test_sanitize_credentials_with_no_credentials_in_input_leak_in_input(self):
```
rane-rajasi: Name is confusing. This could be:

test_sanitize_credentials_from_error_message_with_no_cred_leak()

src/core/tests/Test_TelemetryWriter.py
```diff
@@ -311,5 +314,180 @@ def test_write_event_with_buffer_true_and_empty_string_and_then_flush_with_non_e
...
+            events = json.load(f)
+            self.assertTrue(events is not None)
+            self.assertEqual(events[-1]["TaskName"], "Test Task")
+            # Password should be removed
+            self.assertNotIn("password123", events[-1]["Message"])
+            # Username should be preserved
+            self.assertIn("user123@example.com", events[-1]["Message"])
+            f.close()
+
+    def test_sanitize_credentials_multiple_urls_with_credentials_leak_in_input(self):
```
rane-rajasi: optional name:
test_sanitize_credentials_from_text_with_multiple_cred_leaks()

src/core/tests/Test_TelemetryWriter.py
rane-rajasi: When will a dependency injection need sanitization?

rane-rajasi: Looking at the implementation, this is manually doing what the test suite is already built to do. 

In your other tests, you invoke TelemtryWriter.write_event(), fetch events from the events folder and validate what's written in those events.

Which means you are also validating all the steps you have listed in here: i.e:
1. Write an event
2. Verify mock sanitizer was called
3. Verify the call was made with a message containing the original error info
4. Verify telemetry event was written with the mock-sanitized message

This test suite is designed to imitate real time execution flow and to not need individual manual mocking of a complete e2e scenario in each tests.

You can remove this test, and if you still feel the need of additional verification, in your other tests, before calling TelemtryWriter.write_event(), check the events folder is empty. After calling write_event(), assert that eventsfolder only contains the message you sent in write_event()

src/extension/src/CredentialSanitizer.py
```diff
@@ -0,0 +1,49 @@
...
+# Requires Python 2.7+
+
+import logging
+import re
+
+
+class CredentialSanitizer(object):
+    """Service that sanitizes credential-like values from URIs by removing password/token from URI userinfo."""
+
+    def __init__(self):
```
rane-rajasi: Same comments on logging as mentioned in core/

Use composite_logger, less lines for sanitized_message regex and empty lines at the end of the file

src/extension/src/TelemetryWriter.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2020 Microsoft Corporation
+# Copyright 2026 Microsoft Corporation
```
rane-rajasi: revert to 2020

src/extension/src/TelemetryWriter.py
```diff
@@ -37,14 +38,20 @@ def __init__(self, logger, env_layer):
         self.__agent_is_compatible = self.__get_agent_supports_telemetry_from_env_var()
         self.__task_name_watermark = "." + str(datetime.datetime.utcnow().hour) + "." + str(datetime.datetime.utcnow().minute) + "." + str(datetime.datetime.utcnow().second) + "." + str(os.getpid())
         self.__task_name = Constants.TELEMETRY_TASK_NAME + self.__task_name_watermark
+        self.credential_sanitizer = credential_sanitizer or CredentialSanitizer()
```
rane-rajasi: Same comment as in core\

src/extension/tests/Test_TelemetryWriter.py
```diff
@@ -158,10 +161,209 @@ def test_events_deleted_outside_of_extension_while_extension_is_running(self):
...
 
-if __name__ == '__main__':
-    SUITE = unittest.TestLoader().loadTestsFromTestCase(TestTelemetryWriter)
-    unittest.TextTestRunner(verbosity=2).run(SUITE)
+        Args:
+            message: The message to write to telemetry
+
+        Returns:
+            The sanitized message from the event
+        """
```
rane-rajasi, line 170: remove blank and extra lines

src/core/tests/Test_TelemetryWriter.py
```diff
@@ -19,7 +19,10 @@
 import re
 import time
 import unittest
+from unittest.mock import Mock
 from core.src.bootstrap.Constants import Constants
+from core.src.service_interfaces.CredentialSanitizer import CredentialSanitizer
```
rane-rajasi: This should not be needed now

src/core/tests/Test_TelemetryWriter.py
rane-rajasi: This is still pending @yashnap

src/core/tests/Test_TelemetryWriter.py
rane-rajasi: Still pending

src/core/tests/Test_TelemetryWriter.py
rane-rajasi: Current name works too, I'm thinking we don't need '<>_in_input' at the end

rane-rajasi (changes requested):
All the comments in core\ apply to extension\ too

src/core/src/service_interfaces/CredentialSanitizer.py
yashnap: Agree with your thought process about adding this into error object. 

My thoughts:
The message first goes through the restriction compliance checks and then into the sanitization step. If sanitization fails and we return the original message, we risk leaking credentials if the message contains sensitive data. On the other hand, if we completely drop the message, we lose useful information even in cases where the message did not contain anything sensitive.
If sanitization fails, we should return a safe placeholder or a redacted version instead.

Thoughts ? @rane-rajasi @kjohn-msft

yashnap:
Working on fixing coverage

src/core/src/service_interfaces/CredentialSanitizer.py
kjohn-msft: We are not cleaning up a real credential here (at this time). The current ask is to avoid false positive detections upstream in a specific telemetry pathway. Telemetry writer has credsanitizer injected and it's consumed. We can leave it at that for now.

src/core/src/service_interfaces/CredentialSanitizer.py
```diff
@@ -0,0 +1,46 @@
...
+        """
+        try:
+            # Pattern matches: scheme://user:password@host  →  scheme://user@host
+            # Handles credentials containing special characters (except @, /, whitespace)
+            # Groups:
+            # (1) scheme: https://, http://, or ftp://
+            # (2) username: one or more non-whitespace, non-slash, non-colon, non-@ characters
+            # (3) password: zero or more non-whitespace, non-slash, non-@ characters
+            sanitized_message = re.sub(
+                r'(https?://|ftp://)([^:/@\s]+):([^@/\s]*)@',r'\1\2@',message)
```
rane-rajasi: nit: move line 40 to 39

i.e. sanitized_message = re.sub(r'(https?://|ftp://)([^:/@\s]+):([^@/\s]*)@',r'\1\2@',message)

src/core/src/service_interfaces/CredentialSanitizer.py
```diff
@@ -0,0 +1,46 @@
...
+            self.composite_logger.log_verbose("Message was sanitized to remove sensitive information. [InputMessage={0}][SanitizedMessage={1}]".format(str(message), str(sanitized_message)))
+            return sanitized_message
+        except Exception as error:
+            self.composite_logger.log_error("Error occurred while sanitizing credentials from message: {0}".format(repr(error)))
```
rane-rajasi: nit: 
self.composite_logger.log_error("Error occurred while sanitizing credentials from message: [Error={0}]".format(repr(error)))

We use this logging format throughout LPE

src/core/tests/Test_TelemetryWriter.py
rane-rajasi: @yashnap this q is still open

src/extension/src/TelemetryWriter.py
```diff
@@ -68,6 +74,7 @@ def __ensure_message_restriction_compliance(self, full_message):
             self.logger.log_telemetry_module_error("Error occurred while formatting message for a telemetry event. [Error={0}]".format(repr(e)))
             raise
 
+
```
rane-rajasi: Remove the extra line break

src/core/tests/Test_TelemetryWriter.py
```diff
@@ -311,5 +311,145 @@ def test_write_event_with_buffer_true_and_empty_string_and_then_flush_with_non_e
...
+        event_files_count = len([f for f in os.listdir(self.runtime.telemetry_writer.events_folder_path) if re.search('^[0-9]+.json$', f)])
+        self.assertEqual(event_files_count, 1, "Events folder should contain exactly one event file")
+
+        with open(os.path.join(self.runtime.telemetry_writer.events_folder_path, latest_event_file), 'r+') as f:
+            events = json.load(f)
+            self.assertTrue(events is not None)
+            self.assertEqual(events[-1]["TaskName"], "Test Task")
+            message_without_tc = events[-1]["Message"][:events[-1]["Message"].rfind(" [TC=")]
+            self.assertEqual("Error connecting to https://testuser@invalid.repo.example/rpm/repodata/repomd.xml", message_without_tc)
+            f.close()
```
rane-rajasi: Extract out common code reused in multiple tests to their own functions.

For eg: one function to open and read event from events folder, one to validate one event file and expected vs actual event (this internally calls the previous one), one for any setup process (such as clearing events folder and the assert), etc

src/extension/src/CredentialSanitizer.py
```diff
@@ -0,0 +1,47 @@
...
+            # Pattern matches: scheme://user:password@host  →  scheme://user@host
+            # Handles credentials containing special characters (except @, /, whitespace)
+            # Groups:
+            # (1) scheme: https://, http://, or ftp://
+            # (2) username: one or more non-whitespace, non-slash, non-colon, non-@ characters
+            # (3) password: zero or more non-whitespace, non-slash, non-@ characters
+            sanitized_message = re.sub(
+                r'(https?://|ftp://)([^:/@\s]+):([^@/\s]*)@',r'\1\2@',message)
+            self.logger.log_verbose(
+                "Message was sanitized to remove sensitive information. [InputMessage={0}][SanitizedMessage={1}]".format(str(message), str(sanitized_message)))
```
rane-rajasi: nit: single line for sanitized_message  and the log statement

src/extension/tests/Test_TelemetryWriter.py
```diff
@@ -5,8 +5,8 @@
 import time
 import unittest
 from extension.src.Constants import Constants
-from extension.tests.helpers.VirtualTerminal import VirtualTerminal
 from extension.tests.helpers.RuntimeComposer import RuntimeComposer
+from extension.tests.helpers.VirtualTerminal import VirtualTerminal
```
rane-rajasi, line 11: Revert this to its original version i.e. move this to line 8. This goes with not changing anything outside of your PRs footprint

src/extension/tests/Test_TelemetryWriter.py
```diff
@@ -158,10 +158,129 @@ def test_events_deleted_outside_of_extension_while_extension_is_running(self):
...
+        message = "Failed to fetch from https://user1:pass1@host1.com/api and http://user2:pass2@host2.com/data"
+        self.telemetry_writer.write_event(message, Constants.TelemetryEventLevel.Error, "Test Task")
+
+        event_files = os.listdir(self.telemetry_writer.events_folder_path)
+        with open(os.path.join(self.telemetry_writer.events_folder_path, event_files[0]), 'r+') as f:
+            events = json.load(f)
+            self.assertTrue(events is not None)
+            self.assertEqual(events[-1]["TaskName"], "Test Task")
+            self.assertEqual("Failed to fetch from https://user1@host1.com/api and http://user2@host2.com/data",
+                             events[-1]["Message"])
```
rane-rajasi: nit: single line

src/extension/tests/Test_TelemetryWriter.py
```diff
@@ -158,10 +158,129 @@ def test_events_deleted_outside_of_extension_while_extension_is_running(self):
         self.telemetry_writer.write_event("testing telemetry write to file", Constants.TelemetryEventLevel.Error, "Test Task")
         os.listdir = backup_os_listdir
 
+    # ==================== Integration test for credential sanitization in telemetry ====================
```
rane-rajasi: nit: These are Unit tests

src/extension/tests/Test_TelemetryWriter.py
```diff
@@ -158,10 +158,129 @@ def test_events_deleted_outside_of_extension_while_extension_is_running(self):
...
+    def _load_sanitized_event(self, message):
+        """
+        Helper method to write event to telemetry and load the sanitized message.
+        The regex sanitization happens automatically in TelemetryWriter.
+        Args:
+            message: The message to write to telemetry
+        Returns: The sanitized message from the event
+        """
+        if self.runtime.is_github_runner:
+            return None
```
rane-rajasi: Remove this check if you want this test to run always (for CI coverage), so you won't need to set and reset it in tests

src/extension/tests/Test_TelemetryWriter.py
```diff
@@ -158,10 +158,129 @@ def test_events_deleted_outside_of_extension_while_extension_is_running(self):
...
+        self.telemetry_writer.write_event(message, Constants.TelemetryEventLevel.Error, "Test Task")
+
+        event_files = os.listdir(self.telemetry_writer.events_folder_path)
+        with open(os.path.join(self.telemetry_writer.events_folder_path, event_files[0]), 'r+') as f:
+            events = json.load(f)
+            self.assertTrue(events is not None)
+            self.assertEqual(events[-1]["TaskName"], "Test Task")
+            self.assertEqual("Failed to fetch from https://user1@host1.com/api and http://user2@host2.com/data",
+                             events[-1]["Message"])
+            f.close()
```
rane-rajasi: Why not use _load_sanitized_event() here?

rane-rajasi: Refer my comment in core/Test_TelemetryWriter on extracting code reused across tests

src/extension/tests/Test_TelemetryWriter.py
```diff
@@ -158,10 +158,129 @@ def test_events_deleted_outside_of_extension_while_extension_is_running(self):
...
 
+        # On non-GitHub runner, should return the sanitized message
+        self.assertIsNotNone(result)
+        self.assertIn("user@example.com", result)
+        self.assertEqual("https://user@example.com", result)
 
+        # Restore
+        self.runtime.is_github_runner = original_is_github_runner
 
+    def test_sanitize_credentials_multiple_urls_with_credentials_leak_in_input(self):
```
rane-rajasi: Follow the function naming format you've used in core TelemetryWriter tests

rane-rajasi: Applies for all function names in this file

rane-rajasi (changes requested):
Feedback comments inline. I've closed all of my comments from previous reviews that were addressed, the ones still pending are left open/unresolved. Make sure you address all the open comments from previous reviews too.

And codecoverage checks are failing, please address those

src/core/tests/Test_TelemetryWriter.py
yashnap: This was updated accidently. I've reverted back the change

src/core/src/bootstrap/Bootstrapper.py
```diff
@@ -49,6 +49,7 @@ def __init__(self, argv, capture_stdout=True):
         if capture_stdout:
             self.stdout_file_mirror = StdOutFileMirror(self.env_layer, self.file_logger)
         self.composite_logger = self.container.get('composite_logger')
+        self.credential_sanitizer = self.container.get("credential_sanitizer")
```
michellemcdaniel, line 52: test comment

src/extension/src/TelemetryWriter.py
```diff
@@ -37,14 +37,20 @@ def __init__(self, logger, env_layer):
         self.__agent_is_compatible = self.__get_agent_supports_telemetry_from_env_var()
         self.__task_name_watermark = "." + str(datetime.datetime.utcnow().hour) + "." + str(datetime.datetime.utcnow().minute) + "." + str(datetime.datetime.utcnow().second) + "." + str(os.getpid())
         self.__task_name = Constants.TELEMETRY_TASK_NAME + self.__task_name_watermark
+        self.credential_sanitizer = credential_sanitizer
 
     def __new_event_json(self, event_level, message, task_name):
+        # Step 1: Apply message restrictions (formatting, truncation)
+        restricted_message = self.__ensure_message_restriction_compliance(message)
+        # Step 2: Sanitize credentials from URIs
+        sanitized_message = self.credential_sanitizer.sanitize(restricted_message)
```
kjohn-msft: Sanitization can break message restriction (size) compliance so this is the wrong order of execution.

src/core/src/service_interfaces/TelemetryWriter.py
```diff
@@ -127,12 +128,17 @@ def __get_events_folder_path_exists(events_folder_path):
         return events_folder_path is not None and os.path.exists(events_folder_path)
 
     def __new_event_json(self, event_level, message, task_name):
+        # Step 1: Apply message restrictions (formatting, truncation)
+        restricted_message = self.__ensure_message_restriction_compliance(message)
+        # Step 2: Sanitize credentials from URIs
+        sanitized_message = self.credential_sanitizer.sanitize(restricted_message)
```
kjohn-msft: Sanitization can break message restriction (size) compliance so this is the wrong order of execution.

kjohn-msft (changes requested):

src/core/src/service_interfaces/TelemetryWriter.py
yashnap: Fixed the ordering

src/extension/src/TelemetryWriter.py
yashnap: Fixed

yashnap:
@rane-rajasi / @kjohn-msft I am trying to understand the reason behind Code Coverage Failure. When I try UT's from local everything works fine. I even tried to set GitHub Runner_Temp as true in the code and mimic the github CI run and it shows proper coverage but when it runs on the CI it shows as no coverage for the test class at all : https://app.codecov.io/gh/Azure/LinuxPatchExtension/pull/340?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=Azure 
I've spend a lot of time understanding and debugging but having a hard time getting it right. Any thoughts/hints would be helpful

src/core/src/service_interfaces/CredentialSanitizer.py
```diff
@@ -0,0 +1,44 @@
...
+        """
+        try:
+            # Pattern matches: scheme://user:password@host ->  scheme://user@host
+            # Handles credentials containing special characters (except @, /, whitespace)
+            # Groups:
+            # (1) scheme: https://, http://, or ftp://
+            # (2) username: one or more non-whitespace, non-slash, non-colon, non-@ characters
+            # (3) password: zero or more non-whitespace, non-slash, non-@ characters
+            sanitized_message = re.sub(r'(https?://|ftp://)([^:/@\s]+):([^@/\s]*)@',r'\1\2@',message)
+            self.composite_logger.log_verbose("Message was sanitized to remove sensitive information. [InputMessage={0}][SanitizedMessage={1}]".format(str(message), str(sanitized_message)))
```
nikhim-um, line 39: logging original message defeat's purpose isn't it? why do we want to log InputMessage which has credentials?

src/core/src/service_interfaces/CredentialSanitizer.py
```diff
@@ -0,0 +1,44 @@
...
+            return sanitized_message
+        except Exception as error:
+            self.composite_logger.log_error("Error occurred while sanitizing credentials from message: [Error={0}]".format(repr(error)))
+            return message
```
nikhim-um, line 43: I think we should not be returning message which contain credentials in case of exception. when we cannot sanitize i think its better to take safer route to avoid credential leak. --> check for alternatives instead of returning original message

nikhim-um (changes requested):

src/core/src/service_interfaces/CredentialSanitizer.py
yashnap: Please refer this thread : https://github.com/Azure/LinuxPatchExtension/pull/340#discussion_r3175371194

yashnap:
@rane-rajasi  @kjohn-msft 
I spent some time investigating the GitHub coverage issue that started appearing with this PR.
Initially, I suspected the newly added CredentialSanitizer integration, or the write_event() path was not being executed in CI. To validate this, I added temporary debug logging and verified that:

RuntimeComposer is constructed successfully.
CredentialSanitizer is instantiated.
TelemetryWriter.write_event() is entered.
The code reaches __new_event_json().
The new sanitization path is exercised.

During troubleshooting, I noticed that RuntimeComposer overrides tempfile.mkdtemp() on GitHub runners:
`if self.is_github_runner:
    def mkdtemp_runner():
        temp_path = os.path.join(os.getenv('RUNNER_TEMP'), str(uuid.uuid4()))
        os.mkdir(temp_path)
        return temp_path
    tempfile.mkdtemp = mkdtemp_runner`

I tried removing this override, and after doing so the coverage issue disappeared and the tests/coverage behaved as expected.
At the moment, my working theory is that the custom mkdtemp() override was causing GitHub-specific test behavior differences in the Extension test suite. The override changes the behavior of the standard library tempfile.mkdtemp() globally for the entire test process, and some Extension tests appear to be sensitive to that behavior.

What is still unclear to me is why the same pattern continues to work in the Core test framework (RuntimeCompositor) without causing similar issues. It is possible that the Core tests are designed around this behavior, whereas the Extension tests rely on the standard tempfile implementation or interact with temporary paths differently.

I've currently removed the GitHub-specific mkdtemp() override from RuntimeComposer, which resolves the coverage issue for this PR.

Does anyone know the original reason this override was introduced in the Extension test framework, or have context on why GitHub runners require it here? I'd appreciate any suggestions or explanations before finalizing the change. I see this PR from earlier where this change was added : https://github.com/Azure/LinuxPatchExtension/pull/129 but unclear on why its failing now

src/extension/src/TelemetryWriter.py
michellemcdaniel, line 112: is this this only place where we care that the message is sanitized? If so, then this is fine. If not, we may want to sanitize the message before sending it to __new_event_json so that all logging of it here is sanitized.

yashnap: For this particular issue, we only want to sanitize it here before it gets uploaded to geneva telemetry which is what the credscan reads and flags.

kjohn-msft (approved):

michellemcdaniel (approved):

rane-rajasi (approved):

nikhim-um (approved):

PR #341, opened 2026-04-22
Feature: [Livepatching][MVP] Extension changes to support livepatching in Canonical

rane-rajasi:
**Livepatching in Canonical**:
The apt package manager is responsible for installing .deb packages on Ubuntu LTS (long-term support) and interim releases, including the .deb package for the Linux kernel. Updating kernel packages require a system restart and leave the system vulnerable between the update and a restart. Canonical has offered livepatch as a solution to protect the system during this vulnerable period. Livepatch will modify the vulnerable kernel code in memory, and re-direct function calls to this version before a reboot.
We aim to leverage this livepatch solution in our product to offer customers the ability to have their VMs patches without needing frequent reboot.

Ubuntu has introduced a timestamp based livepatch rollout, which will restrict the livepatch being applied/installed to only the ones available on or before said timestamp
For eg:
$ canonical-livepatch config cutoff-date="2024-10-01T12:00:00Z" 
NOTE: cutoff-date must be in the past 
Once the cutoff-date is set, a Livepatch client, during its next check-in, will apply any livepatches available <= cutoff-date
Livepatch client: software running on a machine, that periodically checks for the availability of new patches. Once new patches are available, they are downloaded, verified, and applied to the current kernel. 

**Ref**:
[Live Linux kernel patching with progressive timestamped rollouts | Ubuntu](https://ubuntu.com/blog/live-linux-kernel-patching-with-progressive-timestamped-rollouts) 
[Livepatch Documentation | Ubuntu](https://ubuntu.com/security/livepatch/docs)

**AzGPS solution**:
The solution implemented in this PR is an MVP to demonstrate how AzGPS will incorporate livepatching. In order to achieve MVP, we are adding livepatching as a part of Patch installation operation. i.e. Livepatching will only be performed during an Installation operation and livepatches will be applied before any regular patches are installed on the machine.

- **Customer pre-reqs**:
        - Feature flag: Customers need to opt-in to livepatching via an in-VM configuration. Only the customers opted in to livepatching will be considered.
        - Livepatching is only available on Ubuntu LTS VMs with a paid pro subscription. Customer needs to attach a paid pro subscription and ensure 'livepatch' pro service is enabled on their VMs.
   
- AzGPS reads the in-VM config/feature flag set by the customer. 
- Configure Patching and Assess Patches are completed without any changes.
- During patch installation, if livepatching is requested, as a first step, 
          - AzGPS will ensure all other pre-reqs are met.
          - Set cutoff date for canonical livepatch service using max_patch_publish_date
          - Trigger Livepatch client (this is to ensure we fetch and report the latest livepatch status)
          - Fetch and report livepatch status as an individual patch entry in 'PatchInstallationSummary' subsection of the status blob
          - Continue with the regular patch installation

**Reporting livepatch status in AzGPS**:
- Canonical provides the following livepatch status:
{
              "Client-Version": "<>",
    "Machine-Id": "<>",
    "Architecture": "<>",
    "CPU-Model": "<>",
    "Last-Check": "<>",
    "Boot-Time": "<>",
    "Uptime": "<>",
    "Status": [
    {
        "Kernel": "<>",
        "Running": true,
        "Livepatch": {
            "CheckState": "checked",
            "State": "<>", // "nothing-to-apply" or "applied"
            "Version": "" // "" or a version such as "1.0",
            "Fixes": // empty if no livepatches available or a list of CVEs installed
            [{
                "Name": "<>", //cve identifier such as CVE-000-0000
                "Description": "<>", // description of the livepatch fix
                "Bug": "",
                "Patched": <bool> // boolean value indicating status
            }]
        },
        "Supported": "<>", // "supported" or a quick text on what is needed such as "kernel-upgrade-required"
        "UpgradeRequiredDate": "<>" // date
    }],
    "tier": "updates",
    "Excluded-LSNs": [], // List of excluded LSNs
    "Fixed-CVEs": {
        "Timestamp": "",
        "Kernel-Package-Fixes": [], // list of all kernel packages fixed
        "Installed-Kernels": [],
        "Patched-CVEs": [], // list of patched CVEs identifiers
        "Digest": ""
    },
    "Blocking-Options": [ // List of configs blocking livepatch, if any. For eg: cutoff-date set for livepatch client
        "cutoff-date"
    ],
    "Using-Cutoff-Date": <bool> // boolean value indicating whether livepatch client is using cutoff-date config or not
}

- From this livepatch status, we fetch "CheckState", "State" and "Version" from [Status][Livepatch] and add it as a single patch entry in PatchInstallationSummary. with patch_name: "livepatch_"<CheckState>_<State> and version: <Version>
- For eg: A livepatch status added as a singular patch within PatchInstallationSummary: 
{\"patchId\": \"livepatch_checked_nothing-to-apply__Ubuntu_22.04\", \"name\": \"livepatch_checked_nothing-to-apply\", \"version\": \"\", \"classifications\": [\"Other\"], \"patchInstallationState\": \"NotSelected\"}

**Logs from in VM test runs**:
1. Test case: Livepatching is not requested by the customer.
    Expected output: No livepatches applied: 
[5.core.log](https://github.com/user-attachments/files/27013520/5.core.log)
[5.status.txt](https://github.com/user-attachments/files/27013567/5.status.txt)

2. Test case: Livepatching is requested by the customer and not enabled on the machine (i.e. all livepatching pre-reqs not enabled For eg: pro sub or setting livepatch state to enabled in pro services): 
    Expected output: No livepatches applied with or without error added in status
   - Test case 1: livepatch service not enabled: 
      Expected output: No livepatches applied, no error added in status: 
      [11.core.log](https://github.com/user-attachments/files/27016044/11.core.log)
      [11.status.txt](https://github.com/user-attachments/files/27016047/11.status.txt)
      sample run 2:
      [3.core.log](https://github.com/user-attachments/files/27025192/3.core.log)
      [3.status.txt](https://github.com/user-attachments/files/27025197/3.status.txt)
 
   - Test case 2: Pro NOT attached
      Expected output: No livepatches applied, no error added in status
      [4.core.log](https://github.com/user-attachments/files/27025603/4.core.log)
      [4.status.txt](https://github.com/user-attachments/files/27025605/4.status.txt)

   - Test case 3: Free pro sub attached:
      Expected ouput: No livepatches applied, error added in status blob: 
      [14.core.log](https://github.com/user-attachments/files/27022447/14.core.log)
      [14.status.txt](https://github.com/user-attachments/files/27022451/14.status.txt)
      sample run 2:
      [5.core.log](https://github.com/user-attachments/files/27026258/5.core.log)
      [5.status.txt](https://github.com/user-attachments/files/27026262/5.status.txt)

3. Test case: Livepatching is requested by the customer and enabled on the VM (i.e. pre-reqs met), but no livepatches available: 
    Expected output: No new livepatches applied, livepatch status reported in status blob with patch name ~= 'livepatch_checked_nothing-to-apply____' and an empty patch version
[10.core.log](https://github.com/user-attachments/files/27015220/10.core.log)
[10.status.txt](https://github.com/user-attachments/files/27015233/10.status.txt)
[livepatching.settings.txt](https://github.com/user-attachments/files/27015249/livepatching.settings.txt)
sample run 2:
[8.core.log](https://github.com/user-attachments/files/27027192/8.core.log)
[8.status.txt](https://github.com/user-attachments/files/27027196/8.status.txt)

image

4. Test case: Livepatching is requested by the customer and enabled on the VM (i.e. pre-reqs met) and livepatches are applied
    Expected output: Livepatches applied, livepatch status reported in status blob with patch name ~= 'livepatch_checked_applied____' and non-empty patch version


src/core/src/core_logic/PatchInstaller.py
```diff
@@ -58,14 +58,22 @@ def start_installation(self, simulate=False):
...
+        # check if livepatching is enabled,
+        # if yes, check if pro is attached
+        # set configdate, and launch livepatch client.
+        # fetch response and set it in patch installation status
+        if self.execution_config.livepatching_enabled:
+            package_manager.start_livepatching()
+
+        self.composite_logger.log("\nStarting patch installation... [MachineId: " + self.env_layer.platform.vm_name() + "][ActivityId: " + self.execution_config.activity_id + "][StartTime: " + self.execution_config.start_time + "][MaintenanceWindowDuration: " + self.execution_config.duration + "]")
+
+        self.stopwatch.start()
```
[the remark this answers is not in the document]
rane-rajasi: Stopwatch here is used to track only the patch installation duration and is nowhere used n maintenance window duration calculations. And since this livepatching solution is an MVP and will have it's own separate solution soon, the stopwatch measurements for installation should be kept separate from livepatching.

Maintenance window exceeded is purely calculated based on the customer input for the overall operation (CP + AP + IP) and as such will also include livepatching now. Stopwatch and maintenance window calculations and usages are not inter-dependent

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -91,6 +91,11 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
         
         self.package_install_expected_avg_time_in_seconds = 90  # As per telemetry data, the average time to install package is around 81 seconds for apt.
 
+        # Livepatching service
+        self.set_cutoff_date_in_livepatch_config_cmd = "canonical-livepatch config cutoff-date=" + execution_config.max_patch_publish_date
+        self.launch_livepatch_client_cmd = "sudo systemctl restart snap.canonical-livepatch.canonical-livepatchd"
+        self.get_livepatch_status_cmd = "canonical-livepatch status --verbose --format json"
+
```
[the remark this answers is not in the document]
rane-rajasi: In progress

src/core/src/package_managers/UbuntuProClient.py
```diff
@@ -172,3 +173,35 @@ def is_reboot_pending(self):
...
+                json_output = json.loads(output)
+                livepatch_service = next((service for service in json_output.get("services", []) if service.get('name').lower() == 'livepatch'), None)
+                self.composite_logger.log_debug("[APM][Pro] Livepatch service status from Ubuntu Pro Client: [LivepatchService={0}]".format(str(livepatch_service)))
+                livepatch_status = livepatch_service.get("status", "unknown") if livepatch_service is not None else "unknown"
+                if livepatch_status.lower() == 'enabled' or livepatch_status.lower() == 'warning':
+                    self.composite_logger.log_info("[APM][Pro] Livepatching is enabled for the machine.")
+                    return True
+                else:
+                    self.composite_logger.log_warning("[APM][Pro] Livepatching is NOT enabled for the machine.")
+                    return False
```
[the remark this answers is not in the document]
rane-rajasi: In progress

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -839,6 +844,117 @@ def do_processes_require_restart(self):
...
+            else:
+                self.composite_logger.log_warning("[APM] Failed to launch livepatch client. [Output={0}]".format(str(output)))
+        except Exception as error:
+            livepatch_launch_exception = repr(error)
+            self.composite_logger.log_warning("[APM] Exception while launching livepatch client. [Exception={0}]".format(livepatch_launch_exception))
+
+    def fetch_and_update_livepatch_status_in_status_blob(self):
+        """Fetches livepatch status and if a livepatch/es is/are applied, updates it as a new patch entry in PatchInstallationSummary"""
+        livepatch_status = self.try_get_livepatch_status()
+        if livepatch_status is not {}:
```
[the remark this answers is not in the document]
rane-rajasi: In progress

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -839,6 +844,117 @@ def do_processes_require_restart(self):
...
+        return livepatch_status
+
+    def update_livepatch_status_in_patch_installation_summary(self, livepatch_status):
+        """ Updates livepatch status in patch installation summary as a new patch record """
+        self.composite_logger.log_debug("[APM] Updating patch installation summary with livepatch status. [LivepatchStatus={0}]".format(str(livepatch_status)))
+        extracted_livepatch_fields = self.extract_livepatch_fields(livepatch_status)
+        check_state = extracted_livepatch_fields[0]["CheckState"]
+        state = extracted_livepatch_fields[0]["State"]
+        patch_name = "livepatch_" + check_state + "_" + state
+        patch_version = extracted_livepatch_fields[0]["PatchVersion"]
```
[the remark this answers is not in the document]
rane-rajasi: In progress

src/core/src/core_logic/ExecutionConfig.py
```diff
@@ -260,9 +265,48 @@ def __is_eula_accepted_for_all_patches(self):
 
         return is_eula_accepted
 
+    def __get_livepatching_config_in_json(self):
+        """ Reads customer provided config on live patching from disk and returns a dict with the config values.""
```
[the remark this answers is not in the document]
rane-rajasi: In Progress

src/core/src/core_logic/PatchInstaller.py
```diff
@@ -58,14 +58,22 @@ def start_installation(self, simulate=False):
...
         package_manager = self.package_manager
         reboot_manager = self.reboot_manager
 
+        # todo: Livepatching code
+        # check if livepatching is enabled,
+        # if yes, check if pro is attached
+        # set configdate, and launch livepatch client.
+        # fetch response and set it in patch installation status
+        if self.execution_config.livepatching_enabled:
+            package_manager.start_livepatching()
```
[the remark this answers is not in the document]
rane-rajasi: In Progress

src/core/src/package_managers/UbuntuProClient.py
```diff
@@ -172,3 +173,35 @@ def is_reboot_pending(self):
 
         self.composite_logger.log_debug("[APM][Pro] Ubuntu Pro Client Reboot Required: [UbuntuProClientSuccess={0}][RebootRequiredFlag={1}][Error={2}]".format(ubuntu_pro_client_api_success, ubuntu_pro_client_reboot_required, ubuntu_pro_client_exception))
         return ubuntu_pro_client_api_success, ubuntu_pro_client_reboot_required
+
+    # region Livepatching
+    def is_livepatching_applicable_for_machine(self):
+        """ Verifies if livepotching is applicable for the machine by checking if the machine is an Ubuntu LTS Pro VM """
```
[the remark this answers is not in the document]
rane-rajasi: In Progress

src/core/src/core_logic/ExecutionConfig.py
```diff
@@ -260,9 +265,49 @@ def __is_eula_accepted_for_all_patches(self):
...
+        is_livepatching_enabled = False
+
+        enable_livepatching = self.__fetch_specific_setting(livepatching_settings, Constants.LivePatchingSettings.ENABLE_LIVEPATCHING)
+        enabled_by = self.__fetch_specific_setting(livepatching_settings, Constants.LivePatchingSettings.ENABLED_BY)
+        last_modified = self.__fetch_specific_setting(livepatching_settings, Constants.LivePatchingSettings.LAST_MODIFIED)
+        if enable_livepatching is not None and enable_livepatching in [True, 'True', 'true', '1', 1]:
+            is_livepatching_enabled = True
+            self.composite_logger.log_debug("Livepatching config values read from disk: [EnableLivePatching={0}] [EnabledBy={1}] [LastModified={2}]. Computed value of [IsLivePatchingEnabled={3}]"
+                                            .format(str(enable_livepatching), str(enabled_by), str(last_modified), str(is_livepatching_enabled)))
+        else:
```
kjohn-msft, line 296: The opposite of true isn't implicitly false. It can always be true, false or 'do nothing' (either because of bad data or because of no intent). 

You're logging the computed intent but never logging the actual data read out.

src/core/src/core_logic/ExecutionConfig.py
```diff
@@ -260,9 +265,49 @@ def __is_eula_accepted_for_all_patches(self):
...
+                self.composite_logger.log_debug("Livepatching config values from disk: [Config={0}]".format(str(livepatching_config)))
+            else:
+                self.composite_logger.log_debug("No livepatching config found on the VM. Returning empty config.")
+        except Exception as error:
+            self.composite_logger.log_debug("Error occurred while reading and parsing live patching config. Returning empty config. Error=[{0}]".format(repr(error)))
+
+        return livepatching_config
+
+    def __is_livepatching_enabled(self, livepatching_settings):
+        """ Determines if livepatching is enabled or disabled. """
```
kjohn-msft: Ambiguous function naming and comment -- livepatching being enabled or not on the system is (separate) system state. Our goal is to capture the customer's intent on what they want us to do regardless of the system state. That goal wouldn't map 1:1 to this in naming given that separate state.

src/core/src/core_logic/ExecutionConfig.py
```diff
@@ -260,9 +265,49 @@ def __is_eula_accepted_for_all_patches(self):
 
         return is_eula_accepted
 
+    def __get_livepatching_config_in_json(self):
+        """ Reads customer provided config on live patching from disk and returns a dict with the config values.
```
kjohn-msft: nit: consistency in phrasing. livepatching as the noun is livepatch. (all versions without a space)

Comment in general applies everywhere.

kjohn-msft: Specifically use livepatch over livepatching if livepatch makes gramatical sense in the context as the Linux-wide usage of the phrase is 'livepatch' -- https://www.kernel.org/doc/html/latest/livepatch/livepatch.html

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -839,6 +845,147 @@ def do_processes_require_restart(self):
         return False
     # endregion Reboot Management
 
+    #region Livepatching
+    def start_livepatching(self):
+        """ Applies livepatches on the machine, if it's pre-req are met"""
```
kjohn-msft: nit: its pre-req

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -839,6 +845,147 @@ def do_processes_require_restart(self):
...
+        # todo: is simulation (simulate) needed? Should be clearer in UTs
```
kjohn-msft: marking todo for later

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -839,6 +845,147 @@ def do_processes_require_restart(self):
...
+                    "State": livepatch.get("State", ""),
+                    "Version": livepatch.get("Version", "")
+                })
+                break
+
+        return extracted
+
+    @staticmethod
+    def __reformat_date_for_livepatch(date_str):
+        """Converts AzGPS date format (20240401T000000Z) to ISO 8601 date string (2024-04-01T00:00:00Z)."""
```
kjohn-msft: Both are ISO 8601 date formats - one is basic and one is extended. Include in Envlayer->Datetime as generic.

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -839,6 +845,147 @@ def do_processes_require_restart(self):
...
+        extracted_livepatch_fields = self.extract_livepatch_fields(livepatch_status)
+        if len(extracted_livepatch_fields) == 0:
+            self.composite_logger.log_warning("[APM] No supported running livepatch entry found in livepatch status. Nothing to update in patch installation summary")
+            return
+
+        livepatch_fields = extracted_livepatch_fields[0]
+        check_state = livepatch_fields["CheckState"]
+        state = livepatch_fields["State"]
+        patch_name = "livepatch_" + check_state + "_" + state
+        patch_version = livepatch_fields["Version"]
```
kjohn-msft: All this needs to be documented in the spec.

kjohn-msft: Please provide a range of end-user-facing examples

src/core/src/bootstrap/Constants.py
```diff
@@ -87,6 +88,12 @@ class EulaSettings(EnumBackport):
         ACCEPTED_BY = 'AcceptedBy'
         LAST_MODIFIED = 'LastModified'
 
+    class LivePatchingSettings(EnumBackport):
+        ENABLE_LIVEPATCHING = 'EnableLivePatching'
+        LIVEPATCH_ONLY = 'LivePatchOnly'
+        ENABLED_BY = 'EnabledBy'
+        LAST_MODIFIED = 'LastModified'
```
kjohn-msft: Leave inline comments on the first two on exactly what they are supposed to mean

kjohn-msft (changes requested):
Comments inline.

src/core/src/core_logic/ExecutionConfig.py
```diff
@@ -260,9 +265,49 @@ def __is_eula_accepted_for_all_patches(self):
...
+        return livepatching_config
+
+    def __is_livepatching_enabled(self, livepatching_settings):
+        """ Determines if livepatching is enabled or disabled. """
+        is_livepatching_enabled = False
+
+        enable_livepatching = self.__fetch_specific_setting(livepatching_settings, Constants.LivePatchingSettings.ENABLE_LIVEPATCHING)
+        enabled_by = self.__fetch_specific_setting(livepatching_settings, Constants.LivePatchingSettings.ENABLED_BY)
+        last_modified = self.__fetch_specific_setting(livepatching_settings, Constants.LivePatchingSettings.LAST_MODIFIED)
+        if enable_livepatching is not None and enable_livepatching in [True, 'True', 'true', '1', 1]:
```
michellemcdaniel: what about TRUE? is there any way to do a case sensitive compare/cast to string then tolower the value before the compare?

src/core/src/core_logic/ExecutionConfig.py
```diff
@@ -260,9 +265,49 @@ def __is_eula_accepted_for_all_patches(self):
...
+                                            .format(str(enable_livepatching), str(enabled_by), str(last_modified), str(is_livepatching_enabled)))
+        else:
+            self.composite_logger.log_debug("LivePatching is not enabled for the VM. Computed value of [IsLivePatchingEnabled={0}]".format(str(is_livepatching_enabled)))
+
+        return is_livepatching_enabled
+
+    def __is_livepatch_only_enabled(self, livepatching_settings):
+        """ Determines if customer has set config to only livepatch i.e. no cold patch. """
+        """ NOTE: This is not in use currently but can be added in MVP if needed"""
+        is_livepatch_only_enabled = self.__fetch_specific_setting(livepatching_settings, Constants.LivePatchingSettings.LIVEPATCH_ONLY) in [True, 'True', 'true', '1', 1]
```
yashnap: We currently check for multiple representations of a boolean true value (True, "True", "true", "1", 1) in several places. Extracting this into a shared helper function would improve readability, reduce duplication, and make the intent clearer.

src/core/src/core_logic/ExecutionConfig.py
```diff
@@ -260,9 +265,49 @@ def __is_eula_accepted_for_all_patches(self):
...
+        if enable_livepatching is not None and enable_livepatching in [True, 'True', 'true', '1', 1]:
+            is_livepatching_enabled = True
+            self.composite_logger.log_debug("Livepatching config values read from disk: [EnableLivePatching={0}] [EnabledBy={1}] [LastModified={2}]. Computed value of [IsLivePatchingEnabled={3}]"
+                                            .format(str(enable_livepatching), str(enabled_by), str(last_modified), str(is_livepatching_enabled)))
+        else:
+            self.composite_logger.log_debug("LivePatching is not enabled for the VM. Computed value of [IsLivePatchingEnabled={0}]".format(str(is_livepatching_enabled)))
+
+        return is_livepatching_enabled
+
+    def __is_livepatch_only_enabled(self, livepatching_settings):
```
michellemcdaniel: Can we standardize function names when it comes to using "livepatch" and "livepatching"?

src/core/src/core_logic/ExecutionConfig.py
michellemcdaniel: +1 here. If you can also combine that with my comment about casting to string, tolower-ing and then you will only need to compare to 'true' and '1'

src/core/src/core_logic/ExecutionConfig.py
rane-rajasi: Added the actual value (read from file config) in the else block

src/core/src/core_logic/ExecutionConfig.py
rane-rajasi: I've changed livepatching to livepatch in all places except a few where using livepatching seemed more correct. Please review and let me know your thoughts

src/core/src/core_logic/ExecutionConfig.py
rane-rajasi: See my response: https://github.com/Azure/LinuxPatchExtension/pull/341#discussion_r3169998092

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: done

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: Done

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: Added a sample format with possible values for different fields both in code and in this PR's description: https://github.com/Azure/LinuxPatchExtension/pull/341#issue-4309506752

src/core/src/bootstrap/Constants.py
rane-rajasi: done

src/core/src/core_logic/ExecutionConfig.py
rane-rajasi: Made a code change to address this and the comment below on a helper function

src/core/src/core_logic/ExecutionConfig.py
rane-rajasi: done

src/core/src/core_logic/ExecutionConfig.py
rane-rajasi: Done in most places, refer my other comment: https://github.com/Azure/LinuxPatchExtension/pull/341#discussion_r3169998092

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -839,6 +845,196 @@ def do_processes_require_restart(self):
...
+        try:
+            code, output = self.env_layer.run_command_output(cmd, False, False)
+            if code == 0:
+                self.composite_logger.log_debug("[APM] Successfully set cutoff date in livepatch config. [ConfigSet={0}]".format(str(output)))
+                return True
+            else:
+                error_msg = "[APM] Command to set cutoff date in livepatch config failed. [Cmd={0}][Code={1}][Output={2}]".format(str(cmd), str(code), str(output))
+                self.composite_logger.log_error(error_msg)
+                self.status_handler.add_error_to_status(error_msg, Constants.PatchOperationErrorCodes.LIVEPATCH_ERROR)
+                # Q: should we disable livepatching if we fail to set cutoff date in config since it's a critical config for livepatching to work properly?
```
rane-rajasi, line 903: @kjohn-msft, please share your thoughts on this one. I'm referring to the livepatch service on a VM. Our current expectation is that the customer needs to enable/activate it. This is one of the prereqs and is enforced in ubuntu_pro_client.is_livepatch_service_enabled_on_machine() 

Livepatching is a service managed by the livepatch client, when this service is enabled, Livepatch client is responsible to apply livepatches during its scheduled run. When we set a cutoff-date, livepatch client honors it and applies only those patches that are available on or before the cutoff date. Without a cutoff-date, livepatch client will apply all patches available at the time it executes.

Ideally, we should not reset livepatch service, since we require the customer to enable it. But if this service is enabled and AzGPS is not able to set a cutoff-date, livepatching will continue on the VM at its preset cadence and we will never report any status for livepatching. This creates an information gap. 

Imo, we LPE should disable livepatching if we fail to set a cutoff date and report it as an error in status blob. Let me know your thoughts

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -421,4 +421,9 @@ def datetime_string_to_posix_time(datetime_string, format_string):
             epoch = datetime.datetime(1970, 1, 1)
             return int((dt - epoch).total_seconds())
 
+        @staticmethod
+        def datetime_iso_basic_string_to_extended_string(datetime_string):
+            """Converts ISO 8601 basic date format (20240401T000000Z) to extended format (2024-04-01T00:00:00Z)."""
```
kjohn-msft, line 427: type information missing for new code. see 413 for example

src/core/src/bootstrap/Constants.py
```diff
@@ -87,6 +88,12 @@ class EulaSettings(EnumBackport):
         ACCEPTED_BY = 'AcceptedBy'
         LAST_MODIFIED = 'LastModified'
 
+    class LivePatchSettings(EnumBackport):
+        ENABLE_LIVEPATCH = 'EnableLivePatch' # boolean config that if set to true indicates customer's ask on AzGPS to apply livepatches on their VM
+        LIVEPATCH_ONLY = 'LivePatchOnly' # boolean config that indicates if the customer is requesting only livepatches i.e. no cold patch or regular patching with reboots
```
kjohn-msft, line 93: What happens if both are true?

src/core/src/core_logic/ExecutionConfig.py
```diff
@@ -92,6 +92,11 @@ def __init__(self, env_layer, composite_logger, execution_parameters):
         # EULA config
         self.accept_package_eula = self.__is_eula_accepted_for_all_patches()
 
+        # LivePatch config
+        self.livepatch_customer_config_settings = self.__get_livepatch_customer_config_in_json()
+        self.is_livepatch_requested = self.__is_livepatch_requested(self.livepatch_customer_config_settings)
+        self.is_livepatch_only_requested = self.__is_livepatch_only_requested(self.livepatch_customer_config_settings)
```
kjohn-msft, line 98: Same as in the config section - what happens if both are true

src/core/src/core_logic/ExecutionConfig.py
```diff
@@ -260,10 +265,68 @@ def __is_eula_accepted_for_all_patches(self):
...
+                self.composite_logger.log_debug("Livepatch customer config values from disk: [Config={0}]".format(str(livepatch_customer_config)))
+            else:
+                self.composite_logger.log_debug("No livepatch customer config found on the VM. Returning empty config.")
+        except Exception as error:
+            self.composite_logger.log_debug("Error occurred while reading and parsing livepatch customer config. Returning empty config. Error=[{0}]".format(repr(error)))
+
+        return livepatch_customer_config
+
+    def __is_livepatch_requested(self, livepatch_settings):
+        """ Determines if livepatch is requested in config settings. Returns a boolean."""
```
kjohn-msft, line 286: All new functions require type information (in and out)

src/core/src/core_logic/ExecutionConfig.py
```diff
@@ -260,10 +265,68 @@ def __is_eula_accepted_for_all_patches(self):
...
+        self.composite_logger.log_debug("Livepatch only config values from disk: [LivePatchOnly={0}]. Computed value of [LivePatchOnlyRequested={1}]"
+                                        .format(str(self.__fetch_specific_setting(livepatch_settings, Constants.LivePatchSettings.LIVEPATCH_ONLY)),
+                                                str(livepatch_only_requested)))
+        return livepatch_only_requested
+
     @staticmethod
-    def __fetch_specific_eula_setting(settings_source, setting_to_fetch):
-        """ Returns the specific setting value from eula_settings_source or None if not found """
+    def __fetch_specific_setting(settings_source, setting_to_fetch):
+        """ Returns the specific setting value from the given settings_source or None if not found """
```
kjohn-msft, line 315: Function type info here and elsewhere

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -839,6 +845,196 @@ def do_processes_require_restart(self):
         return False
     # endregion Reboot Management
 
+    #region Livepatch
+    def start_livepatch(self):
+        """ Applies livepatches on the machine, if its pre-req are met"""
+        if self.are_livepatch_prereq_met():
+            self.start_livepatching_on_machine()
+        else:
+            self.composite_logger.log_warning("[APM] Livepatches are not applied since the pre-requisites were not met")
```
kjohn-msft, line 855: Log is unclear on what was not met and how to remediate.

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -839,6 +845,196 @@ def do_processes_require_restart(self):
...
+
+    def are_livepatch_prereq_met(self):
+        """ Validates whether livepatch prereqs are met.
+        These pre-reqs are: Machine should be attached to a pro subscription and livepatch service should be enabled on the VM. """
+        self.composite_logger.log_debug("[APM] Checking if all the pre-reqs to receive livepatches are met. NOTE: Livepatches is only available on Ubuntu LTS paid pro VMs and has to be in enabled state")
+        if not self.ubuntu_pro_client.is_livepatching_applicable_for_machine():
+            error_message = "[APM] Livepatching is not applicable for this machine, hence no livepatches will be installed"
```
kjohn-msft: It's unclear from a customer perspective what this actually means. This is superficially tied to an internal definition of 'applicable'.

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -839,6 +845,196 @@ def do_processes_require_restart(self):
...
+            self.composite_logger.log_error(error_message)
+            self.status_handler.add_error_to_status(error_message, Constants.PatchOperationErrorCodes.LIVEPATCH_ERROR)
+            return False
+
+        if not self.ubuntu_pro_client.is_livepatch_service_enabled_on_machine():
+            error_message = ("[APM] Livepatch service is not enabled on this machine, hence no livepatches will be installed."
```
kjohn-msft: The Ubuntu Pro client reported that the Livepatch service is not enabled. Please enable it for Livepatching to succeed.

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -839,6 +845,196 @@ def do_processes_require_restart(self):
...
+        return True
+
+    def start_livepatching_on_machine(self):
+        """Starts livepatching on the machine according to the configurations set in AzGPS and updates livepatch status in status blob"""
+        self.composite_logger.log_debug("[APM] Starting livepatching on the machine...")
+        if self.try_set_livepatch_cutoff_date_in_config():
+            is_livepatch_client_launch_successful = self.launch_livepatch_client()
+            if not is_livepatch_client_launch_successful:
+                self.composite_logger.log_warning("[APM] A stale livepatch status may be reported since a manual launch/restart of the livepatch client failed")
+            self.fetch_and_update_livepatch_status_in_status_blob()
```
kjohn-msft, line 885: Why are we reporting a stale status? (at all)

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -839,6 +845,196 @@ def do_processes_require_restart(self):
...
+        else:
+            self.composite_logger.log_warning("[APM] AzGPS will not apply livepatch on the VM since the livepatch cutoff date was not set. "
+                                            "Please check previous logs for more details on why it failed and fix the issue before trying to apply livepatches again")
```
kjohn-msft: This isn't readily actionable to the customer (or internally). Who didn't set the livepatch cutoff date?

Try to consolidate messages into fewer items but be clear on: what is wrong, and what to do. And leave comments inline on the expected outcome when someone gets this message.

src/core/src/package_managers/AptitudePackageManager.py
kjohn-msft: This isn't a code question - this is a design-time workflow question. It's good to have a flow chart shared internally so that there's consensus on what the behaviors in various states are supposed to be. It'll help with faster turnaround in code.

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -839,6 +845,196 @@ def do_processes_require_restart(self):
...
+            else:
+                error_msg = "[APM] Command to set cutoff date in livepatch config failed. [Cmd={0}][Code={1}][Output={2}]".format(str(cmd), str(code), str(output))
+                self.composite_logger.log_error(error_msg)
+                self.status_handler.add_error_to_status(error_msg, Constants.PatchOperationErrorCodes.LIVEPATCH_ERROR)
+                # Q: should we disable livepatching if we fail to set cutoff date in config since it's a critical config for livepatching to work properly?
+        except Exception as error:
+            error_msg = "[APM] Livepatch config update Exception: [Exception={0}]".format(repr(error))
+            self.composite_logger.log_error(error_msg)
+            self.status_handler.add_error_to_status(error_msg, Constants.PatchOperationErrorCodes.LIVEPATCH_ERROR)
+            # Q: should we disable livepatching if we fail to set cutoff date in config since it's a critical config for livepatching to work properly?
```
kjohn-msft, line 908: Try to dig deeper into this question to arrive at actual proposals - what happens if you disable livepatching? What is the risk to the customer between the choices available? etc

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -839,6 +845,196 @@ def do_processes_require_restart(self):
...
+        return False
+
+    def launch_livepatch_client(self):
+        """ Launch livepatch client manually as best case effort to ensure livepatches are applied in a timely manner.
+        If this fails, livepatches will still be applied but it will be up to the machine's cron to trigger it"""
```
kjohn-msft, line 914: "it will be up to the machine's cron to trigger it"
- In isolation, this is a yellow flag suggesting that the workflow needs to be reviewed closer separate from the code. What are guaranteed outcomes in the flow? What are bad outcomes? In the case of bad outcomes, what are the mitigation steps? When they fail, what are the clear messages given to the customer + actionable guidance given?   [whole workflow]

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -839,6 +845,196 @@ def do_processes_require_restart(self):
...
+                launch_successful = True
+            else:
+                self.composite_logger.log_warning("[APM] Failed to launch livepatch client. [Output={0}]".format(str(output)))
+        except Exception as error:
+            livepatch_launch_exception = repr(error)
+            self.composite_logger.log_warning("[APM] Exception while launching livepatch client. [Exception={0}]".format(livepatch_launch_exception))
+        return launch_successful
+
+    def fetch_and_update_livepatch_status_in_status_blob(self):
+        """Fetches livepatch status and if a livepatch/es is/are applied, updates it as a new patch entry in PatchInstallationSummary"""
```
kjohn-msft, line 931: It's unclear why this is in patchinstallationsummary (only?). What is the assessment-side story?

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -839,6 +845,196 @@ def do_processes_require_restart(self):
...
+        cmd = self.get_livepatch_status_cmd
+        try:
+            code, output = self.env_layer.run_command_output(cmd, False, False)
+            if code == 0:
+                livepatch_status = json.loads(output)
+                self.composite_logger.log_debug("[APM] Successfully fetched livepatch status. [Status={0}]".format(str(livepatch_status)))
+            else:
+                error_msg = "[APM] Failed to fetch livepatch status. [Cmd={0}][Code={1}][Output={2}]".format(str(cmd), str(code), str(output))
+                self.composite_logger.log_error(error_msg)
+                self.status_handler.add_error_to_status(error_msg, Constants.PatchOperationErrorCodes.LIVEPATCH_ERROR)
```
kjohn-msft: Isn't this redundant over a controlled throw into the exception block with a message of your choice when the code is non-zero?

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -839,6 +845,196 @@ def do_processes_require_restart(self):
...
+
+        livepatch_fields = extracted_livepatch_fields[0]
+        check_state = str(livepatch_fields["CheckState"])
+        state = str(livepatch_fields["State"])
+        patch_name = "livepatch_" + check_state + "_" + state
+        patch_version = str(livepatch_fields["Version"])
+
+        patch_status = Constants.NOT_SELECTED
+        if state.lower() == "applied":
+            patch_status = Constants.INSTALLED
```
kjohn-msft, line 973: These are the only 2 states?

For the previously discussed workflow diagram, can you map what goes into each status write in various branches of the flow?

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -839,6 +845,196 @@ def do_processes_require_restart(self):
...
+                extracted.append({
+                    "CheckState": livepatch.get("CheckState", ""),
+                    "State": livepatch.get("State", ""),
+                    "Version": livepatch.get("Version", "")
+                })
+                break
+
+        return extracted
+
+    def __reformat_date_for_livepatch(self, date_str):
```
kjohn-msft: This is not a try_reformat   -- to silently fail and return empty string

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -839,6 +845,196 @@ def do_processes_require_restart(self):
...
+                "Kernel-Package-Fixes": [], // list of all kernel packages fixed
+                "Installed-Kernels": [],
+                "Patched-CVEs": [], // list of patched CVEs identifiers
+                "Digest": ""
+            },
+            "Blocking-Options": [ // List of configs blocking livepatch, if any. For eg: cutoff-date set for livepatch client
+                "cutoff-date"
+            ],
+            "Using-Cutoff-Date": <bool> // boolean value indicating whether livepatch client is using cutoff-date config or not
+        } """
```
kjohn-msft: Dump into a new suitably named file at src\tools\references
And reference that file here. We shouldn't bloat code where avoidable.

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -839,6 +845,196 @@ def do_processes_require_restart(self):
...
+
+        for status_item in livepatch_status.get("Status", []):
+            if status_item.get("Running", False) == True and status_item.get("Supported", "unsupported").lower() == "supported":
+                livepatch = status_item.get("Livepatch", {})
+                extracted.append({
+                    "CheckState": livepatch.get("CheckState", ""),
+                    "State": livepatch.get("State", ""),
+                    "Version": livepatch.get("Version", "")
+                })
+                break
```
kjohn-msft, line 990: Is the entirety of the livepatch status written to logs at any point?

src/core/src/package_managers/UbuntuProClient.py
```diff
@@ -172,3 +173,35 @@ def is_reboot_pending(self):
 
         self.composite_logger.log_debug("[APM][Pro] Ubuntu Pro Client Reboot Required: [UbuntuProClientSuccess={0}][RebootRequiredFlag={1}][Error={2}]".format(ubuntu_pro_client_api_success, ubuntu_pro_client_reboot_required, ubuntu_pro_client_exception))
         return ubuntu_pro_client_api_success, ubuntu_pro_client_reboot_required
+
+    # region Livepatch
+    def is_livepatching_applicable_for_machine(self):
+        """ Verifies if livepatching is applicable for the machine by checking if the machine is an Ubuntu LTS Pro VM """
```
kjohn-msft: Capable or supported is a lot more accurate than applicable in this context.

vm_supports_livepatching    for e.g. is a lot more precise

if vm_supports_livepatching() :     // consider the clarity here

or 
pro_client_attached_for_livepatching

if pro_client_attached_for_livepatching():    // all the code is easier to read - it's precise and crisp without having to go to the function definition

And use this for the message on 181 too (which + 180 is the most useful part that's helped so far)

src/core/src/package_managers/UbuntuProClient.py
```diff
@@ -172,3 +173,35 @@ def is_reboot_pending(self):
...
+        if not self.is_ubuntu_pro_client_attached:
+            self.composite_logger.log_warning("[APM][Pro] Livepatching is not applicable for the machine since it is not an Ubuntu Pro VM.")
+            return False
+
+        return True
+
+    def is_livepatch_service_enabled_on_machine(self):
+        """ Verifies if livepatch service is enabled on the machine """
```
kjohn-msft, line 189: is_ is implicit

Separately: all functions that are new need to have type data

src/core/src/package_managers/UbuntuProClient.py
```diff
@@ -172,3 +173,35 @@ def is_reboot_pending(self):
...
+                if livepatch_status.lower() == 'enabled' or livepatch_status.lower() == 'warning':
+                    self.composite_logger.log_debug("[APM][Pro] Livepatch service is enabled on the machine.")
+                    livepatch_service_enabled = True
+                else:
+                    self.composite_logger.log_warning("[APM][Pro] Livepatch service is NOT enabled on the machine.")
+        except Exception as error:
+            ubuntu_pro_client_exception = repr(error)
+            self.composite_logger.log_debug("[APM][Pro] Ubuntu Pro Client status Exception: [Exception={0}]".format(ubuntu_pro_client_exception))
+            self.composite_logger.log_warning("[APM][Pro] Failed to determine if livepatch service is enabled on the machine due to error while querying Ubuntu Pro Client status.")
+        return livepatch_service_enabled
```
kjohn-msft, line 207: It's not known in the exception case whether it's actually true or false. And the code returns false.

This is a detail that is hidden from the caller (and is implicitly tied together today).

src/core/src/core_logic/ExecutionConfig.py
kjohn-msft: Why do we have this code here if it's not consumed?

kjohn-msft (commented):
Please share a flow chart of what this code is expected to do offline (especially for each consequential failure state - the message and the remediation recommended at each failure leaf).

Other comments inline

src/core/src/core_logic/PatchInstaller.py
```diff
@@ -58,14 +58,17 @@ def start_installation(self, simulate=False):
...
         maintenance_window = self.maintenance_window
         package_manager = self.package_manager
         reboot_manager = self.reboot_manager
 
+        if self.execution_config.is_livepatch_requested:
+            package_manager.start_livepatch()
+
+        self.composite_logger.log("\nStarting patch installation... [MachineId: " + self.env_layer.platform.vm_name() + "][ActivityId: " + self.execution_config.activity_id + "][StartTime: " + self.execution_config.start_time + "][MaintenanceWindowDuration: " + self.execution_config.duration + "]")
+
+        self.stopwatch.start()
```
nikhim-um: should the stopwatch start before start_livepatch()?

src/core/src/bootstrap/Constants.py
rane-rajasi: @kjohn-msft If both are true, LPE should perform only livepatching i.e. no regular patch installation. 

We had discussed the possibility of having this in MVP, back when we discussed MVP implementation. I've added only the config read for now, not using or applyiing it anywhere. Let me know if livepatch_only code flow should be added in MVP or to remove this section

src/core/src/core_logic/ExecutionConfig.py
rane-rajasi: See: https://github.com/Azure/LinuxPatchExtension/pull/341#discussion_r3226066810

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: Livepatch status depends on when livepatching client last ran, so a stale status could still contain livepatch details, if the client ran before LPE could explicitly launch it. LPE explicitly launching livepatch client in the previous step is a best effort attempt to ensure livepatches have been attempted at least once on the VM

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: This follows both our initial sync on MVP code flow/design and also how the livepatch client works. When a livepatch service is enabled and the livepatch client runs, it simply identifies and applies the patches in the same run. There is no way of only assessing which livepatches are available. Hence the logic and status report only in patchinstallationsummary

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: Following the reasoning here: https://github.com/Azure/LinuxPatchExtension/pull/341#discussion_r3226552295

Livepatches are auto applied whenever found, there's no only assess scenario for livepatching. So, livepatches are either available and applied or not available. NOT_SELECTED was the closest status to denote patches are not available.

This will change with the complete Livepatch design, since it will report livepatchsummary as a separate substatus, outside of patchinstallationsummary. We'll have more control over what to fetch from livepatch client and report in LPE

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: Yes, in try_get_livepatch_status() on success, i.e. when the cmd to get status returns code=0

https://github.com/Azure/LinuxPatchExtension/pull/341/changes#diff-c2a1213b3b9833583763d2503ea3d68d2de57c249db6223c532878114274621eR940

image

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: Will share the flow chart in the PR review request

src/core/src/bootstrap/EnvLayer.py
rane-rajasi: Done

src/core/src/core_logic/ExecutionConfig.py
rane-rajasi: Done

src/core/src/core_logic/ExecutionConfig.py
rane-rajasi: Done

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: Which pre-reqs were not met and what to do, if any, is mentioned in are_livepatch_prereq_met() wherever they fail

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: Modified

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: Changed

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: Modified

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: changed

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: Done

src/core/src/package_managers/UbuntuProClient.py
rane-rajasi: Renamed to pro_client_attached_for_livepatching() and updated logs whereever applicable

src/core/src/package_managers/UbuntuProClient.py
rane-rajasi: Done

src/core/src/package_managers/UbuntuProClient.py
rane-rajasi: Appended the log with "____ AzGPS will consider the service to be disabled."

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: Modified the log

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: Livepatch client (once livepatch service is enabled) runs at its preset schedule, will fetch and apply all available livepatches during its run. 

LPE by setting config date is ensuring only a certain set of patches are applied (available on or before config date). If we fail to set config date, livepatch client will continue to run and apply all patches available to the VM. This is not the livepatching with strict SDP scenario that LPE is committing to. Hence, it makes sense for LPE to disable livepatch service and report an error, so no new livepatches will be applied by the livepatch client.

However, since we require customer to enable livepatch service, LPE disabling it blurs the line between service and customer roles. LPE could simply go with the current implementation of not interfering with any customer action and simply reporting an error to notify the customer of this scenario

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: I think this covers everything: https://github.com/Azure/LinuxPatchExtension/pull/341#discussion_r3227429809

PR #344, opened 2026-05-12
Safety net: Explicitly block Azure Linux 4 and RHEL 10 until LPE support is available

yashnap:
- Detect Azure Linux 4.x and RHEL 10.x distributions and enforce a safety‑net exception until LPE dnf5 support is fully implemented and validated.

**Linux4 VM ARM ID** : [/subscriptions/6acc8a91-e2b0-4041-a069-c2932ab42fd9/resourceGroups/rg-yashna-linux4-test-wus2/providers/Microsoft.Compute/virtualMachines/vm-yashna-linux4  ](url)

LogFolder : /tmp/linux-patch/linux4_2/log

linux4_error

**Linux 3 ARM ID** :[ /subscriptions/6acc8a91-e2b0-4041-a069-c2932ab42fd9/resourceGroups/rg-linux3-tdnf/providers/Microsoft.Compute/virtualMachines/vm-linux3-tdnf-yashna](url)

LogFolder:  /tmp/linux-patch/linux4_2/log

workinglinux3

**Red Hat Arm ID** : /[subscriptions/6acc8a91-e2b0-4041-a069-c2932ab42fd9/resourceGroups/rg-rhel_10_yashna/providers/Microsoft.Compute/virtualMachines/vm-rhel-10-yashna](url)

LogFolder : /tmp/linux-patch/linux4_1/log
rhel_10_error_logs

### DIFF
--- a/src/core/src/bootstrap/Constants.py
+++ b/src/core/src/bootstrap/Constants.py
@@ -197,7 +197,7 @@ class Constants(object):
     RED_HAT = 'Red Hat'
     SUSE = 'SUSE'
     CENTOS = 'CentOS'
-    AZURE_LINUX = ['Microsoft Azure Linux', 'Common Base Linux Mariner']
+    AZURE_LINUX = ['Azure Linux', 'Microsoft Azure Linux', 'Common Base Linux Mariner']
 
     # Package Managers
     APT = 'apt'
--- a/src/core/src/bootstrap/EnvLayer.py
+++ b/src/core/src/bootstrap/EnvLayer.py
@@ -49,7 +49,7 @@ class EnvLayer(object):
 
     def is_distro_azure_linux_3_or_beyond(self):
         # type: () -> bool
-        """ Checks if the current distro is Azure Linux 3 """
+        """ Checks if the current distro is Azure Linux 3 or beyond """
         if self.is_distro_azure_linux(self.platform.linux_distribution()):
             version = distro.os_release_attr('version')
             major = version.split('.')[0] if version else None
@@ -62,7 +62,19 @@ class EnvLayer(object):
         if platform.system() == 'Windows':
             return Constants.APT
 
+        version = distro.os_release_attr('version')
+        major = version.split('.')[0] if version else None
+        os_id = distro.id()
+
+        # Check for Azure Linux
         if self.is_distro_azure_linux(str(self.platform.linux_distribution())):
+            # Azure Linux 4 not yet supported
+            if major is not None and int(major) >= 4:
+                error_msg = "Azure Linux 4 is not yet supported in your region. Please review aka.ms/LinuxPatchExtension for more information."
+                print("Error: {0}".format(error_msg))
+                raise Exception(error_msg)
+
+            # Azure Linux 3 and below use TDNF
             code, out = self.run_command_output('which tdnf', False, False)
             if code == 0:
                 return Constants.TDNF
@@ -70,7 +82,13 @@ class EnvLayer(object):
                 print("Error: Expected package manager tdnf not found on this Azure Linux VM.")
                 return str()
 
-        # choose default package manager
+        # Check for RHEL 10 (not yet supported)
+        if os_id == "rhel" and major is not None and int(major) >= 10:
+            error_msg = "RHEL 10 is not yet supported in your region. Please review aka.ms/LinuxPatchExtension for more information."
+            print("Error: {0}".format(error_msg))
+            raise Exception(error_msg)
+
+        # Choose default package manager
         package_manager_map = (('apt-get', Constants.APT),
                                ('yum', Constants.YUM),
                                ('zypper', Constants.ZYPPER))
--- a/src/core/src/bootstrap/EnvLayer.py
+++ b/src/core/src/bootstrap/EnvLayer.py
@@ -93,7 +93,7 @@ class EnvLayer(object):
 
         # Check for unsupported distros
         if self.is_distro_azure_linux_4(os_name) or self.is_distro_rhel_10(os_name):
-            error_msg = "This distro is not yet supported in your region. Please review https://aka.ms/VMGuestPatchingCompatibility for more information. [Distro={0}]".format(str(os_name))
+            error_msg = "This distro is not yet supported in your region. Please review https://aka.ms/VMGuestPatchingCompatibility for more information. [Distro={0}][Version={1}][Code={2}]".format(str(os_name), os_version, os_code)
             print("Error: {0}".format(error_msg))
             return str()
 
--- a/src/core/src/bootstrap/EnvLayer.py
+++ b/src/core/src/bootstrap/EnvLayer.py
@@ -50,12 +50,10 @@ class EnvLayer(object):
     @staticmethod
     def __try_get_major_version():
         # type: () -> str or None
-        """ Attempts to get major version from distro.os_release_attr('version').
-        Returns major version as string or None if not available. """
+        """ Attempts to get major version from distro.os_release_attr('version').Returns major version as string or None if not available. """
         version = distro.os_release_attr('version')
         major = version.split('.')[0] if version else None
         return major
-
 
     def __is_matching_distro_and_version(self, distro_name, distro_to_match, version_to_match):
         # type: (str, str, int) -> bool
@@ -108,8 +106,7 @@ class EnvLayer(object):
                 print("Error: Expected package manager tdnf not found on this Azure Linux VM.")
                 return str()
 
-
-        # Choose default package manager
+        # choose default package manager
         package_manager_map = (('apt-get', Constants.APT),
                                ('yum', Constants.YUM),
                                ('zypper', Constants.ZYPPER))
--- a/src/core/src/bootstrap/EnvLayer.py
+++ b/src/core/src/bootstrap/EnvLayer.py
@@ -47,13 +47,21 @@ class EnvLayer(object):
     def is_distro_azure_linux(distro_name):
         return any(x in distro_name for x in Constants.AZURE_LINUX)
 
+    @staticmethod
+    def _get_major_version():
+        # type: () -> str
+        """ Extracts major version from distro.os_release_attr('version').
+        Returns major version as string or None if not available. """
+        version = distro.os_release_attr('version')
+        major = version.split('.')[0] if version else None
+        return major
+
     def is_distro_azure_linux_3_or_beyond(self):
         # type: () -> bool
         """ Checks if the current distro is Azure Linux 3 """
         if self.is_distro_azure_linux(self.platform.linux_distribution()):
-            version = distro.os_release_attr('version')
-            major = version.split('.')[0] if version else None
-            return major is not None and int(major) >= 3
+            major = self._get_major_version()
+            return major is not None and int(major) == 3
         return False
 
     def get_package_manager(self):
@@ -62,31 +70,20 @@ class EnvLayer(object):
         if platform.system() == 'Windows':
             return Constants.APT
 
-        version = distro.os_release_attr('version')
-        major = version.split('.')[0] if version else None
+        major = self._get_major_version()
         os_id = distro.id()
 
         # Check for Azure Linux
         if self.is_distro_azure_linux(str(self.platform.linux_distribution())):
-            # Azure Linux 4 not yet supported
-            if major is not None and int(major) == 4:
-                error_msg = "Azure Linux 4 is not yet supported in your region. Please review aka.ms/LinuxPatchExtension for more information."
-                print("Error: {0}".format(error_msg))
-                raise Exception(error_msg)
+            return self._get_azure_linux_package_manager(major)
 
-            # Azure Linux 3 and below use TDNF
-            code, out = self.run_command_output('which tdnf', False, False)
-            if code == 0:
-                return Constants.TDNF
-            else:
-                print("Error: Expected package manager tdnf not found on this Azure Linux VM.")
-                return str()
-
-        # Check for RHEL 10 (not yet supported)
-        if os_id == "rhel" and major is not None and int(major) == 10:
-            error_msg = "RHEL 10 is not yet supported in your region. Please review aka.ms/LinuxPatchExtension for more information."
-            print("Error: {0}".format(error_msg))
-            raise Exception(error_msg)
+        # Check for RHEL
+        if os_id == "rhel":
+            package_manager = self._get_rhel_package_manager(major)
+            # Helper will raise exception for unsupported versions (e.g., RHEL 10).
+            # If it returns a value, use it; otherwise fall through to default detect
+            if package_manager is not None:
+                return package_manager
 
         # Choose default package manager
         package_manager_map = (('apt-get', Constants.APT),
@@ -98,6 +95,38 @@ class EnvLayer(object):
                 return entry[1]
 
         return str()
+
+    def _get_azure_linux_package_manager(self, major):
+        # type: (str) -> str
+        """ Determines the package manager for Azure Linux based on version.
+        Azure Linux 3 and below use TDNF. Azure Linux 4 use DNF (not yet supported).
+        Logs error message if version is unsupported. """
+        # Azure Linux 4 not yet supported
+        if major is not None and int(major) == 4:
+            error_msg = "Azure Linux 4 is not yet supported in your region. Please review https://aka.ms/VMGuestPatchingCompatibility for more information."
+            print("Error: {0}".format(error_msg))
+            return str()
+
+        # Azure Linux 3 and below use TDNF
+        code, out = self.run_command_output('which tdnf', False, False)
+        if code == 0:
+            return Constants.TDNF
+        else:
+            print("Error: Expected package manager tdnf not found on this Azure Linux VM.")
+            return str()
+
+    def _get_rhel_package_manager(self, major):
+        # type: (str) -> str
+        """ Determines the package manager for RHEL based on version.
+        Returns DNF when RHEL 10+ is supported. Logs error message if unsupported. """
+        # RHEL 10 not yet supported
+        if major is not None and int(major) == 10:
+            error_msg = "RHEL 10 is not yet supported in your region. Please review https://aka.ms/VMGuestPatchingCompatibility for more information."
+            print("Error: {0}".format(error_msg))
+            return str()
+
+        # return None to allow default package manager detection
+        return None
 
     def set_env_var(self, var_name, var_value=str(), raise_if_not_success=False):
         # type: (str, str, bool) -> None
--- a/src/core/src/bootstrap/EnvLayer.py
+++ b/src/core/src/bootstrap/EnvLayer.py
@@ -49,7 +49,7 @@ class EnvLayer(object):
 
     def is_distro_azure_linux_3_or_beyond(self):
         # type: () -> bool
-        """ Checks if the current distro is Azure Linux 3 or beyond """
+        """ Checks if the current distro is Azure Linux 3 """
         if self.is_distro_azure_linux(self.platform.linux_distribution()):
             version = distro.os_release_attr('version')
             major = version.split('.')[0] if version else None
@@ -69,7 +69,7 @@ class EnvLayer(object):
         # Check for Azure Linux
         if self.is_distro_azure_linux(str(self.platform.linux_distribution())):
             # Azure Linux 4 not yet supported
-            if major is not None and int(major) >= 4:
+            if major is not None and int(major) == 4:
                 error_msg = "Azure Linux 4 is not yet supported in your region. Please review aka.ms/LinuxPatchExtension for more information."
                 print("Error: {0}".format(error_msg))
                 raise Exception(error_msg)
@@ -83,7 +83,7 @@ class EnvLayer(object):
                 return str()
 
         # Check for RHEL 10 (not yet supported)
-        if os_id == "rhel" and major is not None and int(major) >= 10:
+        if os_id == "rhel" and major is not None and int(major) == 10:
             error_msg = "RHEL 10 is not yet supported in your region. Please review aka.ms/LinuxPatchExtension for more information."
             print("Error: {0}".format(error_msg))
             raise Exception(error_msg)
--- a/src/core/src/bootstrap/EnvLayer.py
+++ b/src/core/src/bootstrap/EnvLayer.py
@@ -48,21 +48,40 @@ class EnvLayer(object):
         return any(x in distro_name for x in Constants.AZURE_LINUX)
 
     @staticmethod
-    def _get_major_version():
-        # type: () -> str
-        """ Extracts major version from distro.os_release_attr('version').
+    def __try_get_major_version():
+        # type: () -> str or None
+        """ Attempts to get major version from distro.os_release_attr('version').
         Returns major version as string or None if not available. """
         version = distro.os_release_attr('version')
         major = version.split('.')[0] if version else None
         return major
 
-    def is_distro_azure_linux_3_or_beyond(self):
-        # type: () -> bool
+
+    def __is_matching_distro_and_version(self, distro_name, distro_to_match, version_to_match):
+        # type: (str, str, int) -> bool
+        """ Checks if distro name matches and version matches the expected version.
+        Returns True if both distro and version match, False otherwise. """
+        if distro_to_match == Constants.AZURE_LINUX and not self.is_distro_azure_linux(str(distro_name)):
+            return False
+        elif distro_to_match == Constants.RED_HAT and not Constants.RED_HAT.lower() in str(distro_name).lower():
+            return False
+        major = self.__try_get_major_version()
+        return major is not None and int(major) == version_to_match
+
+    def is_distro_azure_linux_3(self, distro_name):
+        # type: (str) -> bool
         """ Checks if the current distro is Azure Linux 3 """
-        if self.is_distro_azure_linux(self.platform.linux_distribution()):
-            major = self._get_major_version()
-            return major is not None and int(major) == 3
-        return False
+        return self.__is_matching_distro_and_version(distro_name, Constants.AZURE_LINUX, version_to_match=3)
+
+    def is_distro_azure_linux_4(self, distro_name):
+        # type: (str) -> bool
+        """ Checks if the current distro is Azure Linux 4 """
+        return self.__is_matching_distro_and_version(distro_name, Constants.AZURE_LINUX, version_to_match=4)
+
+    def is_distro_rhel_10(self, distro_name):
+        # type: (str) -> bool
+        """ Checks if the current distro is RHEL 10 """
+        return self.__is_matching_distro_and_version(distro_name, Constants.RED_HAT, version_to_match=10)
 
     def get_package_manager(self):
         # type: () -> str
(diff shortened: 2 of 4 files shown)


src/core/src/bootstrap/EnvLayer.py
```diff
@@ -62,15 +62,33 @@ def get_package_manager(self):
         if platform.system() == 'Windows':
             return Constants.APT
 
+        version = distro.os_release_attr('version')
+        major = version.split('.')[0] if version else None
+        os_id = distro.id()
+
+        # Check for Azure Linux
         if self.is_distro_azure_linux(str(self.platform.linux_distribution())):
```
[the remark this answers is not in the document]
yashnap: Same code as before. Not changing for the purpose of this PR.

src/core/tests/Test_EnvLayer.py
```diff
@@ -71,6 +71,18 @@ def mock_distro_os_release_attr_return_azure_linux_2(self, attribute):
...
+        return '4.0.2'
+
+    def mock_distro_os_release_attr_return_rhel_10(self, attribute):
+        return '10.0'
+
+    def mock_linux_distribution_to_return_azure_linux_4(self):
+        return ['Microsoft Azure Linux', '4.0', '']
+
+    def mock_distro_id_return_rhel(self):
+        return 'rhel'
```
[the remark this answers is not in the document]
yashnap: Done

src/core/tests/Test_EnvLayer.py
```diff
@@ -139,6 +151,43 @@ def test_platform(self):
...
+
+        with self.assertRaises(Exception) as context:
+            self.envlayer.get_package_manager()
+        self.assertEqual(str(context.exception), "RHEL 10 is not yet supported in your region. Please review aka.ms/LinuxPatchExtension for more information.")
+
+        # restore
+        distro.id = self.backup_distro_id
+        distro.os_release_attr = self.backup_distro_os_release_attr
+        self.envlayer.platform.linux_distribution = self.backup_linux_distribution
+
```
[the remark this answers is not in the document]
yashnap: Added

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -62,15 +62,33 @@ def get_package_manager(self):
         if platform.system() == 'Windows':
             return Constants.APT
 
+        version = distro.os_release_attr('version')
+        major = version.split('.')[0] if version else None
+        os_id = distro.id()
```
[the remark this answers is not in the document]
yashnap: Same code logic and method usage in current master branch. Not updating for this PR

src/core/tests/Test_EnvLayer.py
```diff
@@ -79,28 +91,32 @@ def test_get_package_manager(self):
...
+            [lambda cmd, no_output, chk_err: (-1, ''), self.mock_linux_distribution, self.mock_distro_os_release_attr_return_none, str()],    # no matches for any package manager
         ]
 
         for row in test_input_output_table:
             self.envlayer.run_command_output = row[0]
             self.envlayer.platform.linux_distribution = row[1]
+            distro.os_release_attr = row[2]
             package_manager = self.envlayer.get_package_manager()
-            self.assertTrue(package_manager is row[2])
+            self.assertTrue(package_manager is row[3])
```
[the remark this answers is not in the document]
yashnap: Updated the index. Same code as before. Ignoring for purpose of this PR

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -62,15 +62,33 @@ def get_package_manager(self):
...
+
+        # Check for Azure Linux
         if self.is_distro_azure_linux(str(self.platform.linux_distribution())):
+            # Azure Linux 4 not yet supported
+            if major is not None and int(major) >= 4:
```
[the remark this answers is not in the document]
yashnap: done

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -62,15 +62,33 @@ def get_package_manager(self):
...
             code, out = self.run_command_output('which tdnf', False, False)
             if code == 0:
                 return Constants.TDNF
             else:
                 print("Error: Expected package manager tdnf not found on this Azure Linux VM.")
                 return str()
 
-        # choose default package manager
+        # Check for RHEL 10 (not yet supported)
+        if os_id == "rhel" and major is not None and int(major) >= 10:
```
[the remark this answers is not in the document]
yashnap: Done

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -62,15 +62,33 @@ def get_package_manager(self):
...
                 return str()
 
-        # choose default package manager
+        # Check for RHEL 10 (not yet supported)
+        if os_id == "rhel" and major is not None and int(major) == 10:
+            error_msg = "RHEL 10 is not yet supported in your region. Please review aka.ms/LinuxPatchExtension for more information."
+            print("Error: {0}".format(error_msg))
+            raise Exception(error_msg)
+
+        # Choose default package manager
```
kjohn-msft: Functionally correct but doesn't follow 'Single Responsibility' in SOLID principles.

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -62,15 +62,33 @@ def get_package_manager(self):
...
 
+        version = distro.os_release_attr('version')
+        major = version.split('.')[0] if version else None
+        os_id = distro.id()
+
+        # Check for Azure Linux
         if self.is_distro_azure_linux(str(self.platform.linux_distribution())):
+            # Azure Linux 4 not yet supported
+            if major is not None and int(major) == 4:
+                error_msg = "Azure Linux 4 is not yet supported in your region. Please review aka.ms/LinuxPatchExtension for more information."
```
kjohn-msft: https://aka.ms/VMGuestPatchingCompatibility

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -62,15 +62,33 @@ def get_package_manager(self):
...
             if code == 0:
                 return Constants.TDNF
             else:
                 print("Error: Expected package manager tdnf not found on this Azure Linux VM.")
                 return str()
 
-        # choose default package manager
+        # Check for RHEL 10 (not yet supported)
+        if os_id == "rhel" and major is not None and int(major) == 10:
+            error_msg = "RHEL 10 is not yet supported in your region. Please review aka.ms/LinuxPatchExtension for more information."
```
kjohn-msft: https://aka.ms/VMGuestPatchingCompatibility

kjohn-msft (changes requested):
@rane-rajasi to review deeper with context of comments left.

src/core/tests/Test_EnvLayer.py
```diff
@@ -79,28 +91,32 @@ def test_get_package_manager(self):
...
 
         test_input_output_table = [
-            [self.mock_run_command_for_apt, self.mock_linux_distribution, Constants.APT],
-            [self.mock_run_command_for_tdnf, self.mock_linux_distribution_to_return_azure_linux_3, Constants.TDNF],
-            [self.mock_run_command_for_yum, self.mock_linux_distribution_to_return_azure_linux_3, str()],  # check for Azure Linux machine which does not have tdnf
-            [self.mock_run_command_for_tdnf, self.mock_linux_distribution_to_return_azure_linux_2, Constants.TDNF],
-            [self.mock_run_command_for_yum, self.mock_linux_distribution, Constants.YUM],
-            [self.mock_run_command_for_zypper, self.mock_linux_distribution, Constants.ZYPPER],
-            [lambda cmd, no_output, chk_err: (-1, ''), self.mock_linux_distribution, str()],    # no matches for any package manager
+            [self.mock_run_command_for_apt, self.mock_linux_distribution, self.mock_distro_os_release_attr_return_none, Constants.APT],
```
rane-rajasi: Why change existing UTs? Your code change does not affect the behavior for AzLinux3 and AzLinux2/Mariner.

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -62,15 +62,33 @@ def get_package_manager(self):
...
+        major = version.split('.')[0] if version else None
+        os_id = distro.id()
+
+        # Check for Azure Linux
         if self.is_distro_azure_linux(str(self.platform.linux_distribution())):
+            # Azure Linux 4 not yet supported
+            if major is not None and int(major) == 4:
+                error_msg = "Azure Linux 4 is not yet supported in your region. Please review aka.ms/LinuxPatchExtension for more information."
+                print("Error: {0}".format(error_msg))
+                raise Exception(error_msg)
```
rane-rajasi: We don't need to raise an Exception here. Print will log it to a stack trace at the agent level which CRP should surface up in it's supportedstate determination

src/core/src/bootstrap/EnvLayer.py
rane-rajasi: There's a function linux_distribution() in EnvLayer -> Platform. The code uses that to determine distro. Use that unless it doesn't work for your scenario.

src/core/src/bootstrap/EnvLayer.py
rane-rajasi: Refer the function is_distro_azure_linux_3_or_beyond()

We will need to even modify is_distro_azure_linux_3_or_beyond(). It only checks if major version >=3 the check here has to be restricted to AzLinux3 only

This is what I'm thinking implementation wise...
- a function to fetch major verion
- 3 functions, 1 each for AzLinux3, AzLinux4 and RHEL10
- Each of these functions call the first one for major version, and perform their respective actions

rane-rajasi (changes requested):

src/core/src/bootstrap/EnvLayer.py
rane-rajasi: We can also try to optimize the functions for AzLinux4 and RHEL10 into a single generic one which can be used for any such distro blocking ask in the future

src/core/tests/Test_EnvLayer.py
yashnap: This change was made to address the copilot comment : 
Please refer: https://github.com/Azure/LinuxPatchExtension/pull/344#discussion_r3229508238

src/core/src/bootstrap/EnvLayer.py
yashnap: Removed Exception . Now it just has print

src/core/src/bootstrap/EnvLayer.py
yashnap: linux_distribution() return a tuple value('Red Hat', "Linux server", etc). We need to parse the value to get the exact name across versions. I am using distro_id since it directly return id : rhel which can be used to compare directly.

src/core/src/bootstrap/EnvLayer.py
yashnap: Updated the code to address both comments. Bifurcated the implementation to have separate functions for different responsibility.

src/core/tests/Test_EnvLayer.py
```diff
@@ -79,28 +93,32 @@ def test_get_package_manager(self):
...
+            [lambda cmd, no_output, chk_err: (-1, ''), self.mock_linux_distribution, self.mock_distro_os_release_attr_return_none, str()],    # no matches for any package manager
         ]
 
         for row in test_input_output_table:
             self.envlayer.run_command_output = row[0]
             self.envlayer.platform.linux_distribution = row[1]
+            distro.os_release_attr = row[2]
             package_manager = self.envlayer.get_package_manager()
-            self.assertTrue(package_manager is row[2])
+            self.assertTrue(package_manager is row[3])
```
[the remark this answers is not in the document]
yashnap: Ignoring this. Code already present. Only change is index.

src/core/src/bootstrap/EnvLayer.py
yashnap: Updated the code is_distro_azure_linux_3_or_beyond() to strictly restrict to 3.

src/core/src/bootstrap/Constants.py
```diff
@@ -197,7 +197,7 @@ class StatusTruncationConfig(EnumBackport):
     RED_HAT = 'Red Hat'
     SUSE = 'SUSE'
     CENTOS = 'CentOS'
-    AZURE_LINUX = ['Microsoft Azure Linux', 'Common Base Linux Mariner']
+    AZURE_LINUX = ['Azure Linux', 'Microsoft Azure Linux', 'Common Base Linux Mariner']
```
yashnap, line 200: Added this code since linux_distribution() returns a tuple and the value for AzureLinux 4 was ('Azure Linux', '4.0', '') which was failed to recognize and fell into default path instead of error path.

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -47,13 +47,21 @@ def __init__(self):
     def is_distro_azure_linux(distro_name):
         return any(x in distro_name for x in Constants.AZURE_LINUX)
 
+    @staticmethod
+    def _get_major_version():
+        # type: () -> str
```
rane-rajasi: Based on the implementation this will be

# type: () -> str or None

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -47,13 +47,21 @@ def __init__(self):
     def is_distro_azure_linux(distro_name):
         return any(x in distro_name for x in Constants.AZURE_LINUX)
 
+    @staticmethod
+    def _get_major_version():
```
rane-rajasi: rename to __try_get_major_version(), since this returns None when no version is found

Note on code patterns used in LPE (will be helpful in future changes): we always prefix internal functions with '__' (2 underscores)

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -47,13 +47,21 @@ def __init__(self):
...
+        # type: () -> str
+        """ Extracts major version from distro.os_release_attr('version').
+        Returns major version as string or None if not available. """
```
rane-rajasi: nit: """ Attempts to fetch major version of the distro. Returns version as strinf or None if not available. """

src/core/src/bootstrap/EnvLayer.py
rane-rajasi: This falls in the common code path, which will run on all machines be it Windows, Ubuntu, SUSE, RedHat, Mariner, Azure Linux 3, 4, etc. Do all of them support distro.id()? Does it throw an exception for any of those? This has the potential of being a breaking change. We need to ensure any change we add doesn't end up breaking something else in the code flow.

rane-rajasi: And your implementation pattern between RHEL and AzLinux 4 is not consistent. For AzLinux4, you use is_distro_azure_linux() which calls platform.linux_distribution() to get the distro name, which is why you've added it in Constants

But only for RHEL you're using distro.id(). We need to be consistent in our approaches unless something does not work at all.
There's a constant defined for RedHat on line 197 in Constants.py, you can use that to validate. This keeps the implementation consistent with everything else and does not introduce a possible breaking change

Another factor to be aware of in LPE all code should be python version agnostic. In the very least, it should support both Python 2 and 3. platform.linux_distribution() has been implemented to handle that.

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -81,6 +96,38 @@ def get_package_manager(self):
...
+
+        # Azure Linux 3 and below use TDNF
+        code, out = self.run_command_output('which tdnf', False, False)
+        if code == 0:
+            return Constants.TDNF
+        else:
+            print("Error: Expected package manager tdnf not found on this Azure Linux VM.")
+            return str()
+
+    def _get_rhel_package_manager(self, major):
```
rane-rajasi: Previous comment:
https://github.com/Azure/LinuxPatchExtension/pull/344/changes#r3242634825

**_We can also try to optimize the functions for AzLinux4 and RHEL10 into a single generic one which can be used for any such distro blocking ask in the future_**

I meant have a single generic function for AzLinux4 and RHEL10. Think of it this way, their current code flow (distro blocking) and future code flow (using dnf5) are going to be the same. With different functions you end up having repetitive code

rane-rajasi: Here's a sample of what an optimized code could look like: 
NOTE: This is a suggestion, please verify it and validate all existing and new UTs on it and add comments, modify code or variables, etc wherever necessary


    @staticmethod
    def __try_get_major_version():
        # type: () -> str or None
        """ Attempts to fetch major version of the distro. Returns version as strinf or None if not available."""
        version = distro.os_release_attr('version')
        major = version.split('.')[0] if version else None
        return major

    def check_for_distro_and_version_match(self, distro_name, distro_to_match, version_to_match):
        if distro_to_match == Constants.AZURE_LINUX and not self.is_distro_azure_linux(str(distro_name)):
            return False
        elif distro_to_match == Constants.RED_HAT and not Constants.RED_HAT.lower() in str(distro_name).lower():
            return False
        major = self.__try_get_major_version()
        return major is not None and int(major) == version_to_match

    @staticmethod
    def is_distro_azure_linux(distro_name):
        return any(x in distro_name for x in Constants.AZURE_LINUX)

    # comment from PR feedback: This function is used in other places (AzL3TdnfPackageManager), when changing this function name and signature please ensure it reflected accurately everywhere used
    def is_distro_azure_linux_3(self, distro_name):
        # type: (str) -> bool
        """ Checks if the current distro is Azure Linux 3 """
        return self.check_for_distro_and_version_match(distro_name, Constants.AZURE_LINUX, version_to_match=3)

    def is_distro_azure_linux_4(self, distro_name):
        # type: (str) -> bool
        """ Checks if the current distro is Azure Linux 4 """
        return self.check_for_distro_and_version_match(distro_name, Constants.AZURE_LINUX, version_to_match=4)

    def is_distro_rhel_10(self, distro_name):
        return self.check_for_distro_and_version_match(distro_name, Constants.RED_HAT, version_to_match=10)

    def get_package_manager(self):
        # type: () -> str
        """ Detects package manager type """
        if platform.system() == 'Windows':
            return Constants.APT

        # comment for PR review feedback only: platform.linux_distribution() returns a tuple with distro_name, version and a code. Something line: # ['Ubuntu', '20.04', 'focal']. distro_name will always be the first entry in the tuple.
        os_offer, os_version, os_code = self.env_layer.platform.linux_distribution()

        if self.is_distro_azure_linux_4(os_offer) or self.is_distro_rhel_10(os_offer):
            error_msg = "This distro is not yet supported in your region. Please review https://aka.ms/VMGuestPatchingCompatibility for more information. [Distro={0}]".format(str(os_offer)) # Print all 3 values here
            print("Error: {0}".format(error_msg))
            return str()

        if self.is_distro_azure_linux(str(os_offer)):
            code, out = self.run_command_output('which tdnf', False, False)
            if code == 0:
                return Constants.TDNF
            else:
                print("Error: Expected package manager tdnf not found on this Azure Linux VM.")
                return str()

        # choose default package manager
        package_manager_map = (('apt-get', Constants.APT),
                               ('yum', Constants.YUM),
                               ('zypper', Constants.ZYPPER))
        for entry in package_manager_map:
            code, out = self.run_command_output('which ' + entry[0], False, False)
            if code == 0:
                return entry[1]

        return str()

src/core/tests/Test_EnvLayer.py
rane-rajasi: Comment link doesn't take me to any specific comment
Re-iterating a guidance shared earlier, co-pilot may not always be correct in its analysis. Take it as a suggestion and analyze yourself, ensure correctness and apply it only after these considerations. 

In any code, existing UTs should not change if your PR is not about it. In this case, your PR does nothing to change how AzLinux3 and AzLinux2/Mariner work, so those UTs **should not change**

If an existing UT is impacted by your change, when it shouldn't (i.e. their code/functionality is not changed), that indicates the PR (current code change) has an issue and should be fixed

src/core/tests/Test_EnvLayer.py
```diff
@@ -139,6 +157,69 @@ def test_platform(self):
...
+        self.assertEqual(expected_msg in output, True)
+        # Verify empty string is returned
+        self.assertEqual(result, "")
+
+        # restore
+        distro.id = self.backup_distro_id
+        distro.os_release_attr = self.backup_distro_os_release_attr
+        self.envlayer.platform.linux_distribution = self.backup_linux_distribution
+        platform.system = self.backup_platform_system
+
```
rane-rajasi, line 186: optimize these functions, extract common code into generic functions

rane-rajasi (changes requested):

src/core/src/bootstrap/EnvLayer.py
yashnap: Done

src/core/src/bootstrap/EnvLayer.py
yashnap: done

src/core/src/bootstrap/EnvLayer.py
yashnap: Done

src/core/tests/Test_EnvLayer.py
yashnap: Done

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -47,30 +47,69 @@ def __init__(self):
     def is_distro_azure_linux(distro_name):
         return any(x in distro_name for x in Constants.AZURE_LINUX)
 
-    def is_distro_azure_linux_3_or_beyond(self):
-        # type: () -> bool
+    @staticmethod
+    def __try_get_major_version():
+        # type: () -> str or None
+        """ Attempts to get major version from distro.os_release_attr('version').
+        Returns major version as string or None if not available. """
```
rane-rajasi: nit: Make this into a single line comment

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -47,30 +47,69 @@ def __init__(self):
...
+        version = distro.os_release_attr('version')
+        major = version.split('.')[0] if version else None
+        return major
+
+
```
rane-rajasi: Remove extra line break

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -47,30 +47,69 @@ def __init__(self):
...
+        if self.is_distro_azure_linux(str(os_name)):
             code, out = self.run_command_output('which tdnf', False, False)
             if code == 0:
                 return Constants.TDNF
             else:
                 print("Error: Expected package manager tdnf not found on this Azure Linux VM.")
                 return str()
 
-        # choose default package manager
+
```
rane-rajasi: Remove extra line break. There should be no code change in this section

src/core/tests/Test_EnvLayer.py
```diff
@@ -110,20 +125,19 @@ def test_is_distro_azure_linux_3_or_beyond(self):
         self.backup_envlayer_distro_os_release_attr = distro.os_release_attr
 
         test_input_output_table = [
-            [self.mock_linux_distribution_to_return_azure_linux_3, self.mock_distro_os_release_attr_return_azure_linux_3, True],
-            [self.mock_linux_distribution_to_return_azure_linux_2, self.mock_distro_os_release_attr_return_azure_linux_2, False],
-            [self.mock_linux_distribution_to_return_azure_linux_3, self.mock_distro_os_release_attr_return_none, False]
+            [self.mock_linux_distribution_to_return_azure_linux_3(), self.mock_distro_os_release_attr_return_azure_linux_3, True],
```
rane-rajasi: Remove '()'. Refer original code. This is just a pointer to the function, it gets invoked in the for loop below

src/core/tests/Test_EnvLayer.py
rane-rajasi: Is this needed now?

src/core/tests/Test_EnvLayer.py
```diff
@@ -110,20 +125,19 @@ def test_is_distro_azure_linux_3_or_beyond(self):
...
+            [self.mock_linux_distribution_to_return_azure_linux_2(), self.mock_distro_os_release_attr_return_azure_linux_2, False],
+            [self.mock_linux_distribution_to_return_azure_linux_3(), self.mock_distro_os_release_attr_return_none, False]
         ]
 
         for row in test_input_output_table:
-            self.envlayer.platform.linux_distribution = row[0]
+            distro_name = row[0][0]  # Extract distro name from tuple (first element)
```
rane-rajasi: The equivalent of this to original code will be:
distro_name = row[0] () [0] # Adding spaces because of comment formatting in github, remove in code

src/core/tests/Test_EnvLayer.py
rane-rajasi: Edit test name to test_is_distro_azure_linux_3()

src/core/tests/Test_EnvLayer.py
```diff
@@ -139,6 +153,68 @@ def test_platform(self):
...
+        # Verify exact error message is logged
+        expected_msg = "Error: This distro is not yet supported in your region. Please review https://aka.ms/VMGuestPatchingCompatibility for more information. [Distro=Red Hat]"
+        self.assertIn(expected_msg, output)
+        # Verify empty string is returned
+        self.assertEqual(result, "")
+
+        # restore
+        self._restore_mocks()
+
+    def _restore_mocks(self):
```
rane-rajasi: __restore_mocks()

Refer the earlier comment on code formatting: https://github.com/Azure/LinuxPatchExtension/pull/344#discussion_r3249343383

src/core/tests/Test_EnvLayer.py
```diff
@@ -139,6 +153,68 @@ def test_platform(self):
...
+        # Verify exact error message is logged
+        expected_msg = "Error: This distro is not yet supported in your region. Please review https://aka.ms/VMGuestPatchingCompatibility for more information. [Distro=Microsoft Azure Linux]"
+        self.assertIn(expected_msg, output)
+        # Verify empty string is returned
+        self.assertEqual(result, "")
+
+        # restore
+        self._restore_mocks()
+
+    def test_get_package_manager_rhel_10_not_supported(self):
```
rane-rajasi: This and test_get_package_manager_azure_linux_4_not_supported() have a lot of common code that can be made generic

rane-rajasi: Infact, these 2 tests can also be combined into one. Refer the patterns followed in other tests in this class on how the input output matrix is used to test different scenarios
eg: test_is_distro_azure_linux_3_or_beyond()

rane-rajasi (changes requested):
Make sure to upload latest inVM testing logs since the code has changed from its previous implementation

src/core/tests/Test_EnvLayer.py
yashnap: Done

src/core/tests/Test_EnvLayer.py
yashnap: Done

src/core/tests/Test_EnvLayer.py
yashnap: Fixed

src/core/src/bootstrap/EnvLayer.py
yashnap: Removed

yashnap:
> Make sure to upload latest inVM testing logs since the code has changed from its previous implementation

@rane-rajasi The outputs attached in description is latest from yesterday after the change.

kjohn-msft (dismissed):

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -47,22 +47,58 @@ def __init__(self):
...
             return Constants.APT
 
-        if self.is_distro_azure_linux(str(self.platform.linux_distribution())):
+        # platform.linux_distribution() returns a tuple with distro_name, version and a code
+        # Example: ['Azure Linux', '4.0', '']
+        os_name, os_version, os_code = self.platform.linux_distribution()
+
+        # Check for unsupported distros
+        if self.is_distro_azure_linux_4(os_name) or self.is_distro_rhel_10(os_name):
+            error_msg = "This distro is not yet supported in your region. Please review https://aka.ms/VMGuestPatchingCompatibility for more information. [Distro={0}]".format(str(os_name))
```
rane-rajasi: This should log os_version and os_code too because distro_name would only have something like 'Red Hat Enterprise Linux'. This doesn't tell us which image it is.
The tuple on line 92 would be similar to: ('Red Hat Enterprise Linux', '8.6', 'Ootpa')

rane-rajasi (changes requested):
Only one minor comment

rane-rajasi (approved):

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -47,22 +47,58 @@ def __init__(self):
...
         if platform.system() == 'Windows':
             return Constants.APT
 
-        if self.is_distro_azure_linux(str(self.platform.linux_distribution())):
+        # platform.linux_distribution() returns a tuple with distro_name, version and a code
+        # Example: ['Azure Linux', '4.0', '']
+        os_name, os_version, os_code = self.platform.linux_distribution()
+
+        # Check for unsupported distros
+        if self.is_distro_azure_linux_4(os_name) or self.is_distro_rhel_10(os_name):
```
kjohn-msft, line 95: There is a risk that non-marketplace-based RHEL 10 may get a new error. However, this is considered acceptable at this time as we've not formally declared support.

kjohn-msft (approved):

PR #347, opened 2026-05-21
Fix: Failing UT

yashnap:
- Fix failing UT

**Why it Failed in Pycharm:**
The test failed in PyCharm because PyCharm uses a custom unittest runner that wraps and manages sys.stdout for output capture. This conflicts with manual reassignment of sys.stdout, causing it to be replaced at runtime with a _io.TextIOWrapper that does not support getvalue(), leading to an AttributeError.

**Why it passed in PR (CI)  ( My assumption/Copilot Suggestion)** 
The test passed in the PR pipeline because it was executed using the standard Python unittest runner (python -m unittest), which does not override or wrap sys.stdout. As a result, the manual reassignment sys.stdout = StringIO() remained intact, and the captured output could be read correctly using getvalue()

The failure is due to a conflict between manual sys.stdout reassignment and PyCharm’s test runner, which overrides stdout handling, while CI uses the standard unittest runner without such interference.

-Since we will be making changes to the PackageManager code to remove this behavior and actually return a package manager value, this can be used as temporary workaround

### DIFF
--- a/src/core/tests/Test_EnvLayer.py
+++ b/src/core/tests/Test_EnvLayer.py
@@ -17,6 +17,12 @@ import io
 import platform
 import sys
 import unittest
+# Conditional import for StringIO
+try:
+    from StringIO import StringIO  # Python 2
+except ImportError:
+    from io import StringIO  # Python 3
+
 from core.src.bootstrap.EnvLayer import EnvLayer
 from core.src.bootstrap.Constants import Constants
 from core.src.external_dependencies import distro
@@ -165,10 +171,16 @@ class TestExecutionConfig(unittest.TestCase):
         ]
 
         for row in test_input_output_table:
+            captured_output = StringIO()
+            original_output = sys.stdout
+            sys.stdout = captured_output
             self.envlayer.platform.linux_distribution = row[0]
             distro.os_release_attr = row[1]
-            package_manager = self.envlayer.get_package_manager()
-            self.assertEqual(package_manager, "")
+
+            result = self.envlayer.get_package_manager()
+            sys.stdout = original_output
+            self.assertEqual(row[2], captured_output.getvalue())
+            self.assertEqual(result, "")
 
         # restore
         self.__restore_mocks()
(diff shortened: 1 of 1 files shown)


kjohn-msft (dismissed):

src/core/tests/Test_EnvLayer.py
```diff
@@ -167,13 +167,8 @@ def test_get_package_manager_azure_linux_4_and_rhel10_not_supported(self):
         for row in test_input_output_table:
             self.envlayer.platform.linux_distribution = row[0]
             distro.os_release_attr = row[1]
-
-            captured_output = io.StringIO()
-            sys.stdout = captured_output
-            result = self.envlayer.get_package_manager()
-            sys.stdout = sys.__stdout__
```
rane-rajasi, line 174: @yashnap, I think the problem was this line. See other examples in code where this is already implemented, for eg: test_reset_auto_assessment_log_file_if_needed_raise_exception()

you are setting sys.stdout to captured_output on line 171, but on line 174 while resetting sys.stdout it is set to "sys.__ stdout __ " (adding space because of github formatting). What does sys.__ stdout __ represent? Why not reset it to its original value?

rane-rajasi (changes requested):

src/core/tests/Test_EnvLayer.py
yashnap: sys.stdout = sys.____stdout___ - This is basically forcing stdout back to the "real" original console. captured_output = io.StringIO() was also an issue because of this : `TypeError: unicode argument expected, got 'str'` . One way to do it is do the way in other test you pointed out but we will be updating this code again when the actual implementation goes in and this method will return the packageManager value instead (i.e other tests in the same file) not needing the sys_out/capture_output anymore .

rane-rajasi: @yashnap since we already have a way to test this out in code, it's always better to have a UT coverage for a code rather than not having, however small the period of code's existence be.

Also, if you are not going to validate logs, please remove log values from test_input_output_table. That is  what Copilot's comment below is referring too. Removing validation on logs but still keeping it in the test matrix will just lead to confusion in the future. Always ensure your code does not leave any outliers/unreferenced sections.

rane-rajasi (changes requested):
Follow-up on previous comment left inline

rane-rajasi (approved):

kjohn-msft (approved):

PR #313, opened 2025-04-18
Feature: Updated LPE Packager with Auto-version update publish support + misc. improvements

kjohn-msft:
Packager updates pulled out to code base debt erasure (and revised a little further):
https://github.com/Azure/LinuxPatchExtension/pull/224

**Changes in this PR:**
1. Addition of **build.bat / .sh** in the src folder to package all components. [Builds **without** version change]
2. Addition of **publish.bat / .sh** in the src folder to increment the manifest version and package all components. (reduces friction in PR for publish, i.e. new branch -> publish (at prompt) -> push branch -> PR) [Builds **with** automatic version increase]
3. Removal of version information from multiple places and centralizing to the manifest source of truth that gets used in packaging.
4. Moved packager scripts to a packager sub-folder. Updated references. _This will be reflected in the internal make file._
5. Updated component naming to reduce confusion in multi-extension support scenarios.
6. Cleanup output and removed misc. debt.

Example of Publish output now (a super-set of build / package) --
image

PR #352, opened 2026-06-15
Engg. Hygiene: Introduce Extension Status Asserter and use in example Test class

kjohn-msft:
Test harness improvement to allow unit tests to deterministically check extension status.

**In this PR:**
1. Addresses issues like widely broken tests if order of entries in the extension status output changes. Indexing is by **name**, not by order (so no brittle hardcoding).
2. Has performant one-time memory load and evaluation.
3. **Assert mechanics:**
- Status validation - _assert_status_file_substatus(operation, expected_status)_
- Batch status validation - _assert_status_file_substatuses()_
- Patch-level validation - _assert_operation_summary_has_patch(...)_
- Error validation - _assert_operation_summary_has_error(...)_
- Metadata validations - assert_operation_summary_has_started_by(), assert_configure_patching_patch_mode_state(), assert_configure_patching_auto_assessment_state(), assert_healthstore_status_info()
_- Extensible beyond this_

4. Applies all this into the ConfigurePatching unit test flow.
- This exercise revealed that a test written for auto-assessment for Arc wasn't guarding against everything it could guard against. I'm leaving that out of the scope of this PR.

**Future changes:**
All code authors can rely on this for new tests.
Existing tests may be rewritten as a new-hire exercise.

(Note: This was previously bundled into the commit history of https://github.com/Azure/LinuxPatchExtension/pull/313/changes - moved it out separately as good practice)

### DIFF
--- a/src/core/src/core_logic/ConfigurePatchingProcessor.py
+++ b/src/core/src/core_logic/ConfigurePatchingProcessor.py
@@ -17,9 +17,18 @@
 """ Configure Patching """
 from core.src.bootstrap.Constants import Constants
 
+from core.src.bootstrap.EnvLayer import EnvLayer
+from core.src.core_logic.ExecutionConfig import ExecutionConfig
+from core.src.local_loggers.CompositeLogger import CompositeLogger
+from core.src.service_interfaces.StatusHandler import StatusHandler
+from core.src.package_managers.PackageManager import PackageManager
+from core.src.core_logic.ServiceManager import ServiceManager
+from core.src.core_logic.TimerManager import TimerManager
+
 
 class ConfigurePatchingProcessor(object):
     def __init__(self, env_layer, execution_config, composite_logger, telemetry_writer, status_handler, package_manager, auto_assess_service_manager, auto_assess_timer_manager, lifecycle_manager):
+        # type: (EnvLayer, ExecutionConfig, CompositeLogger, TelemetryWriter, StatusHandler, PackageManager, ServiceManager, TimerManager) -> None
         self.env_layer = env_layer
         self.execution_config = execution_config
 
@@ -40,7 +49,7 @@ class ConfigurePatchingProcessor(object):
     def start_configure_patching(self):
         """ Start configure patching """
         try:
-            self.composite_logger.log("\nStarting configure patching... [MachineId: " + self.env_layer.platform.vm_name() + "][ActivityId: " + self.execution_config.activity_id + "][StartTime: " + self.execution_config.start_time + "]")
+            self.composite_logger.log("[CPP] Starting configure patching... [MachineId: " + self.env_layer.platform.vm_name() + "][ActivityId: " + self.execution_config.activity_id + "][StartTime: " + self.execution_config.start_time + "]")
             self.status_handler.set_current_operation(Constants.CONFIGURE_PATCHING)
             self.__raise_if_telemetry_unsupported()
 
@@ -73,6 +82,7 @@ class ConfigurePatchingProcessor(object):
     def __try_set_patch_mode(self):
         """ Set the patch mode for the VM """
         try:
+            self.composite_logger.log_verbose("[CPP] Processing patch mode configuration...")
             self.status_handler.set_current_operation(Constants.CONFIGURE_PATCHING)
             self.current_auto_os_patch_state = self.package_manager.get_current_auto_os_patch_state()
 
@@ -87,9 +97,9 @@ class ConfigurePatchingProcessor(object):
 
             if self.execution_config.patch_mode == Constants.PatchModes.AUTOMATIC_BY_PLATFORM and self.current_auto_os_patch_state == Constants.AutomaticOSPatchStates.UNKNOWN:
                 # NOTE: only sending details in error objects for customer visibility on why patch state is unknown, overall configurepatching status will remain successful
-                self.configure_patching_exception_error = "Could not disable one or more automatic OS update services. Please check if they are configured correctly"
+                self.configure_patching_exception_error = "Could not disable one or more automatic OS update services. Please check if they are configured correctly."
 
-            self.composite_logger.log_debug("Completed processing patch mode configuration.")
+            self.composite_logger.log_debug("[CPP] Completed processing patch mode configuration.")
         except Exception as error:
             self.composite_logger.log_error("Error while processing patch mode configuration. [Error={0}]".format(repr(error)))
             self.configure_patching_exception_error = error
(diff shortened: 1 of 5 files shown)


src/core/src/core_logic/ConfigurePatchingProcessor.py
```diff
@@ -40,7 +49,7 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
     def start_configure_patching(self):
         """ Start configure patching """
         try:
-            self.composite_logger.log("\nStarting configure patching... [MachineId: " + self.env_layer.platform.vm_name() + "][ActivityId: " + self.execution_config.activity_id + "][StartTime: " + self.execution_config.start_time + "]")
+            self.composite_logger.log("[CPP] Starting configure patching... [MachineId: " + self.env_layer.platform.vm_name() + "][ActivityId: " + self.execution_config.activity_id + "][StartTime: " + self.execution_config.start_time + "]")
```
michellemcdaniel, line 52: Not for this PR but an idea for future logging improvement: it seems like every piece of the code base wants the logging to start with [abbreviation of where the logging comes from], but sometimes it gets missed. Would it be at all feasible to have a custom logger that enforces this, either by taking two strings as parameters, the first being the abbreviation (best practice a const used everywhere in the class) and the second the actual logging message, and then the logger combines them, or having some sort of templated logger that automatically derives the abbreviation? I know this is python so I don't know if that's something we can do, but having a code-based enforcement of the logging standards would be much easier/dev and reviewer friendly than combing over every logging statement and something having to make changes like those in this file.

michellemcdaniel:
> This exercise revealed that a test written for auto-assessment for Arc wasn't guarding against everything it could guard against. I'm leaving that out of the scope of this PR.

Do we have a work item for improving the noted test?

michellemcdaniel (approved):

src/core/tests/library/ExtStatusAsserter.py
```diff
@@ -0,0 +1,179 @@
+# Copyright 2025 Microsoft Corporation
```
rane-rajasi, line 1: nit: 2026

src/core/tests/library/ExtStatusAsserter.py
```diff
@@ -0,0 +1,179 @@
...
+            Constants.PATCH_METADATA_FOR_HEALTHSTORE: {"index": -1, "status": None},
+        }
+
+    def __get_substatus_index_with_assert(self, operation):
+        # type: (str) -> int
+        """ Get the index of the substatus """
+        if operation not in self.__substatus_high_level_summary:
+            raise KeyError("Unknown operation: {0}".format(operation))
+
+        substatus_index = self.__substatus_high_level_summary[operation]["index"]
```
rane-rajasi, line 48: What will "index" be? We don't have this in the structure for status

src/core/tests/Test_ConfigurePatchingProcessor.py
```diff
@@ -76,24 +80,15 @@ def test_operation_success_for_configure_patching_request_for_apt_with_default_u
...
-        with runtime.env_layer.file_system.open(runtime.execution_config.status_file_path, 'r') as file_handle:
-            substatus_file_data = json.load(file_handle)[0]["status"]["substatus"]
-
-        # check status file for configure patching patch state (and including for 'Platform' initiated assessment data)
+        # assertions
         self.assertTrue(runtime.package_manager.image_default_patch_configuration_backup_exists())
-        self.assertEqual(len(substatus_file_data), 2)
-        self.assertTrue(substatus_file_data[0]["name"] == Constants.PATCH_ASSESSMENT_SUMMARY)
-        message = json.loads(substatus_file_data[0]["formattedMessage"]["message"])
-        self.assertTrue(message["startedBy"], Constants.PatchAssessmentSummaryStartedBy.PLATFORM)
```
rane-rajasi, line 88: This assert is missing

src/core/tests/Test_ConfigurePatchingProcessor.py
```diff
@@ -338,19 +299,18 @@ def test_configure_patching_raise_exception_auto_os_patch_state(self):
 
         runtime.configure_patching_processor.start_configure_patching()
 
-        # restore sdt.out ouptput
+        # restore sdt.out output
```
rane-rajasi, line 302: nit: sys.stdout or just stdout

src/core/tests/library/ExtStatusAsserter.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+class ExtStatusAsserter(object):
+    def __init__(self, status_file_path, env_layer):
+        self.__status_file_path = status_file_path
+        self.__env_layer = env_layer
+
+        self.__substatus_file_data = self.__read_status_file(self.__status_file_path)
+        self.__substatus_high_level_summary = None
+        self.__load_substatus_high_level_summary(self.__substatus_file_data)
```
rane-rajasi, line 28: Wouldn't self.__substatus_file_data be available to __load_substatus_high_level_summary()? Need not pass in as an argument

src/core/tests/library/ExtStatusAsserter.py
```diff
@@ -0,0 +1,179 @@
...
+        with self.__env_layer.file_system.open(status_file_path, 'r') as file_handle:
+            return json.load(file_handle)[0]["status"]["substatus"]
+
+    def __load_substatus_high_level_summary(self, substatus_file_data):
+        # type: (dict) -> None
+        """ Makes one-time inferences about the structure of the status file """
+        self.__substatus_high_level_summary = self.__get_high_level_summary_template()
+
+        for index, substatus in enumerate(substatus_file_data):
+            summary_name = substatus["name"]
```
rane-rajasi, line 78: This should be summary_name = substatus[index]["name"], that would contain 'ConfigurePatching' or 'PatchAssessmentSummary', etc

src/core/tests/library/ExtStatusAsserter.py
rane-rajasi: I think I see how 'index' works  from __get_high_level_summary_template()

src/core/tests/library/ExtStatusAsserter.py
```diff
@@ -0,0 +1,179 @@
...
+        substatus_message = self.__get_substatus_message_as_dict(Constants.CONFIGURE_PATCHING_SUMMARY)
+        if substatus_message["autoAssessmentStatus"]["autoAssessmentState"] != expected_state:
+            raise AssertionError("Auto-assessment state '{0}' does not match expected value '{1}'.".format(substatus_message["autoAssessmentStatus"]["autoAssessmentState"], expected_state))
+
+    def assert_healthstore_status_info(self, patch_version, should_report=True):
+        # type: (str, bool) -> None
+        """Check if the healthstore patch version is as expected"""
+        healthstore_summary = self.__get_substatus_message_as_dict(Constants.PATCH_METADATA_FOR_HEALTHSTORE)
+
+        if should_report and healthstore_summary["shouldReportToHealthStore"] != True:
```
rane-rajasi, line 174: nit or maybe another PR: All of these keys should be defined in Constants and referred from there. We will also need to modify StatusHandler.py for this

rane-rajasi (changes requested):
Comments inline

rane-rajasi (approved):

PR #354, opened 2026-06-17
Fix: [UT fix] Fixing DNF5/AzL4 UTs failing in Py2.7

rane-rajasi:
DNF5 UTs were failing when run on python 2.7 on my dev machine due to unimplemented abstract methods

Code coverage, failures in this PR, are due to the abstract methods bypassing and can be ignored

eg:
image

Local UT runs after change:
Py 3.14:
image

Py 2.7:
image

### DIFF
--- a/src/core/src/package_managers/Dnf5PackageManager.py
+++ b/src/core/src/package_managers/Dnf5PackageManager.py
@@ -18,7 +18,6 @@
 import json
 import re
 
-from abc import ABCMeta, abstractmethod
 from core.src.core_logic.VersionComparator import VersionComparator
 from core.src.bootstrap.Constants import Constants
 from core.src.package_managers.PackageManager import PackageManager
@@ -80,8 +79,6 @@ class Dnf5PackageManager(PackageManager):
         self.STR_TOTAL_DOWNLOAD_SIZE = "Total download size: "
         self.package_install_expected_avg_time_in_seconds = 90
         self.version_comparator = VersionComparator()
-
-    __metaclass__ = ABCMeta  # For Python 3.0+, it changes to class Abstract(metaclass=ABCMeta)
 
     def refresh_repo(self):
         self.composite_logger.log("[DNF5] Refreshing local repo...")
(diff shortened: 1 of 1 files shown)


rane-rajasi:
The code coverage failures are due to the abstract methods bypassing and can be ignored

eg:
image

src/core/src/package_managers/Dnf5PackageManager.py
```diff
@@ -81,8 +80,6 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
         self.package_install_expected_avg_time_in_seconds = 90
         self.version_comparator = VersionComparator()
 
-    __metaclass__ = ABCMeta  # For Python 3.0+, it changes to class Abstract(metaclass=ABCMeta)
```
michellemcdaniel, line 84: Does this still work on python 3.0+? Based on the previous comment, i'm wondering if this is expected to work on both

kjohn-msft: This isn't an abstract class and doesn't need any abstractions.

src/core/src/package_managers/Dnf5PackageManager.py
```diff
@@ -206,9 +202,8 @@ def get_composite_package_identifier(self, package, package_version):
             package_identifier += arch
         return package_identifier
 
-    @abstractmethod
     def install_updates_fail_safe(self, excluded_packages):
-        pass
+        return
```
michellemcdaniel, line 206: why is this one return when the others are pass?

kjohn-msft (approved):

michellemcdaniel (approved):

src/core/src/package_managers/Dnf5PackageManager.py
rane-rajasi: For this instance or examples where the function does nothing, pass or return both do the same thing. 

Ideally, this is use case:
Pass: Use pass for placeholders/stubs (TODO, abstract-ish empty body, empty class, empty except block).
Return: Use empty return when you want to explicitly stop execution at that point (early exit or explicit no-op override).

I tried to match these with how this is in existing package managers, hence the return in one function and pass in other

michellemcdaniel: That's what I thought. I guess I was more wondering why change it? Was it being pass breaking a test?

rane-rajasi: No, it wasn't breaking the UT. Just kept the format similar to other package managers

PR #343, opened 2026-05-11
Feature: Adding Azure Linux 4.0 Base Support using dnf5 package manager

yashnap:
Implemented DnfPackageManager by extending PackageManager.
- Added ConfigurationFactory, EnvLayer and Constants change needed across the board for all operations

[x] DNF5 package manager
[x] ConfigurePatching operation
[x] Auto Assessment operation
[x] Assessment operation
[x] Installation operation
[x] UTs
[x] Disabling Auto OS updates

**_FIRST ITERATION_** ( Enablement/Disablement) 
- Added changes to support detection, enablement and disablement.

 Implemented the below methods:
- get_current_auto_os_patch_state,
- disable_auto_os_update, 
- disable_auto_os_update_for_dnf5_automatic, 
- disable_auto_update_on_reboot, 
- is_auto_update_service_installed, 
- is_service_set_to_enable_on_reboot, 
- enable_auto_update_on_reboot

- Added few UT's to cover the implementation(more to follow)

**TESTING**
- Tested it on Linux 4 VM by creating a helper script to detect, disable and enable OS update using dnf5-automatic

ARM ID : /subscriptions/6acc8a91-e2b0-4041-a069-c2932ab42fd9/resourceGroups/rg-yashna-linux4-test-wus2/providers/Microsoft.Compute/virtualMachines/vm-yashna-linux4

Enablement_Disablement

<--------------------ITERATION2------------------------->
-  Automatic assessment (configuring, running, reporting, disablement)

[output.txt](https://github.com/user-attachments/files/28124378/output.txt)

<------------------ITERATION3---------------->
- Assess, Install and Reboot Detection

1. Install Update package available
[output.txt](https://github.com/user-attachments/files/28285246/output.txt)

2. All packages are upto date and already installed
[output_alreadyUptoDate.txt](https://github.com/user-attachments/files/28285247/output_alreadyUptoDate.txt)

<-------ITERATION4 ------->
- Updates to Install/AssessPatch methods 

[output_new.txt](https://github.com/user-attachments/files/28315038/output_new.txt)

<---- ITERATION5-------->
- UT updates, minor code updates to have consistency.

<-------------------AUTO-OS -------------------------->
**AUTO OS** 

ANALYSIS:
1. Check systemd service definition
- Command : systemctl cat dnf5-automatic.service 
- Output : ExecStart=/usr/bin/dnf5 automatic --timer ( Default) 
[...]

PR #355, opened 2026-06-22
Address Pending Comments from Dnf5 Original PR

yashnap:
- Address Pending comments from original PR: https://github.com/Azure/LinuxPatchExtension/pull/343 

1. https://github.com/Azure/LinuxPatchExtension/pull/343#discussion_r3396080638  ( Only shows 1 extra line when pulled latest from master)
Comment_1

2. https://github.com/Azure/LinuxPatchExtension/pull/343#discussion_r3396264815 ( This is fixed since new changes were added at the end of file(
Comment_2

3.https://github.com/Azure/LinuxPatchExtension/pull/343#discussion_r3396615113 (I already see 2 extra lines at the bottom )
Comment_3

4. https://github.com/Azure/LinuxPatchExtension/pull/343#discussion_r3405475084 - Updated
5. https://github.com/Azure/LinuxPatchExtension/pull/343#discussion_r3405480078 - Updated
6. https://github.com/Azure/LinuxPatchExtension/pull/343#discussion_r3405703821 - Added UT
7. https://github.com/Azure/LinuxPatchExtension/pull/343#discussion_r3406646381 - Fixed

Verified that all tests are running in both Python2 and Python3


src/core/tests/Test_EnvLayer.py
```diff
@@ -96,18 +107,20 @@ def test_get_package_manager(self):
...
+            [self.mock_run_command_for_dnf5, self.mock_linux_distribution_to_return_azure_linux_4, Constants.DNF5, self.mock_distro_os_release_attr_return_azure_linux_4],
+            [self.mock_run_command_for_tdnf, self.mock_linux_distribution_to_return_azure_linux_3, Constants.TDNF, self.mock_distro_os_release_attr_return_azure_linux_3],
+            [self.mock_run_command_for_yum, self.mock_linux_distribution_to_return_azure_linux_3, str(), self.mock_distro_os_release_attr_return_none],  # check for Azure Linux machine which does not have tdnf
+            [self.mock_run_command_for_tdnf, self.mock_linux_distribution_to_return_azure_linux_2, Constants.TDNF, self.mock_distro_os_release_attr_return_azure_linux_3],
+            [self.mock_run_command_for_yum, self.mock_linux_distribution, Constants.YUM, self.mock_distro_os_release_attr_return_none],
+            [self.mock_run_command_for_zypper, self.mock_linux_distribution, Constants.ZYPPER, self.mock_distro_os_release_attr_return_none],
+            [lambda cmd, no_output, chk_err: (-1, ''), self.mock_linux_distribution, str(), self.mock_distro_os_release_attr_return_none],    # no matches for any package manager
```
yashnap: Added 4th column to address the following issue:
1. After adding DNF5 UT change
2.. Calls get_package_manager() → is_distro_azure_linux_4()
3. is_distro_azure_linux_4() calls __try_get_major_version()
4. Since we don't set the os_release_attr in the method , it defaults to None and major = None → is_distro_azure_linux_4() returns False
5. This leads to is_distro_azure_linux_4 returning false and call goes to tdnf if check and fails
- Added 4th column to have os_release_attribute for each so version can be set.

- Let me know your thoughts on this.

src/core/tests/Test_EnvLayer.py
```diff
@@ -96,18 +107,20 @@ def test_get_package_manager(self):
...
         ]
 
         for row in test_input_output_table:
             self.envlayer.run_command_output = row[0]
             self.envlayer.platform.linux_distribution = row[1]
+            distro.os_release_attr = row[3]
             package_manager = self.envlayer.get_package_manager()
             self.assertTrue(package_manager is row[2])
```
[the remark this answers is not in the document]
yashnap: Existing code but correct suggestion. Fixed this

src/core/tests/Test_EnvLayer.py
```diff
@@ -96,18 +107,20 @@ def test_get_package_manager(self):
...
-            [self.mock_run_command_for_yum, self.mock_linux_distribution, Constants.YUM],
-            [self.mock_run_command_for_zypper, self.mock_linux_distribution, Constants.ZYPPER],
-            [lambda cmd, no_output, chk_err: (-1, ''), self.mock_linux_distribution, str()],    # no matches for any package manager
+            [self.mock_run_command_for_apt, self.mock_linux_distribution, Constants.APT, self.mock_distro_os_release_attr_return_none],
+            [self.mock_run_command_for_dnf5, self.mock_linux_distribution_to_return_azure_linux_4, Constants.DNF5, self.mock_distro_os_release_attr_return_azure_linux_4],
+            [self.mock_run_command_for_tdnf, self.mock_linux_distribution_to_return_azure_linux_3, Constants.TDNF, self.mock_distro_os_release_attr_return_azure_linux_3],
+            [self.mock_run_command_for_yum, self.mock_linux_distribution_to_return_azure_linux_3, str(), self.mock_distro_os_release_attr_return_none],  # check for Azure Linux machine which does not have tdnf
+            [self.mock_run_command_for_tdnf, self.mock_linux_distribution_to_return_azure_linux_2, Constants.TDNF, self.mock_distro_os_release_attr_return_azure_linux_3],
```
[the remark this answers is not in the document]
yashnap: Fixed this

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -99,8 +99,8 @@ def get_package_manager(self):
 
         # Check for Azure Linux 4 or Above( uses dnf5)
         if self.is_distro_azure_linux_4(str(os_name)):
-            code, out = self.run_command_output('which dnf', False, False)
-            if code == 0:
+            code, out = self.run_command_output('dnf --version', False, False)
+            if code == 0 and out and str(out).strip().startswith('5'):
                 return Constants.DNF5
```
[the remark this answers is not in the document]
yashnap: Combined which dnf + version check into one. Currently we have a separate check for dnf4 use-case( RHEL 10) in our case. Keeping this as is. If there is no dnf present it will error out.

src/core/tests/Test_EnvLayer.py
```diff
@@ -71,6 +74,14 @@ def mock_run_command_for_tdnf(self, cmd, no_output=False, chk_err=False):
             return 0, ''
         return -1, ''
 
+    def mock_run_command_for_dnf5(self, cmd, no_output=False, chk_err=False):
+        if cmd.find("dnf --version") > -1:
+            return 0, '5.2.0'
+        return -1, ''
```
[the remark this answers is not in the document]
yashnap: Since this is a mock, no need for actual path. Keeping it as is

src/core/src/bootstrap/EnvLayer.py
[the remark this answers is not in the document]
kjohn-msft: Not diagnosable as-is in the error condition. If we see this error line provided to you: "Error: Expected package manager dnf5 not found on this Azure Linux4 VM."   -- without getting access to the VM, what do we know about what went wrong?

kjohn-msft (changes requested):

src/core/src/bootstrap/EnvLayer.py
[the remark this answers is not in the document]
yashnap: Make sense. I've made changes to separate the check and add logging for easier diagnosis without having access o the VM.

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -99,13 +99,20 @@ def get_package_manager(self):
...
+                print("Error: Expected package manager dnf not found on this Azure Linux4 VM.")
                 return str()
 
+            code, out = self.run_command_output('dnf --version', False, False)
+            if code == 0 and out:
+                #Output : dnf5 version 5.2.18.0
+                version = str(out).split()[-1]
+                if version.startswith('5'):
+                    return Constants.DNF5
+            print("Error: Expected dnf version 5 not found on this Azure Linux4 VM.")
```
kjohn-msft: Yes, and what did dnf --version actually return.

Imagine it returns:
- an error
- 4 - surprising but clear
- 6 - something changed

As it is here, it's not possible to distinguish between these 3 states (or more).

src/core/tests/Test_EnvLayer.py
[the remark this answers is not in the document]
yashnap: Added path for which dnf command

src/core/src/bootstrap/EnvLayer.py
yashnap: Added logic to log version if not 5 and generic error message for dnf --version command error with logs.

src/core/tests/Test_EnvLayer.py
```diff
@@ -168,6 +201,50 @@ def test_is_distro_azure_linux_3(self):
...
+        self.backup_platform_system = platform.system
+        self.backup_linux_distribution = self.envlayer.platform.linux_distribution
+        self.backup_run_command_output = self.envlayer.run_command_output
+        self.backup_distro_os_release_attr = distro.os_release_attr
+
+        platform.system = self.mock_platform_system
+        self.envlayer.platform.linux_distribution = self.mock_linux_distribution_to_return_azure_linux_4
+        distro.os_release_attr = self.mock_distro_os_release_attr_return_azure_linux_4
+
+        test_input_output_table = [
```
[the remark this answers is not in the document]
yashnap: Fixed

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -99,13 +111,19 @@ def get_package_manager(self):
...
-                return Constants.DNF5
-            else:
-                print("Error: Expected package manager dnf5 not found on this Azure Linux4 VM.")
+            if not self.__is_dnf_available():
+                print("Error: Expected package manager dnf not found on this Azure Linux4 VM.")
                 return str()
 
+            code, out, version = self.__get_dnf_version()
+            if version:
+                if version.startswith('5'):
```
michellemcdaniel: I'm wondering if we should be smarter about the version comparison here. I know this is highly unlikely, but if we get up to dnf50 (i know, i know), then this would return true. Can we instead split on periods, then compare the first part to 5 directly, rather than using starts with?

yashnap: We can do that. I'll address this in the next iteration.

src/core/src/package_managers/Dnf5PackageManager.py
rane-rajasi, line 723: Comment 2 is pending

rane-rajasi: Comment mentioned 2 empty lines I see only one both in my local and your screenshot in PR description

Image

src/core/tests/Test_EnvLayer.py
```diff
@@ -71,6 +74,33 @@ def mock_run_command_for_tdnf(self, cmd, no_output=False, chk_err=False):
...
+        if "dnf --version" in cmd:
+            return 0, 'dnf version 4.14.0'
+        return -1, ''
+
+    def mock_run_command_for_dnf_version_command_failure(self, cmd, no_output=False, chk_err=False):
+        if "which dnf" in cmd:
+            return 0, '/usr/bin/dnf'
+        if "dnf --version" in cmd:
+            return -1, 'dnf version command failure'
+        return -1, ''
```
rane-rajasi: Combine all mock_run_command_for_dnf_version_command_<>() into 1 function with input parameter defining the usecase

src/core/tests/Test_EnvLayer.py
rane-rajasi: Keep parameter to assert as the last one in a row.
i.e. [self.mock_run_command_for_apt, self.mock_linux_distribution, self.mock_distro_os_release_attr_return_none, Constants.APT]

src/core/tests/Test_EnvLayer.py
```diff
@@ -168,6 +201,51 @@ def test_is_distro_azure_linux_3(self):
...
+            self.envlayer.run_command_output = row[0]
+            result = self.envlayer.get_package_manager()
+            self.assertEqual(result, row[1])
+
+        # restore original methods
+        self.envlayer.run_command_output = self.backup_run_command_output
+        self.envlayer.platform.linux_distribution = self.backup_linux_distribution
+        distro.os_release_attr = self.backup_distro_os_release_attr
+        platform.system = self.backup_platform_system
+
```
rane-rajasi: This code should go in test_get_package_manager() since this is testing get_package_manager()

src/core/src/bootstrap/EnvLayer.py
```diff
@@ -99,13 +111,19 @@ def get_package_manager(self):
...
-            code, out = self.run_command_output('which dnf', False, False)
-            if code == 0:
-                return Constants.DNF5
-            else:
-                print("Error: Expected package manager dnf5 not found on this Azure Linux4 VM.")
+            if not self.__is_dnf_available():
+                print("Error: Expected package manager dnf not found on this Azure Linux4 VM.")
                 return str()
 
+            code, out, version = self.__get_dnf_version()
```
rane-rajasi, line 110: Won't __get_dnf_version() i.e. command 'dnf --version' return a non zero exit code if dnf is not available? What is the need for a separate __is_dnf_available() then?

if could simply be:

 # Check for Azure Linux 4 or Above( uses dnf5)
        if self.is_distro_azure_linux_4(str(os_name)):
            code, out, version = self.__get_dnf_version()
            if code == 0 and version == 5:
                return Constants.DNF5
            else:
                print("Error: Expected package manager dnf5 not found on this Azure Linux4 VM.")
                return str()

rane-rajasi (changes requested):
Comments inline

src/core/src/bootstrap/EnvLayer.py
yashnap: Good point. I split the checks intentionally to address the diagnosability concerns from the earlier review comment: https://github.com/Azure/LinuxPatchExtension/pull/355#discussion_r3455204512 

With only dnf --version, multiple failure modes get folded into the same code path:
1. dnf is not installed / not found
2. dnf is present but returns version 4
3. dnf is present but returns version 6
4. dnf --version itself fails for some other reason 

By checking __is_dnf_available() first, we can provide a specific error when dnf is missing and __get_dnf_version() is only responsible for version-related validation. While dnf --version alone can be used to detect all cases, the separate availability check allows to distinguish "dnf missing" from "dnf present but invalid/unexpected"

src/core/tests/Test_EnvLayer.py
yashnap: I am assuming you mean : I remove this newly added test and add whatever its doing(checking error cases) into test_get_package_manager()

src/core/tests/Test_EnvLayer.py
yashnap: Done

src/core/src/package_managers/Dnf5PackageManager.py
yashnap: Fixed

src/core/src/bootstrap/EnvLayer.py
yashnap: Updated this

src/core/tests/Test_EnvLayer.py
yashnap: Moved to  test_get_package_manager()

yashnap:
https://github.com/Azure/LinuxPatchExtension/pull/355#discussion_r3645367901 - Fixed this, Somehow I am not able to reply to this and can't see it inline

kjohn-msft (dismissed):

michellemcdaniel (dismissed):

src/core/tests/Test_EnvLayer.py
```diff
@@ -126,20 +153,25 @@ def test_get_package_manager(self):
...
+            [self.mock_run_command_for_yum, self.mock_linux_distribution, self.mock_distro_os_release_attr_return_none, Constants.YUM],
+            [self.mock_run_command_for_zypper, self.mock_linux_distribution,  self.mock_distro_os_release_attr_return_none, Constants.ZYPPER],
+            [lambda cmd, no_output, chk_err: (-1, ''), self.mock_linux_distribution, self.mock_distro_os_release_attr_return_none, str()], # no matches for any package manager
+            [self.mock_run_command_for_dnf_not_found, self.mock_linux_distribution_to_return_azure_linux_4, self.mock_distro_os_release_attr_return_azure_linux_4, str()],
+            [lambda cmd, no_output, chk_err=False: self.mock_run_command_for_dnf_version_command(cmd, "wrong_version"), self.mock_linux_distribution_to_return_azure_linux_4, self.mock_distro_os_release_attr_return_azure_linux_4, str()],
+            [lambda cmd, no_output, chk_err=False: self.mock_run_command_for_dnf_version_command(cmd,"version_command_failure"), self.mock_linux_distribution_to_return_azure_linux_4, self.mock_distro_os_release_attr_return_azure_linux_4, str()]
```
rane-rajasi: Why the use of lambda here instead of mocks?

src/core/src/bootstrap/EnvLayer.py
rane-rajasi: I see these 3 points in Koshy's comment:

an error
4 - surprising but clear
6 - something changed

All of these points can still be addressed with just running one command 'dnf --version' because this is what determines if the code continues.
If 'which dnf' fails OR if 'which dnf' succeeds and 'dnf --version', both of these indicate a problem in the VM which needs to be logged for the VM owner to address.


    if self.is_distro_azure_linux_4(str(os_name)):
        code, out, version = self.__get_dnf_version()
        if code == 0 and version == 5:
            return Constants.DNF5
        elif code=0 and version !=5:
            print("Error: Expected dnf version not found on this Azure Linux4 VM. [Expected={0}][Found={1}]".format("5", str(version))
            return str()
        else:
            print("Error: Expected package manager dnf5 not found on this Azure Linux4 VM.")
            return str()

src/core/tests/Test_EnvLayer.py
```diff
@@ -71,6 +74,30 @@ def mock_run_command_for_tdnf(self, cmd, no_output=False, chk_err=False):
...
+    def mock_run_command_for_dnf_not_found(self, cmd, no_output=False, chk_err=False):
+        return -1, ''
+
+    def mock_run_command_for_dnf_version_command(self, cmd, usecase="success"):
+        if "which dnf" in cmd:
+            return 0, "/usr/bin/dnf"
+        if "dnf --version" in cmd:
+            if usecase == "wrong_version":
+                return 0, "dnf version 4.14.0"
+            if usecase == "version_command_failure":
```
rane-rajasi: This should be a switch case or an if else block with a single return at the end of the function

src/core/tests/Test_EnvLayer.py
```diff
@@ -280,6 +330,23 @@ def test_get_package_manager_azure_linux_4_and_rhel10_not_supported(self):
         # restore
         self.__restore_mocks()
 
+    def test_mock_command_fallback_paths(self):
+        """Test that mock commands return -1 for unexpected commands"""
+        code, out = self.mock_run_command_for_apt('which apt')
```
rane-rajasi, line 344: Why a test for mocked functions?

rane-rajasi (changes requested):
Comments inline

src/core/tests/Test_EnvLayer.py
yashnap: Done

src/core/tests/Test_EnvLayer.py
yashnap: Was trying to avoid adding more wrapper function hence the lamda. I've updated the code to use mocks

src/core/src/bootstrap/EnvLayer.py
yashnap: If VM owner needs to address it then it makes sense. Thanks
Updated the code.

src/core/tests/Test_EnvLayer.py
yashnap: Added this to have full coverage. One of Koshy's comments from Teams thread. https://teams.microsoft.com/l/message/19:a30b906bea084801ae7406e964dc929e@thread.skype/1782148751374?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=ff48c73a-32a8-4ce4-b51a-322055b1633c&parentMessageId=1782148751374&teamName=GuestPatching&channelName=%F0%9F%91%80%20PR%20Reviews&createdTime=1782148751374&ngc=true

rane-rajasi (dismissed):

kjohn-msft (approved):

rane-rajasi (approved):

PR #357, opened 2026-06-25
BugFix for DNF5: Exclusion list not honored

yashnap:
_### **Problem Statement**_
Linux Patch Extension performs dependency discovery during Include/Exclude processing so that exclusions can be propagated to dependent packages. i.e.
1: "patchesToInclude": ["openssl"], patchesToExclude": ["openssl-libs"]

In this scenario, if openssl-libs is identified as a dependency of openssl, both packages should be excluded from the final installation set.
However, dependency discovery was not working correctly for DNF5 upgrade scenarios. As a result, dependent packages were getting missed during Include/Exclude evaluation, causing exclusions to not be propagated as expected.

_### **Root Cause**_
Dependency discovery was using:
`dnf5 install --assumeno --skip-broken <package>`
This worked correctly for new package installation scenarios, which is what was validated during the initial DNF5 development effort.

Example:
```
dnf5 install --assumeno --skip-broken jq
Installing:
jq
Installing dependencies:
oniguruma
Operation aborted by the user.
```
In this case, DNF5 exposes the dependency graph because the package is not already installed.
However, LPE dependency discovery is executed for packages returned by:
`dnf5 check-update`
which are already-installed packages with updates available. Using the same simulation command for an installed package produces:
```
dnf5 install --assumeno --skip-broken openssl
Package "openssl" is already installed.
Nothing to do.
```
Since no transaction is generated, no dependency information is exposed and LPE is unable to discover dependent packages.

### _**Solution**_
Replace the dependency simulation command with:
`dnf5 upgrade --assumeno <package>`
This more accurately simulates the actual operation performed by LPE for packages returned from check-update.

Example:
```
dnf5 upgrade --assumeno openssl
Upgrading:
openssl
Upgrading dependencies:
openssl-libs
Transaction Summary:
Upgrading: 2 packages
Operation aborted by the user.
```
This allows dependency extraction to identify and correctly propagate exclusions.
openssl -> openssl-libs

Additionally, DNF5 dependency parsing was updated to recognize both in dependency sections:
```
Installing dependencies:
Upgrading dependencies:
```

### **Note on --skip-broken**
The previous implementation used:
`dnf5 install --assumeno --skip-broken`

while the new implementation uses without --skip-broken.: ( Dnf5 doesnt support skip-broken with uprade command: https://dnf5.readthedocs.io/en/latest/commands/upgrade.8.html#upgrade-command ) 
`dnf5 upgrade --assumeno`

As part of validation, DNF5 upgrade simulations were observed to return exit code 1 for both successful simulations and transaction resolution failures. Therefore, dependency simulation continues to evaluate both command output and exit code.
Additionally, dependency parsing now treats transaction resolution failures (for example, output containing:
Failed to resolve the transaction:) as non-blocking dependency discovery failures and returns an empty dependency list, maintaining behavior consistent with the existing dependency simulation flow.

**### _Validation_**
**1. Include + Exclude Dependency Scenario (Root Cause)** 
"patchesToInclude": ["openssl"],
"patchesToExclude": ["openssl-libs"]
Verified that dependency discovery correctly identified and exclusion propagation removed both packages from the installation set.
[3.core.openssl.log](https://github.com/user-attachments/files/29341724/3.core.openssl.log)

Additional Dependency Validation
"patchesToInclude": ["p11-kit"],
"patchesToExclude": ["p11-kit-trust"]
[new_include_exclude.txt](https://github.com/user-attachments/files/29341779/new_include_exclude.txt)

**2. Included Package With Dependencies**
"patchesToInclude": ["kernel"]
Verified that normal kernel installation behavior remains unchanged and required kernel packages continue to be installed successfully:
```
kernel
kernel-core
kernel-modules
kernel-modules-core
```
[4.core.included_dependencies.log](https://github.com/user-attachments/files/29341937/4.core.included_dependencies.log)

**3. Package-Only Inclusion Scenario**
[2.core.inclusion.log](https://github.com/user-attachments/files/29341838/2.core.inclusion.log)

**4. Include + Exclude Validation**
Include: kernel, kernel-modules, Exclude: kernel-core, kernel-modules-core.
[3.core.kernel.log](https://github.com/user-attachments/files/29341966/3.core.kernel.log)


src/core/src/package_managers/Dnf5PackageManager.py
```diff
@@ -36,7 +36,7 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
         self.single_package_check_versions = 'sudo dnf5 list --available <PACKAGE-NAME> '
         self.single_package_check_installed = 'sudo dnf5 list --installed <PACKAGE-NAME> '
 
-        self.single_package_upgrade_simulation_cmd = "sudo dnf5 install --assumeno --skip-broken "
+        self.single_package_upgrade_simulation_cmd = "sudo dnf5 upgrade --assumeno "
```
michellemcdaniel, line 39: Is this only used for upgrade scenarios?

yashnap: Yes, this is used for upgrade scenarios only. In cases where package is already installed and a newer/better version is available.

src/core/src/package_managers/Dnf5PackageManager.py
```diff
@@ -257,9 +257,8 @@ def get_dependent_list(self, packages):
...
         package_names = " ".join(packages)
         cmd = self.single_package_upgrade_simulation_cmd + package_names
-        # Dependency simulation using dnf5 install --assumeno --skip-broken has non-standard exit code behavior. A valid simulation run may return exit code 1 (e.g., "Operation aborted by the user"),
-        # while dependency resolution failures may still return exit code 0 with output indicating skipped packages (e.g., "Skipping packages with broken dependencies" and "Nothing to do.").
-        # calling the runcommand directly to get the output as well as code to determine failure/success cases
+        # DNF5 dependency simulation using `upgrade --assumeno` may return non-standard exit codes. Successful simulations and transaction
+        # resolution failures can both return exit code 1, therefore both command output and exit code are evaluated.
         code, output = self.env_layer.run_command_output(cmd, False, False)
         self.composite_logger.log_verbose("[DNF5] Dependency simulation. [Command={0}][Code={1}]".format(cmd, str(code)))
```
rane-rajasi: This should log output too. Complete output of every command needs to be logged at least once (generally with the run command log). 

2 possible ways to achieve this:
1. (Preferred) Use invoke_package_manager() to run command, which takes care of logging output. Refer TdnfPackageManager on how to address special exit code/cases for simulation commands

2. Explicitly log complete output here.

Option 1 is preferred since it keeps all command run behavior in a single place

rane-rajasi (changes requested):
Comment inline

src/core/src/package_managers/Dnf5PackageManager.py
yashnap: Updated the code with Option1

src/core/src/package_managers/Dnf5PackageManager.py
kjohn-msft: Why is  --skip-broken removed?
Fundamentally this is a bigger change that has potential impact on actual success and failure rates in various operations. How are these planned to be observed, understood and mitigated?

michellemcdaniel: Koshy, Yashna included why in the PR description:

> Note on --skip-broken
> The previous implementation used:
> dnf5 install --assumeno --skip-broken
> 
> while the new implementation uses without --skip-broken.: ( Dnf5 doesnt support skip-broken with uprade command: https://dnf5.readthedocs.io/en/latest/commands/upgrade.8.html#upgrade-command )
> dnf5 upgrade --assumeno
> 
> As part of validation, DNF5 upgrade simulations were observed to return exit code 1 for both successful simulations and transaction resolution failures. Therefore, dependency simulation continues to evaluate both command output and exit code.
> Additionally, dependency parsing now treats transaction resolution failures (for example, output containing:
> Failed to resolve the transaction:) as non-blocking dependency discovery failures and returns an empty dependency list, maintaining behavior consistent with the existing dependency simulation flow.

michellemcdaniel (approved):

rane-rajasi (approved):

src/core/src/package_managers/Dnf5PackageManager.py
yashnap: @kjohn-msft  I've tested a few scenarios ( include/exclude, dependencies, package only inclusion and install all.
This DNF5 page doesn't show skip-broken available as an option with upgrade : [Upgrade Command — dnf5 documentation](https://dnf5.readthedocs.io/en/latest/commands/upgrade.8.html#upgrade-command) ( mentioned in the PR) 
Can you elaborate on what do you mean by less apparent on standard systems ? Any scenario you think needs to be thoroughly evaluated ?

- Also the current code takes care of any failures that occur with the simulation command based on the output code and simulation command usage

kjohn-msft: There have been cases where an upgrade command blocks the installation of new dependencies a newer version of a package may have. This may not be apparent in point-in-time testing.

More concretely upgrade and install can drive the dependency solver differently, and install is often more tolerant of package-set expansion, especially when new dependencies are introduced. The presence of --skip-broken in the install command strengthens that difference even further.

kjohn-msft: Specifically package X version 1 may rely on package Y.

If package X version 2 relies on package Y and Z, it may fail on upgrade and succeed on install.

kjohn-msft (changes requested):
Please review the install vs. upgrade context deeper.

src/core/src/package_managers/Dnf5PackageManager.py
yashnap: Based on the DNF5 documentation and the validation logs, upgrade appears to be the correct command for our scenario and is producing the expected dependency information. The logs show that dnf5 upgrade --assumeno openssl correctly exposes the dependency on openssl-libs, allowing exclusion propagation to work as intended.
If you have an example where we've encountered the following scenario with any package manager, I'd be interested in taking a look at how it was handled:
`Specifically package X version 1 may rely on package Y.
If package X version 2 relies on package Y and Z, it may fail on upgrade and succeed on install.`

install_upgrade
 
I haven't been able to reproduce that behavior with DNF5.
Also, the DNF5 install documentation explicitly states that if the package is already installed, it does not re-evaluate dependencies. That aligns with what we're seeing:
```
dnf5 install --assumeno --skip-broken openssl
Package "openssl" is already installed.
Nothing to do.
```

kjohn-msft: @rane-rajasi to review tdnf install vs. upgrade as this nuance exists there too.

kjohn-msft (approved):

PR #363, opened 2026-07-10
Feature: [Ubuntu, no-CVM] Adding reboot, hibernate checks and aligning all pre-reqs together

rane-rajasi:
This PR defines all pre-requisites for updating certificates in Canonical VMs, some of which were already existing and are now re-aligned in appropriate sections and some are added as new.

The pre-reqs for certificate updates are:
1. Should not be a CVM
2. Hibernation should be turned off. If it is enabled, we report an error and do not update any certificates
3. Should not contain latest certificates already
4. System uptime should not be more than 7 days, if it is, we issue a reboot if permitted"""

This change also introduces a parent package manager flag, **is_certificate_update_enabled_for_package_manager**, which can be overridden by individual distro package managers as certificate update support is added.

In-VM Test logs:
- Certificate updated in Ubuntu 22.04: [3.core.log](https://github.com/user-attachments/files/29970859/3.core.log), [3.status.txt](https://github.com/user-attachments/files/29970860/3.status.txt)

### DIFF
--- a/src/core/src/core_logic/PatchInstaller.py
+++ b/src/core/src/core_logic/PatchInstaller.py
@@ -808,20 +808,17 @@ class PatchInstaller(object):
             self.composite_logger.log_debug("Not updating certificates since this is not a default patching operation.")
             return
 
-        try:
-            is_confidential_vm, detection_details = self.env_layer.detect_confidential_vm()
-        except Exception as e:
-            self.composite_logger.log_warning("Unable to determine whether the VM is a Confidential VM before attempting the UEFI certificate update. Continuing with patch installation... [Error: {0}]".format(str(e)))
-            return
-
-        if is_confidential_vm:
-            self.composite_logger.log("Skipping UEFI certificate update because this VM was detected as a Confidential VM. [Detection={0}]".format(detection_details))
-            return
-
-        try:
-            self.package_manager.update_certs()
-        except Exception as e:
-            self.composite_logger.log_warning("An error was encountered while attempting to update certificates. Continuing with patch installation... [Error: {0}]".format(str(e)))
+        if self.__are_prerequisites_for_updating_certs_met():
+            try:
+                certs_updated = self.package_manager.update_certs()
+                if not certs_updated:
+                    error_msg = "UEFI certificate update did not complete successfully. Certificates may not have been updated."
+                    self.composite_logger.log_error(error_msg)
+                    self.status_handler.add_error_to_status(error_msg, Constants.PatchOperationErrorCodes.CERTIFICATE_UPDATE)
+            except Exception as e:
+                error_msg = "An error was encountered while attempting to update certificates. Continuing with patch installation... [Error: {0}]".format(str(e))
+                self.composite_logger.log_error(error_msg)
+                self.status_handler.add_error_to_status(error_msg, Constants.PatchOperationErrorCodes.CERTIFICATE_UPDATE)
 
     def __is_default_patching(self):
         # type: () -> bool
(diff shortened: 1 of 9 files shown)


src/core/src/core_logic/PatchInstaller.py
```diff
@@ -808,24 +808,139 @@ def try_update_certificates_for_default_patching(self):
...
+
+        return True
+
+    def __is_cvm(self):
+        # type: () -> bool
+        """ Returns true if this is a Confidential VM (CVM) """
         try:
             is_confidential_vm, detection_details = self.env_layer.detect_confidential_vm()
         except Exception as e:
             self.composite_logger.log_warning("Unable to determine whether the VM is a Confidential VM before attempting the UEFI certificate update. Continuing with patch installation... [Error: {0}]".format(str(e)))
```
kjohn-msft, line 861: Code works, but log is ambiguous (true, when non-determinable). 

And the function name thus becomes misleading. What you can only know for sure then is is_definitely_not_cvm.

If it's cvm, false, if non-determinable, false. And log for reason why it's false should be clear.

Calling function then handles the additional negation appropriately.

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -102,6 +102,9 @@ def __init__(self, env_layer, execution_config, composite_logger, telemetry_writ
         self.install_fwupd_cmd = "sudo apt-get install -y fwupd"
         self.fwupd_refresh_cmd = "sudo fwupdmgr refresh" # NOTE: This could be made generic in package manager, depending on what solution type is adopted for other distros
         self.fwupd_update_cmd = "sudo fwupdmgr update -y"
+        self.get_uptime_seconds_cmd = "cat /proc/uptime"
+        self.get_hibernation_state_cmd = "cat /sys/power/disk"
+        self.pre_cert_reboot_uptime_threshold_seconds = 7 * 24 * 60 * 60
```
kjohn-msft: _Not for this PR_ - these aren't specific to this class. Ignore for the PR but note on what needs to be done later.

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -957,6 +960,57 @@ def get_package_install_expected_avg_time_in_seconds(self):
...
+        """Return True when certificate update should be preceded by a reboot."""
+        uptime_seconds = self.__get_vm_uptime_seconds()
+        if uptime_seconds is None:
+            # Best-effort check: do not block cert flow if uptime cannot be determined.
+            return False
+
+        requires_reboot = uptime_seconds > self.pre_cert_reboot_uptime_threshold_seconds
+        self.composite_logger.log_debug("[APM][Certs] Pre-cert reboot check. [UptimeInSeconds={0}][ThresholdInSeconds={1}][RequiresReboot={2}]"
+                                        .format(str(uptime_seconds), str(self.pre_cert_reboot_uptime_threshold_seconds), str(requires_reboot)))
+        return requires_reboot
```
kjohn-msft: There is a specific reason why this function exists - the nuance for why is lost in comments (or absence thereof).

src/core/src/core_logic/PatchInstaller.py
```diff
@@ -808,24 +808,139 @@ def try_update_certificates_for_default_patching(self):
...
 
         if is_confidential_vm:
             self.composite_logger.log("Skipping UEFI certificate update because this VM was detected as a Confidential VM. [Detection={0}]".format(detection_details))
-            return
 
+        return is_confidential_vm
+
+    def can_continue_cert_update_after_reboot_check(self):
+        # type: () -> bool
+        """Attempts pre-cert reboot when required.
```
michellemcdaniel, line 871: I'm not sure if it's the naming of this function or something else, but it feels like this function is doing too many things: saying it's ok to do the cert update and also doing the reboot. I'd prefer if it was either made more clear in the function name that this will attempt to perform an update if required or to have the reboot side effect moved to a different function.

kjohn-msft (changes requested):
Review code coverage + minor comments

src/core/src/package_managers/AptitudePackageManager.py
```diff
@@ -957,6 +960,57 @@ def get_package_install_expected_avg_time_in_seconds(self):
         return self.package_install_expected_avg_time_in_seconds
 
     # region Update certificates in factory defaults
+    def should_reboot_before_cert_update(self):
```
michellemcdaniel: Wondering if this could be renamed to "is_reboot_required_before_cert_update"

src/core/src/core_logic/PatchInstaller.py
rane-rajasi: Addressed

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: Added a comment in code to address this later

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: Added a comment in the function explaining the reason for a reboot check

src/core/src/core_logic/PatchInstaller.py
rane-rajasi: Renamed, let me know if this works

src/core/src/package_managers/AptitudePackageManager.py
rane-rajasi: Renamed

kjohn-msft (approved):

src/core/src/core_logic/PatchInstaller.py
michellemcdaniel: works for me

michellemcdaniel (approved):

PR #368, opened 2026-07-24
Bugfix: Azure Linux 4 AutoByPlat issue observed in Canary(/etc/dnf/automatic.conf absent by design in Linux4) 

yashnap:
**REPRO STEPS:**
Provision an Azure Linux 4.0 (AzL4) VM with DNF5
Install or verify DNF5 automatic plugin:  rpm -qa | grep dnf5-plugin-automatic
Enable machine default auto updates using timer:  
sudo systemctl enable --now dnf5-automatic.timer
Verify timer is enabled: 
systemctl is-enabled dnf5-automatic.timer -> ENABLED

No configuration file exist yet : 
/etc/dnf/automatic.conf
Run Assess/Install Patch from Azure Portal.

**Expected**: LPE should read current state which is enabled, disable it, perform the task and leave it in disabled state

**Actual**: LPE correctly Detects the current state of the system but when trying to disable, it runs into File not found error since /etc/dnf/automatic.conf is not automatically created when service is enabled in Azl4 (DNF5).

**SOLUTION**

According to their DNF5 doc ([Automatic Command — dnf5 documentation](https://dnf5.readthedocs.io/en/stable/dnf5_plugins/automatic.8.html)), it uses 2 configuration files. 

/usr/share/dnf5/dnf5-plugins/automatic.conf contains the default values and should be available once dnf5 is installed
/etc/dnf/automatic.conf contains Host-specific overrides and may not be available/created during dnf5 installation. Has to be explicitly created  
They advise using /etc/dnf/automatic.conf to customize configurations for the automatic service.

Sequence of steps for disabling machine default OS updates should be:

1. Read and backup both /usr/share/dnf5/dnf5-plugins/automatic.conf and /etc/dnf/automatic.conf .
2. Backup file contains a json structure : have 2 json structures, one named 'default-dnf5-automatic' and the other 'override-dnf5-automatic'. A non existent override file can be represented with empty apply_updates and download-updates values in backup.
3.Check if service is installed
  If not, do nothing
  If installed, pre-emptively disable it
  For this, if the override file does not exist, create a copy from default file and set apply_updates and download_updates to false in the override. DO NOT modify the default file

Now to revert auto OS update to machine default OS update on when AutoPatching is disabled:
1. Get the current auto OS config on the machine, which will include its installation, enable state on reboot, apply and download updates values in both default and override file.
2. If service is not installed, do nothing
  If installed, log current state 
  Read backup file and revert to the state it contains. i.e. if override file did not exist before, remove it. If values in default config were different to what we had logged in backup, modify default config to backup values. 
 
Since override config does not always exist, code does not throw an exception for something that is by design

**TESTING**

**ARM ID** : /subscriptions/6acc8a91-e2b0-4041-a069-c2932ab42fd9/resourceGroups/azl4-rg-canary/providers/Microsoft.Compute/virtualMachines/linux4-auto-os 

1. No service Installed
[no_service_installed.log](https://github.com/user-attachments/files/30358397/no_service_installed.log)

2. Service instal
[service_installed_no_timer.log](https://github.com/user-attachments/files/30358420/service_installed_no_timer.log)
led but timer not enabled

3. Service inst
[service_installed_yes_timer.log](https://github.com/user-attachments/files/30358424/service_installed_yes_timer.log)
alled and timer enabled

4. Idempotent call checking if backup 
[revert_idempotent.log](https://github.com/user-attachments/files/30358469/revert_idempotent.log)
is not overriden/recreated

5. Revert to Default( Offboarding from GuestPatching) and service installed
[revert_service_installed.log](https://github.com/user-attachments/files/30358473/revert_service_installed.log)

7. Idempotent call after revert
[revert_idempotent.log](https://github.com/user-attachments/files/30358479/revert_idempotent.log)

8. Revert back ( Onboard again to Guest Patching)
[revert_to_Azure.log](https://github.com/user-attachments/files/30358481/revert_to_Azure.log)

<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:dt="uuid:C2F41010-65B3-11d1-A29F-00AA00C14882"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=OneNote.File>
<meta name=Generator content="Microsoft OneNote 15">
</head>

<body lang=en-US style='font-family:Calibri;font-size:11.0pt'>

<p style='margin:0in;font-family:Calibri;font-size:11.0pt'>&nbsp;</p>

<div style='direction:ltr'>

ID | Scenario | Initial State | Findings | Result
-- | -- | -- | -- | --
S1 | Service not   installed | dnf5-plugin-automatic   not installed | No unexpected failures observed. Auto OS state DISABLED | ✅ PASS
S2 | Service installed,   timer not enabled | dnf5-plugin-automatic   installed, timer disabled, no override file | Auto OS state   correctly detected as Disabled. | ✅ PASS
S3 | Service installed,   timer enabled | dnf5-plugin-automatic   installed, timer enabled, no override file | Auto OS state   correctly detected as Enabled. During   AutoByPlatform onboarding, override file was created, timer was disabled,   backup was captured, and Auto OS state transitioned from Enabled → Disabled. | ✅ PASS
R1 | Change patch mode   from AutomaticByPlatform to ImageDefault | AutoByPlatform   state, timer disabled, backup present | Original   configuration restored, timer re-enabled, override file removed, and machine   returned to ImageDefault state. | ✅ PASS
R2 | ImageDefault   idempotent validation | ImageDefault state   already restored | Assessment was run   again. No additional configuration changes were made and existing   ImageDefault state was preserved. | ✅ PASS
R3 | Change patch mode   back to AutomaticByPlatform after restore | Restored   ImageDefault state, timer enabled, override file absent | Override file   recreated, timer disabled, and AutoByPlatform ownership successfully   re-established. | ✅ PASS

</div>

</body>

</html>

### DIFF
--- a/src/core/src/package_managers/Dnf5PackageManager.py
+++ b/src/core/src/package_managers/Dnf5PackageManager.py
@@ -59,7 +59,8 @@ class Dnf5PackageManager(PackageManager):
 
         # auto OS updates
         self.current_auto_os_update_service = None
-        self.os_patch_configuration_settings_file_path = ''
+        self.os_patch_default_configuration_settings_file_path = ''
+        self.os_patch_override_configuration_settings_file_path = ''
         self.auto_update_service_enabled = False
         self.auto_update_config_pattern_match_text = ""
         self.download_updates_identifier_text = ""
@@ -360,13 +361,17 @@ class Dnf5PackageManager(PackageManager):
         self.dnf5_automatic_enable_on_reboot_check_cmd = 'systemctl is-enabled dnf5-automatic.timer'
         self.dnf5_automatic_disable_on_reboot_cmd = 'systemctl disable --now dnf5-automatic.timer'
         self.dnf5_automatic_enable_on_reboot_cmd = 'systemctl enable --now dnf5-automatic.timer'
-        self.dnf5_automatic_configuration_file_path = '/etc/dnf/automatic.conf'
+        self.dnf5_automatic_default_configuration_file_path = '/usr/share/dnf5/dnf5-plugins/automatic.conf'
+        self.dnf5_automatic_override_configuration_file_path = '/etc/dnf/automatic.conf'
         self.dnf5_automatic_config_pattern_match_text = ' = (no|yes)'
         self.dnf5_automatic_download_updates_identifier_text = "download_updates"
         self.dnf5_automatic_apply_updates_identifier_text = "apply_updates"
         self.dnf5_automatic_enable_on_reboot_identifier_text = "enable_on_reboot"
         self.dnf5_automatic_installation_state_identifier_text = "installation_state"
         self.dnf5_auto_os_update_service = "dnf5-automatic"
+        self.dnf5_default_auto_os_config_backup_key = "default-dnf5-automatic"
+        self.dnf5_override_auto_os_config_backup_key = "override-dnf5-automatic"
+        self.dnf5_automatic_remove_override_configuration_file_cmd = 'rm -f /etc/dnf/automatic.conf'
 
     def get_current_auto_os_patch_state(self):
         """ Gets the current auto OS update patch state on the machine """
@@ -401,7 +406,8 @@ class Dnf5PackageManager(PackageManager):
             return Constants.AutomaticOSPatchStates.UNKNOWN
 
     def __init_auto_update_for_dnf5_automatic(self):
-        self.os_patch_configuration_settings_file_path = self.dnf5_automatic_configuration_file_path
+        self.os_patch_default_configuration_settings_file_path = self.dnf5_automatic_default_configuration_file_path
+        self.os_patch_override_configuration_settings_file_path = self.dnf5_automatic_override_configuration_file_path
         self.auto_update_config_pattern_match_text = self.dnf5_automatic_config_pattern_match_text
         self.download_updates_identifier_text = self.dnf5_automatic_download_updates_identifier_text
         self.apply_updates_identifier_text = self.dnf5_automatic_apply_updates_identifier_text
@@ -411,6 +417,8 @@ class Dnf5PackageManager(PackageManager):
         self.enable_on_reboot_cmd = self.dnf5_automatic_enable_on_reboot_cmd
         self.install_check_cmd = self.dnf5_automatic_install_check_cmd
         self.current_auto_os_update_service = self.dnf5_auto_os_update_service
+        self.os_patch_default_configuration_backup_key = self.dnf5_default_auto_os_config_backup_key
+        self.os_patch_override_configuration_backup_key = self.dnf5_override_auto_os_config_backup_key
 
     def __get_current_auto_os_updates_setting_on_machine(self):
         """Gets all auto-OS update settings for dnf5-automatic (DNF5) via config + timer state."""
@@ -437,17 +445,15 @@ class Dnf5PackageManager(PackageManager):
             enable_on_reboot_value = self.is_service_set_to_enable_on_reboot(self.enable_on_reboot_check_cmd)
 
             self.composite_logger.log_verbose("[DNF5] Checking if auto updates are currently enabled...")
-            image_default_patch_configuration = self.env_layer.file_system.read_with_retry(self.os_patch_configuration_settings_file_path, raise_if_not_found=False)
-            if image_default_patch_configuration is not None:
-                settings = image_default_patch_configuration.strip().split('\n')
-                for setting in settings:
-                    match = re.search(self.download_updates_identifier_text + self.auto_update_config_pattern_match_text, str(setting))
-                    if match is not None:
-                        download_updates_value = match.group(1)
+            default_download_updates_value, default_apply_updates_value, override_download_updates_value, override_apply_updates_value = self.__get_default_and_override_config_values()
 
-                    match = re.search(self.apply_updates_identifier_text + self.auto_update_config_pattern_match_text, str(setting))
-                    if match is not None:
-                        apply_updates_value = match.group(1)
+            download_updates_value = (override_download_updates_value if override_download_updates_value != "" else default_download_updates_value)
+
+            apply_updates_value = (
+                override_apply_updates_value
+                if override_apply_updates_value != ""
+                else default_apply_updates_value
+            )
 
             if download_updates_value == "":
                 self.composite_logger.log_verbose("[DNF5] Machine did not have any value set for [Setting={0}]".format(str(self.download_updates_identifier_text)))
@@ -501,7 +507,6 @@ class Dnf5PackageManager(PackageManager):
         """ Disables auto OS updates, using dnf5-automatic service, and logs the default settings the machine comes with """
         self.composite_logger.log_verbose("[DNF5] Disabling auto OS updates using dnf5-automatic")
         self.__init_auto_update_for_dnf5_automatic()
-
         self.backup_image_default_patch_configuration_if_not_exists()
 
         if not self.is_auto_update_service_installed(self.dnf5_automatic_install_check_cmd):
@@ -509,10 +514,21 @@ class Dnf5PackageManager(PackageManager):
             return
 
         self.composite_logger.log_verbose("[DNF5] Preemptively disabling auto OS updates using dnf5-automatic")
+        # Check if override.conf file exists, if not copy/create from default config location
+        self.__ensure_override_configuration_exists()
         self.update_os_patch_configuration_sub_setting(self.download_updates_identifier_text, "no", self.dnf5_automatic_config_pattern_match_text)
         self.update_os_patch_configuration_sub_setting(self.apply_updates_identifier_text, "no", self.dnf5_automatic_config_pattern_match_text)
         self.disable_auto_update_on_reboot(self.dnf5_automatic_disable_on_reboot_cmd)
         self.composite_logger.log_debug("[DNF5] Successfully disabled auto OS updates using dnf5-automatic")
+
+    def __ensure_override_configuration_exists(self):
+        override_config_file = self.env_layer.file_system.read_with_retry(self.os_patch_override_configuration_settings_file_path, raise_if_not_found=False)
+        if override_config_file is not None:
+            return
+
+        self.composite_logger.log_debug("[DNF5] Override configuration file does not exist.Creating it from default configuration.")
+        default_config = self.env_layer.file_system.read_with_retry(self.os_patch_default_configuration_settings_file_path)
+        self.env_layer.file_system.write_with_retry(self.os_patch_override_configuration_settings_file_path, default_config, mode='w+')
 
     def disable_auto_update_on_reboot(self, command):
         """ Disables auto update on reboot by executing systemctl command """
@@ -532,11 +548,17 @@ class Dnf5PackageManager(PackageManager):
             Log the default system settings a VM comes with, any subsequent updates will not be recorded"""
         """ JSON format for backup file:
                     {
-                        "dnf5-automatic": {
-                            "apply_updates": "yes/no/empty string",
-                            "download_updates": "yes/no/empty string",
-                            "enable_on_reboot": true/false,
-                            "installation_state": true/false
+                        "default-dnf5-automatic": {
+                                    "apply_updates": "yes/no/empty string",
+                                    "download_updates": "yes/no/empty string",
+                                    "enable_on_reboot": true/false,
+                                    "installation_state": true/false
+                        },
+                        "override-dnf5-automatic": {
+                                    "apply_updates": "yes/no/empty string",
+                                    "download_updates": "yes/no/empty string",
+                                    "enable_on_reboot": true/false,
+                                    "installation_state": true/false
                         }
                     } """
         try:
@@ -552,12 +574,20 @@ class Dnf5PackageManager(PackageManager):
                 self.composite_logger.log_debug("[DNF5] Since the backup is invalid, will add a new backup with the current auto OS update settings")
                 self.composite_logger.log_verbose("[DNF5] Fetching current auto OS update settings for [AutoOSUpdateService={0}]".format(str(self.current_auto_os_update_service)))
 
-                is_service_installed, enable_on_reboot_value, download_updates_value, apply_updates_value = self.__get_current_auto_os_updates_setting_on_machine()
+                is_service_installed, enable_on_reboot_value, _, _ = self.__get_current_auto_os_updates_setting_on_machine()
+
+                default_download_updates_value, default_apply_updates_value, override_download_updates_value, override_apply_updates_value = self.__get_default_and_override_config_values()
 
                 backup_image_default_patch_configuration_json_to_add = {
-                    self.current_auto_os_update_service: {
-                        self.download_updates_identifier_text: download_updates_value,
-                        self.apply_updates_identifier_text: apply_updates_value,
+                    self.os_patch_default_configuration_backup_key: {
+                        self.download_updates_identifier_text: default_download_updates_value,
+                        self.apply_updates_identifier_text: default_apply_updates_value,
+                        self.enable_on_reboot_identifier_text: enable_on_reboot_value,
+                        self.installation_state_identifier_text: is_service_installed
+                    },
+                    self.os_patch_override_configuration_backup_key: {
+                        self.download_updates_identifier_text: override_download_updates_value,
+                        self.apply_updates_identifier_text: override_apply_updates_value,
                         self.enable_on_reboot_identifier_text: enable_on_reboot_value,
                         self.installation_state_identifier_text: is_service_installed
                     }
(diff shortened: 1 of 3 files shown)


src/core/src/package_managers/Dnf5PackageManager.py
```diff
@@ -309,7 +310,7 @@ def extract_dependencies(self, output, packages):
 
             #  Remove input packages (support both pkg and pkg.arch)
             if len(dependent_package_name) != 0 and dependent_package_name not in packages and dependent_package_name not in dependencies:
-                self.composite_logger.log_verbose("[DNF5] > Dependency detected: " + dependent_package_name)
+                self.composite_logger.log_debug("[DNF5] > Dependency detected: " + dependent_package_name)
```
kjohn-msft: Does it need to be debug (at scale)? If it's more likely to be needed on a per-customer basis (vs. cross-customer), it's ok to be verbose to improve perf.

src/core/src/package_managers/Dnf5PackageManager.py
```diff
@@ -532,68 +543,124 @@ def backup_image_default_patch_configuration_if_not_exists(self):
...
         try:
             self.composite_logger.log_verbose("[DNF5] Ensuring there is a backup of the default patch state for [AutoOSUpdateService={0}]".format(str(self.current_auto_os_update_service)))
-
             image_default_patch_configuration_backup = self.__get_image_default_patch_configuration_backup()
             # verify if existing backup is valid if not, write to backup
             is_backup_valid = self.is_image_default_patch_configuration_backup_valid(image_default_patch_configuration_backup)
 
             if is_backup_valid:
-                self.composite_logger.log_debug("[DNF5] Since extension has a valid backup, no need to log the current settings again. ""[Default Auto OS update settings={0}] [File path={1}]".format(str(image_default_patch_configuration_backup),self.image_default_patch_configuration_backup_path))
+                self.composite_logger.log_debug("[DNF5] Since extension has a valid backup, no need to log the current settings again.[Default Auto OS update settings={0}] [File path={1}]".format(str(image_default_patch_configuration_backup), self.image_default_patch_configuration_backup_path))
```
kjohn-msft: Space between "again.[Default" => "again. [Default"

src/core/src/package_managers/Dnf5PackageManager.py
```diff
@@ -532,68 +543,124 @@ def backup_image_default_patch_configuration_if_not_exists(self):
...
-                and self.dnf5_automatic_enable_on_reboot_identifier_text in image_default_patch_configuration_backup[self.dnf5_auto_os_update_service] \
-                and self.dnf5_automatic_installation_state_identifier_text in image_default_patch_configuration_backup[self.dnf5_auto_os_update_service]:
-            self.composite_logger.log_debug("[DNF5] Extension has a valid backup for default dnf5-automatic configuration settings")
-            return True
-        else:
-            self.composite_logger.log_debug("[DNF5] Extension does not have a valid backup for default dnf5-automatic configuration settings")
+            default_backup_valid = self.__is_backup_valid(image_default_patch_configuration_backup, self.os_patch_default_configuration_backup_key)
+            override_backup_valid = self.__is_backup_valid(image_default_patch_configuration_backup, self.os_patch_override_configuration_backup_key)
+
+            if default_backup_valid and override_backup_valid:
```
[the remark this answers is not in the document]
kjohn-msft: Review this

src/core/src/package_managers/Dnf5PackageManager.py
```diff
@@ -636,25 +702,83 @@ def revert_auto_os_update_to_system_default_for_dnf5_automatic(self):
...
+
+            self.__restore_default_configuration_from_backup(default_backup)
+            self.__restore_override_configuration_from_backup(override_backup)
+            self.__restore_enable_on_reboot_state_from_backup(default_backup)
         else:
             self.composite_logger.log_debug("[DNF5] Since the backup is invalid or does not exist for current service, we won't be able to revert auto OS patch settings to their system default value. [Service={0}]".format(str(self.current_auto_os_update_service)))
 
+    def __remove_override_configuration_if_exists(self):
+        """Removes dnf5-automatic override configuration file if it exists.Missing override file is valid by design, so this method must not throw
+        when the file is absent."""
```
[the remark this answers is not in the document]
kjohn-msft: To be corrected.

src/core/src/package_managers/Dnf5PackageManager.py
```diff
@@ -636,25 +702,83 @@ def revert_auto_os_update_to_system_default_for_dnf5_automatic(self):
...
+        self.composite_logger.log_debug("[DNF5] Removing override configuration file to restore machine default.[Path={0}]".format(self.os_patch_override_configuration_settings_file_path))
+        code, out = self.env_layer.run_command_output(self.dnf5_automatic_remove_override_configuration_file_cmd, False, False)
+
+        if code != 0:
+            error_msg = "[DNF5] Error removing override configuration file. [Command={0}][Code={1}][Output={2}]".format(self.dnf5_automatic_remove_override_configuration_file_cmd, str(code), out)
+            self.composite_logger.log_error(error_msg)
+            self.status_handler.add_error_to_status(error_msg, Constants.PatchOperationErrorCodes.OPERATION_FAILED)
+            raise Exception(error_msg, "[{0}]".format(Constants.ERROR_ADDED_TO_STATUS))
+
+        self.composite_logger.log_debug("[DNF5] Removed override configuration file. [Command={0}][Code={1}][Output={2}]".format(self.dnf5_automatic_remove_override_configuration_file_cmd, str(code), out))
```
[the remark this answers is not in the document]
kjohn-msft: Please review.

kjohn-msft (changes requested):

src/core/src/package_managers/Dnf5PackageManager.py
```diff
@@ -630,25 +695,84 @@ def revert_auto_os_update_to_system_default_for_dnf5_automatic(self):
...
+    def __restore_override_configuration_from_backup(self, override_backup):
+        """Restore override dnf5-automatic configuration to its backed up state."""
+        override_download_updates = override_backup[self.download_updates_identifier_text]
+        override_apply_updates = override_backup[self.apply_updates_identifier_text]
+
+        # Empty values indicate override file did not exist before onboarding.
+        if override_download_updates == "" and override_apply_updates == "":
+            self.composite_logger.log_debug("[DNF5] Override dnf5-automatic configuration did not exist before onboarding.Removing override configuration file if it exists.")
+            self.__remove_override_configuration_if_exists()
+            return
```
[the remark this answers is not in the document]
kjohn-msft: Review this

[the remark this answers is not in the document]
yashnap: @kjohn-msft 

**Current Behavior**
During onboarding, we read the override config file to back up its state:
  • If the override file doesn't exist, get_config_values returns ("", "") for download_updates and apply_updates. These empty strings get stored in the backup JSON.
  • If the override file exists, we parse it line-by-line looking for download_updates and apply_updates. If found, we store their values. If not found, they remain "" — the same empty strings.

During revert, restore_override_configuration_from_backup checks:
if override_download_updates == "" and override_apply_updates == "":
    # "Override file did not exist before onboarding" → delete it
This works correctly when:
  • Override file didn't exist → backed up ("", "") → revert deletes the override file the extension created 
  • Override file existed with both settings defined → backed up ("yes", "no") → revert restores those values
 
**Copilot is complaining about the below use case:**
Override file existed but didn't define download_updates/apply_updates → backed up ("", "") → revert thinks the file never existed and deletes it, losing any other user settings in that file 

Note: Override file is created by the code and not present by design. The download/apply updates identifier would be present since it is directly copying from default config.

 **The Question**
Is that broken scenario realistic? Customer might change the values (e.g., yes → no or empty -> yes/no) but would be very unlikely to delete those keys entirely from the file.? 
This is for my understanding so I can determine whether the fix is needed or not. ( Differentiate between no file at all vs no key -> None) .

src/core/src/package_managers/Dnf5PackageManager.py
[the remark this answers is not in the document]
yashnap: Done

yashnap:
UT is failing due to sleep timeout not being enough for one of the test that was being enabled. I am fixing this as a part of another PR

src/core/tests/Test_Dnf5PackageManager.py
```diff
@@ -50,38 +50,55 @@ def mock_run_command_output_no_reboot(self, cmd, no_output=False, chk_err=True):
...
-                                                                    installation_state_value=installation_state_value, set_installation_state=set_installation_state)
+            self.__setup_backup_for_system_default_OS_update_config(package_manager, apply_updates_value=apply_updates_value, download_updates_value=download_updates_value,
+                                                                    override_apply_updates_value = override_apply_updates_value, override_download_updates_value = override_download_updates_value,
+                                                                    enable_on_reboot_value=enable_on_reboot_value, installation_state_value=installation_state_value, set_installation_state=set_installation_state)
         package_manager.revert_auto_os_update_to_system_default()
 
-    def __setup_current_auto_os_update_config(self, package_manager, config_value='',
-                                              config_file_name="automatic.conf"):
+    def __setup_current_auto_os_update_config(self, package_manager, config_value='', config_file_name=""):
```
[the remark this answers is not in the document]
michellemcdaniel: Review this

src/core/src/package_managers/Dnf5PackageManager.py
```diff
@@ -354,13 +355,16 @@ def __init_constants_for_dnf5_automatic(self):
         self.dnf5_automatic_enable_on_reboot_check_cmd = 'systemctl is-enabled dnf5-automatic.timer'
         self.dnf5_automatic_disable_on_reboot_cmd = 'systemctl disable --now dnf5-automatic.timer'
         self.dnf5_automatic_enable_on_reboot_cmd = 'systemctl enable --now dnf5-automatic.timer'
-        self.dnf5_automatic_configuration_file_path = '/etc/dnf/automatic.conf'
+        self.dnf5_automatic_default_configuration_file_path = '/usr/share/dnf5/dnf5-plugins/automatic.conf'
+        self.dnf5_automatic_override_configuration_file_path = '/etc/dnf/automatic.conf'
```
michellemcdaniel, line 359: Should override path have override in the name somewhere to make it more obvious that it is an override path

src/core/tests/Test_Dnf5PackageManager.py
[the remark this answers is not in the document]
yashnap: This is done. Please check my previous commit : https://github.com/Azure/LinuxPatchExtension/pull/368/changes/64846d3bc956d780f680ab7bcc2248bdd766f56e

src/core/src/package_managers/Dnf5PackageManager.py
yashnap: Updated this.

src/core/src/package_managers/Dnf5PackageManager.py
yashnap: Done

src/core/src/package_managers/Dnf5PackageManager.py
```diff
@@ -526,68 +536,124 @@ def backup_image_default_patch_configuration_if_not_exists(self):
...
             raise
 
+    def __get_default_and_override_config_values(self):
+        self.composite_logger.log_verbose("[DNF5] Reading default configuration file. [Path={0}]".format(self.os_patch_default_configuration_settings_file_path))
+        default_download_updates_value, default_apply_updates_value = self.__get_config_values(self.os_patch_default_configuration_settings_file_path)
+        self.composite_logger.log_verbose("[DNF5] Default configuration values found.[download_updates={0}][apply_updates={1}]".format(default_download_updates_value, default_apply_updates_value))
+
+        self.composite_logger.log_verbose("[DNF5] Reading override configuration file. [Path={0}]".format(self.os_patch_override_configuration_settings_file_path))
+        override_download_updates_value, override_apply_updates_value = self.__get_config_values(self.os_patch_override_configuration_settings_file_path)
+        self.composite_logger.log_verbose("[DNF5] Override configuration values found.[download_updates={0}][apply_updates={1}]".format(override_download_updates_value, override_apply_updates_value))
```
michellemcdaniel, line 597: What if override configuration file isn't found/doesn't have content

src/core/src/package_managers/Dnf5PackageManager.py
```diff
@@ -526,68 +536,124 @@ def backup_image_default_patch_configuration_if_not_exists(self):
...
+                and self.dnf5_automatic_installation_state_identifier_text in image_default_patch_configuration_backup[backup_key])
+
+    def update_os_patch_configuration_sub_setting(self, patch_configuration_sub_setting, value="no", config_pattern_match_text="", config_file_path=None):
         try:
             # note: adding space between the patch_configuration_sub_setting and value since, we will have to do that if we have to add a patch_configuration_sub_setting that did not exist before
-            self.composite_logger.log_debug("[DNF5] Updating system configuration settings for auto OS updates. [Patch Configuration Sub Setting={0}] [Value={1}]".format(
-                    str(patch_configuration_sub_setting), value))
-            os_patch_configuration_settings = self.env_layer.file_system.read_with_retry(self.os_patch_configuration_settings_file_path)
+            if config_file_path is None:
+                config_file_path = self.os_patch_override_configuration_settings_file_path
```
michellemcdaniel, line 653: so we use the override by default if the config_file_path is not included?

src/core/src/package_managers/Dnf5PackageManager.py
```diff
@@ -630,25 +695,84 @@ def revert_auto_os_update_to_system_default_for_dnf5_automatic(self):
...
+        self.update_os_patch_configuration_sub_setting(self.apply_updates_identifier_text, default_apply_updates, self.auto_update_config_pattern_match_text, self.os_patch_default_configuration_settings_file_path)
+
+    def __restore_override_configuration_from_backup(self, override_backup):
+        """Restore override dnf5-automatic configuration to its backed up state."""
+        override_download_updates = override_backup[self.download_updates_identifier_text]
+        override_apply_updates = override_backup[self.apply_updates_identifier_text]
+
+        # Empty values indicate override file did not exist before onboarding.
+        if override_download_updates == "" and override_apply_updates == "":
+            self.composite_logger.log_debug("[DNF5] Override dnf5-automatic configuration did not exist before onboarding.Removing override configuration file if it exists.")
```
michellemcdaniel: Missing a space after period. "onboarding.Removing" ->  "onboarding. Removing"

src/core/src/package_managers/Dnf5PackageManager.py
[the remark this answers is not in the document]
michellemcdaniel: Perhaps we could save the entire state of the override file to restore it to its original state, rather than just those two settings?

src/core/src/package_managers/Dnf5PackageManager.py
```diff
@@ -662,7 +786,7 @@ def enable_auto_update_on_reboot(self):
             self.status_handler.add_error_to_status(error_msg, Constants.PatchOperationErrorCodes.OPERATION_FAILED)
             raise Exception(error_msg, "[{0}]".format(Constants.ERROR_ADDED_TO_STATUS))
         else:
-            self.composite_logger.log_debug("[DNF5] Enabled auto update on reboot. [Command={0}][Code={1}][Output={2}]".format(command, str(code),out))
+            self.composite_logger.log_debug("[DNF5] Enabled auto update on reboot.[Command={0}][Code={1}][Output={2}]".format(command, str(code), out))
```
michellemcdaniel: Spacing comment here too.

src/core/tests/Test_Dnf5PackageManager.py
```diff
@@ -399,6 +484,7 @@ def test_revert_auto_os_update_to_system_default(self):
                 # restore sys.stdout output
                 sys.stdout = original_stdout
                 self.__assert_std_io(captured_output=captured_output,expected_output=testcase["stdio"]["expected_output"])
+            print("packagemanager" ,dir(package_manager))
```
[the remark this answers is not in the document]
michellemcdaniel: review this (also spacing is weird)

michellemcdaniel (changes requested):

src/core/src/package_managers/Dnf5PackageManager.py
yashnap: This path is mentioned in the official documentation for host-specific overrides: https://dnf5.readthedocs.io/en/latest/dnf5_plugins/automatic.8.html#run-dnf5-automatic-service

src/core/src/package_managers/Dnf5PackageManager.py
yashnap: Override config file isn't present by default, so the code is creating it when it disables machine default OS updates. 
Hence the override config file will exist. 
For the content, we copy over all the values from default -> Override config. It is unlikely that the file won't have contents unless the customer explicitly deletes them. ( this shouldn't happen IMO)

michellemcdaniel: ack

src/core/src/package_managers/Dnf5PackageManager.py
yashnap: The idea is that once we create the override file, we only use that until the customers decide to offboard. Meaning we don't touch or update the default file values and only use host specific overrides.  Hence the default. This is specifically used when this method is called from disabled_auto_os_update. (Disable OS update use case) 
During revert/offboard, we use the proper file path parameter.

src/core/src/package_managers/Dnf5PackageManager.py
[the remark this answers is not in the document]
yashnap: That is one way to do it but we want to keep LPE light-weight and storing full content would bloat the backup unnecessarily. Ther current design of storing only specific settings is consistent across the codebase. 
If the scenario I mentioned is indeed a valid/possible scenario then the fix I am suggesting is just store None as their values instead of "". It will be able to differentiate between file exists/content exist without increasing storage space. I am open to thoughts/suggestions from others.

src/core/src/package_managers/Dnf5PackageManager.py
yashnap: Updated

src/core/src/package_managers/Dnf5PackageManager.py
yashnap: done

src/core/tests/Test_Dnf5PackageManager.py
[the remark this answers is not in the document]
yashnap: Removed the print. I may have added for some debugging,

src/core/src/package_managers/Dnf5PackageManager.py
michellemcdaniel: Got it. Thanks for the context

src/core/src/package_managers/Dnf5PackageManager.py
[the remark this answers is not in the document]
michellemcdaniel: None vs empty string makes sense in that case. But yeah, I don't know if the scenario is something that is likely. Might be reasonable to guard against it even if we don't expect it to happen

src/core/src/package_managers/Dnf5PackageManager.py
```diff
@@ -526,68 +536,124 @@ def backup_image_default_patch_configuration_if_not_exists(self):
...
+        return False
+
+    def __is_backup_valid(self, image_default_patch_configuration_backup, backup_key):
+        return (backup_key in image_default_patch_configuration_backup
+                and self.dnf5_automatic_download_updates_identifier_text in image_default_patch_configuration_backup[backup_key]
+                and self.dnf5_automatic_apply_updates_identifier_text in image_default_patch_configuration_backup[backup_key]
+                and self.dnf5_automatic_enable_on_reboot_identifier_text in image_default_patch_configuration_backup[backup_key]
+                and self.dnf5_automatic_installation_state_identifier_text in image_default_patch_configuration_backup[backup_key])
+
+    def update_os_patch_configuration_sub_setting(self, patch_configuration_sub_setting, value="no", config_pattern_match_text="", config_file_path=None):
```
rane-rajasi, line 649: This is a method override of one declared in PackageManager.py (parent). While it works here because config_file_path has a default value, be careful and use new parameters in overridden methods only when necessary

rane-rajasi: Review once if this is the only way or if there is an alternative available

src/core/src/package_managers/Dnf5PackageManager.py
```diff
@@ -526,68 +536,124 @@ def backup_image_default_patch_configuration_if_not_exists(self):
...
+                        self.download_updates_identifier_text: default_download_updates_value,
+                        self.apply_updates_identifier_text: default_apply_updates_value,
+                        self.enable_on_reboot_identifier_text: enable_on_reboot_value,
+                        self.installation_state_identifier_text: is_service_installed
+                    },
+                    self.os_patch_override_configuration_backup_key: {
+                        self.download_updates_identifier_text: override_download_updates_value,
+                        self.apply_updates_identifier_text: override_apply_updates_value,
                         self.enable_on_reboot_identifier_text: enable_on_reboot_value,
                         self.installation_state_identifier_text: is_service_installed
```
rane-rajasi, line 578: Do the values for enable on reboot and installation state differ between default and configuration files? If no, there could maybe be a different solution to use here. This could also be used to solve the issue of unwanted removal of override config when download and apply updates values are "": https://github.com/Azure/LinuxPatchExtension/pull/368/changes#r4020532629

src/core/src/package_managers/Dnf5PackageManager.py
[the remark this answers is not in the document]
rane-rajasi: @yashnap The concern raised by Co-pilot is a valid and possible scenario. On override file does not exist today, in a VM you are creating from say a default template, but this environment won't always be true for all customers. Customers have their own configurations and keep changing those too. So, 
1. it is possible for an override file to already exist when the customer onboards to our service for the very first time. 
2. It is also possible for a pre-existing override config to have "" as a value for both apply updates and download updates.
3. This config is not restricted to only have the 2 configs: apply updates and download updates. It can contain other configurations. 

In a scenario where all of these 3 points apply, this code will log "" as override apply update and download updates in our backup file. And on reversion, it will delete the override config because these values being "" makes the code assume that an override file did not exist. This will be a bug where we interfere with a customer environment that did not have prior approval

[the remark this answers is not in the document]
rane-rajasi: These are the possible scenarios when a customer onboards to AutoByPlat :

1. Override config does not exist -> we create one with apply and download updates set to 'no'
2. Override config exists with apply and/or download updates -> we update these values in it to "no"
3. Override config exists but does not contain apply and download updates -> we create a new override config and write apply and download updates in it.

3 is a problematic scenario, where even during onboarding we have overridden a customer's config and have no way to recover it later.

Please re-evaluate and consider this in your design

rane-rajasi (changes requested):
Comments inline

src/core/src/package_managers/Dnf5PackageManager.py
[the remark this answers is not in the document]
yashnap: Make sense, Thanks for the explanation. 
I've implemented a solution to cover all the use-cases mentioned above:

**Fix**: Added an explicit override_file_exists boolean to the override backup block (captured before onboarding modifies anything). Revert now uses this flag plus per-key logic, so it never deletes a pre-existing file. Also trimmed the redundant enable_on_reboot/installation_state from the override block (leaner schema), with is_backup_valid updated to validate each block's own keys.
| # | Onboarding Scenario | What Revert Does | End Result |
|---|--------|------------------|------------|
| 1 | Override file did not exist (extension created one with `download_updates=no` and `apply_updates=no`) | `override_file_exists=false` → remove the override file created by the extension | Override file deleted. Machine returns to default state with no override file present. |
| 2 | Override file already existed and contained both `download_updates` and `apply_updates` keys (extension modified their values) | `override_file_exists=true` and backed-up values exist → restore each key to its original backed-up value | Override file preserved. Original `download_updates` and `apply_updates` values restored. |
| 3 | Override file already existed, but `download_updates` and `apply_updates` keys were absent while other customer configuration was present (extension appended the two keys) | `override_file_exists=true` and backed-up values are empty (`""`) → remove only the two keys appended by the extension | Override file preserved. Customer's existing configuration remains intact, and only the extension-added keys are removed. |

Overall: A pre-existing customer override file is never deleted; it's always returned to its exact pre-onboarding state (restored values or removed appended keys), and only extension-created files are removed. Covered by 3 unit tests in Test_Dnf5PackageManager.py

src/core/src/package_managers/Dnf5PackageManager.py
yashnap: No, they are the same. Please refer to my latest code changes. (i.e https://github.com/Azure/LinuxPatchExtension/pull/368#discussion_r4066375410 )

src/core/src/package_managers/Dnf5PackageManager.py
yashnap: I've updated the code to keep the override method intact(no new parameters).  Added a new private method instead that it calls internally to pass the config value and updated code to use the new private method where we were passing the config file directly.

src/core/src/package_managers/Dnf5PackageManager.py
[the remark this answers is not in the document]
yashnap: This is fixed

src/core/src/package_managers/Dnf5PackageManager.py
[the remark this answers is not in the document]
yashnap: Done

src/core/src/package_managers/Dnf5PackageManager.py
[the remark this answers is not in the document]
yashnap: The distinction comes from the backup file.
1.When file present and identifier present, download/apply updates will have yes/no value
2. When file present and identifier absent, download/apply values will be ""
3. When backup file exist, copy from default and updates the values to no

src/core/src/package_managers/Dnf5PackageManager.py
```diff
@@ -561,6 +562,7 @@ def backup_image_default_patch_configuration_if_not_exists(self):
                 self.composite_logger.log_debug("[DNF5] Since the backup is invalid, will add a new backup with the current auto OS update settings")
                 self.composite_logger.log_verbose("[DNF5] Fetching current auto OS update settings for [AutoOSUpdateService={0}]".format(str(self.current_auto_os_update_service)))
 
+                override_file_exists = self.env_layer.file_system.read_with_retry(self.os_patch_override_configuration_settings_file_path, raise_if_not_found=False) is not None
```
michellemcdaniel, line 565: This seems a bit heavy handed for checking if a file exists. Can we use something like os.path.exists() instead (or an EnvLayer wrapper if needed for mocking)

src/core/src/package_managers/Dnf5PackageManager.py
```diff
@@ -744,27 +765,36 @@ def __restore_default_configuration_from_backup(self, default_backup):
...
             self.__remove_override_configuration_if_exists()
             return
 
+        # File existed (with or without our keys) -> never delete it; restore each key to its pre-onboarding state.
         self.composite_logger.log_debug("[DNF5] Restoring override dnf5-automatic configuration values from backup. [Path={0}][download_updates={1}][apply_updates={2}]".format(
                 self.os_patch_override_configuration_settings_file_path, str(override_download_updates), str(override_apply_updates)))
 
-        self.__ensure_override_configuration_exists()
+        # A present key parses to yes/no; "" means the key was absent originally, so remove the one we appended.
+        if override_download_updates in ("yes", "no"):
```
michellemcdaniel, line 789: We are potentially reading/writing the file multiple times, could this all be done in one pass?

src/core/src/package_managers/Dnf5PackageManager.py
yashnap: Updated the code to use os.path.exists

src/core/src/package_managers/Dnf5PackageManager.py
yashnap: You are right. I've updated the code to make it single pass.

PR #370, opened 2026-07-30
Fix Packager base dir for Azure Arc

kjohn-msft:
This addresses an issue in Arc where unzipping the latest version is not working with PowerShell Core-based unzip. 

There is no issue with this in Azure.


yashnap (approved):

kjohn-msft:
Had a different team member push the change.

PR #378, opened 2026-08-06
Add LinuxPatchExtensionTeam to PR reviews

kjohn-msft:
Modifies the workflow to add the team to reviews automatically.


kjohn-msft:
Team addition doesn't work. This is not a priority - cancelling PR.

PR #384, opened 2026-09-03
Bugfix/28537460 auto assessment timeout 10min

SathishMSFT:
The `MsftLinuxPatchAutoAssess.service` uses `Type=forking`. systemd keeps the unit in its startup phase until it observes the expected forking lifecycle. If assessment startup and package-manager work exceed the default `TimeoutStartSec=90s`, systemd terminates the service before the extension writes terminal assessment status. As the result, the status remains in-progress.

Fix: 
Increased the timeout from 90 seconds ( default) to 10 minutes. This 10 minutes is inline with p99 time taken for on-demand operations. This avoids changing the established lifecycle model while allowing slow assessments to finish. The timeout remains bounded so a process that never reaches the expected state (or takes longer time than 10 minutes) is still terminated.

Testing:
Verified this by adding (>90 seconds) timeouts in the respective apt, yum,zypper package manager commands. Adding command used for apt for reference:

```
DELAY_SECONDS=120
sudo test ! -e /usr/bin/apt-get.distrib || {
    echo "apt-get diversion already exists; inspect before continuing."
    exit 1
}
sudo dpkg-divert --local --rename --add /usr/bin/apt-get
sudo tee /usr/bin/apt-get >/dev/null <<EOF
#!/usr/bin/env bash
MARKER=/var/tmp/lpe-28537460-apt-delay-used
if [ ! -e "\$MARKER" ]; then
    touch "\$MARKER"
    logger -t lpe-28537460-repro \
        "Delaying first apt-get invocation for ${DELAY_SECONDS} seconds"
    sleep ${DELAY_SECONDS}
fi
exec /usr/bin/apt-get.distrib "\$@"
EOF
sudo chmod 755 /usr/bin/apt-get
sudo rm -f /var/tmp/lpe-28537460-apt-delay-used

```


src/core/src/core_logic/ServiceManager.py
```diff
@@ -89,17 +89,18 @@ def is_service_enabled(self):
     # endregion
 
     # region - Service Unit Management
-    def create_service_unit_file(self, exec_start, desc, after="network.target", service_type="forking", wanted_by="multi-user.target"):
+    def create_service_unit_file(self, exec_start, desc, after="network.target", service_type="forking", timeout_start_sec="10min", wanted_by="multi-user.target"):
```
nikhim-um, line 92: Now that the potential collision surface between Auto Assessment and imperative patching calls has increased, could you please validate that this scenario is handled correctly? Specifically, it would be useful to initiate an InstallPatches call while an assessment is in progress and verify the behavior.

Although this is a relatively small change, I believe we should let it bake sufficiently in the canary region to observe and address any unforeseen issues before broader rollout.

As I understand it, with this approach, if an assessment completes quickly, say within 60 seconds, the process exits immediately and gracefully. The 10-minute timeout only comes into play for assessments that are delayed or take longer than expected to complete.

SathishMSFT: yes

PR #386, opened 2026-09-10
Bugfix/28537460 Fix Auto assessment status stuck in-progress

SathishMSFT:
## Issue:
`MsftLinuxPatchAutoAssess.service` declared `Type=forking`, but the shell wrapper it executes runs Core in the **foreground** and never daemonizes. systemd therefore treated the entire assessment as service *startup* and SIGTERM'd the cgroup at `TimeoutStartSec` (default 90s), before Core could write terminal status. Azure Update Manager was left showing a stale "In Progress" assessment.

## fix:
1. The fix we are doing here is to declare `MsftLinuxPatchAutoAssess.service` to `Type=simple`. This makes Auto assess service to run without terminating after 90 seconds.

2. This introduces newer issue. What if the AutoAssess service stalls/ hung indefinitely:
To fix this part, we call the bash script for auto assessment with a time out which sends termination signal at 50 minutes and kills the process at 55 minutes (if it's still running).

With both these fixes, Auto assessment service can run to completion taking > 30 seconds and upto 50 minutes. In our queries, we found that on-demand assessment operation`s p99 was close to 10 minutes. This should be sufficiently enough complete the auto assesses operations and set the status to terminal state. 

## test:
1. Added unit tests. could not simulate time out as tests are in windows. so only check of script details.
2.  verified the same in ubuntu 22, RHEL8  and SLES12 VMs. And also in ARC


src/extension/src/ProcessHandler.py
```diff
@@ -119,10 +119,11 @@ def stage_auto_assess_sh_safely(self, core_process_command):
                                   .format(cmd_core_py_path, exec_dir, core_py_path, auto_assess_sh_path, core_process_command))
 
             # generating exec script
+            auto_assess_core_command = core_process_command + " -" + Constants.AUTO_ASSESS_ONLY + " True"
             auto_assess_sh_data = "#!/usr/bin/env bash" +\
                                   "\n# Copyright 2021 Microsoft Corporation." + \
                                   "\ncd \"$(dirname \"$0\")\"" + \
-                                  "\n" + core_process_command + " -" + Constants.AUTO_ASSESS_ONLY + " True"
+                                  "\nexec timeout -s TERM -k " + str(Constants.AUTO_ASSESSMENT_KILL_GRACE_IN_SECS) + " " + str(Constants.AUTO_ASSESSMENT_MAX_RUNTIME_IN_SECS) + " " + auto_assess_core_command
```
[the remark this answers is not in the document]
SathishMSFT: Timeout command is already used in sudo permissions check which is called before this code is called.

src/extension/src/Constants.py
```diff
@@ -59,6 +59,13 @@ def __iter__(self):
     ENABLE_MAX_RUNTIME = 3
     DISABLE_MAX_RUNTIME = 13
 
+    # Bug 28537460: Under Type=simple a hung assessment keeps the unit active forever and every
+    # later timer fire becomes a no-op, so the run must be externally bounded.
+    # Maximum Runtime + grace time must stay under the hourly auto-assessment timer interval
+    # (core AUTO_ASSESSMENT_CRON_INTERVAL = PT1H) so process is killed within 1 hour if still running.
+    AUTO_ASSESSMENT_MAX_RUNTIME_IN_SECS = 3000      # 50m
```
nikhim-um, line 66: this PR https://github.com/Azure/LinuxPatchExtension/pull/87/changes
introduced behavior that waits for 3hrs if main core process is busy ( on demand ) while auto assessment service started. 

current implementation modifies it by introducing 50min timeout. I am actually in favor of this, instead of waiting for 3hrs as every one hour new process starts anyway. 

need koshy/rajasi to comment on this.

nikhim-um: In fact auto assessment can gracefully exit if core process is busy, and do assessment in next assessment if its possible

with existing service type "forking" I am unsure how was waiting 180min was possible. I guess that was never used or never executed.

nikhim-um, line 66: at the time of timeout after 50m, how status file look like? will it write terminal status i.e failed or does it continue to be inprogress state. 

if its InProgress state that means, issue is still exists with significantly reduced probability with timeout increased from 3min to 50min

nikhim-um (changes requested):

src/extension/src/Constants.py
SathishMSFT: Yes. This is the corner case where the status will still remain in-progress. But the number is significantly lower with p99 at 250 Seconds (< 10 minutes) for on-demand assessment. So, this could be caused most likely by customer environment.

SathishMSFT: Will let @kjohn-msft / @rane-rajasi  comment on this. let me know if you need any other explanation here from me.

GAURAVRAMRAKHYANI:
Have you manually tested the scenario where the auto-assessment remains stuck for more than 50 minutes and verified that the process is successfully terminated?

nikhim-um (approved):

src/extension/src/Constants.py
```diff
@@ -59,6 +59,13 @@ def __iter__(self):
     ENABLE_MAX_RUNTIME = 3
     DISABLE_MAX_RUNTIME = 13
 
+    # Bug 28537460: Under Type=simple a hung assessment keeps the unit active forever and every
+    # later timer fire becomes a no-op, so the run must be externally bounded.
+    # Maximum Runtime + grace time must stay under the hourly auto-assessment timer interval
```
GAURAVRAMRAKHYANI, line 64: The comment says runtime plus kill grace must remain below the hourly timer interval, but that relationship is not enforced. Since the timeout constants and  AUTO_ASSESSMENT_CRON_INTERVAL  live in different packages, they can drift independently. Please add a test asserting that the total timeout remains below the timer interval

SathishMSFT: This test will just be adding constants to verify if its less than another constant which does not make sense to me to run as unit test every time. Also, as you mentioned, the constants live in different packages, it will not be possible to test them without creating a duplicate in one of the files. This again defeats the purpose of testing as the original file could still be changed and test uses some other constant.

SathishMSFT:
> Have you manually tested the scenario where the auto-assessment remains stuck for more than 50 minutes and verified that the process is successfully terminated?

Verified by manually setting smaller timeout and adding delay in the package manager commands

GAURAVRAMRAKHYANI (approved):

yashnap (approved):
Please rebase

rane-rajasi (changes requested):
Can you attach your test logs in the PR? If not already done, run this on Azure VM/s and fetch logs under /var/log/azure/Microsoft.CPlat.Core.LinuxPatchExtension/
<seqno>.core.log
<seqno>.aa.core.log

Attest the time taken for both a regular patch operation and an auto assess operation to complete. (Auto assessments run hourly)

